Issue #25, opened 2015-03-08
manpage


cinemast:
Hi Lucas!

Thank you for providing codespell, I use it almost every day.

I recognized that there is no manpage for this tool. It would be nice to have one. However Paul Wise hinted that there might occur inconsistencies between the manpage and the --help option.

See here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=779900

What do you think about adding a `git help` like approach for --help and the manpage?

Greetings
Peter

lucasdemarchi:
I'm ok with it.

anatol:
+1 More documentation is better. @cinemast, please send a pull request.

jwilk:
I'm all for adding a manpage, but please don't make --help spawn man.

When I run `foo --help` I want it to display a short list of options and their descriptions, not a full-fledged manpage.

cinemast:
I think we can close this issue now.

PR #40, opened 2015-08-12
man page creation based on --help and --version output using help2man

cinemast:
Hi!

This is a pull-request for #25. I am using help2man and a include file to generate the manpage. Please review it, I think it could be useful for others too. This is how we generate the manpage of this project for Debian. So we could drop the additional steps and stick more closely to upstream :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ codespell: codespell.py
 manpage: codespell codespell.1.include
 	help2man ./codespell --include codespell.1.include --no-info --output codespell.1
 	sed -i '/\.SS \"Usage/,+2d' codespell.1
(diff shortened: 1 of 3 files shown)


anatol:
The changes look good to me. Tested and it works great.

anatol:
More nitpicks:

 gzip requires '-f' as the file might already exist
 please remove trailing whitespaces from the files. Use `git diff @{u} --check` to detect it.

cinemast:
I did add the -f option. But I could not find trailing whitespaces with the git command you suggested. Where did you find them?

anatol:
After minor cleanup I pushed your changes to the repo. Thanks for your work!

PR #41, opened 2015-08-18
check dictionary and add auxiliar->auxiliary

lucasdemarchi:
Make sure the dictionary is sorted and add another entry


anatol:
Looks good to me

PR #45, opened 2015-09-10
using correct prerequisite and fixing incomplete SORT_ARGS

GamerSource:
As $^ contains all prerequisites the codespell target fails as
sed tries to get the input from codespell.py and check-dictionary
where checkdictionary isn't a file but a makefile target only.
Use $< instead, as it returns only the first prerequisite.

Also the implementation for SORT_ARGS was completed.


GamerSource:
oh sorry missed your pull request, this achieves the same so can be ignored...

PR #52, opened 2015-10-07
Extended the dictionary about all entries from https://github.com/orb…

orbitcowboy:
…itcowboy/codespell_dictionary.


lucasdemarchi:
woow.... 44 thousand words!  I guess we will need some time to review that :-)

anatol:
A good stress test is to run codespell with this change over some large project (e.g. a linux susystem) and see how many true/false positives it reports.

lucasdemarchi:
@anatol are you volunteering for that? :-)

AndreaGiardini:
I have to say that this dictionary works quite well for me, and spotted several errors that the default dict wasn't able to detect. Thank you for this submission

orbitcowboy:
@AndreaGiardini Thank you for this positive echo!

jwilk:
Most of these additions don't look like plausible misspellings at all.
Was this dictionary generated automatically somehow?

orbitcowboy:
@jwilk You are right, parts of this dictionary are generated, others are not.

jwilk:
This PR is basically unreviewable, but here are some bugs I found:

* Many entries contain uppercase letters, or semicolons, or multi-word phrases; codespell doesn't make any use of them (see issue #74). For example:
```
a a->a
ACSII->ASCII
don;t->do not
```

* Many entries that have multiple corrections are missing trailing comma. For example:
```
ablout->absolute, about
```

* Some entries have `\"` instead  of apostrophes, for example:
```
aircrafts\"->aircraft\"s
```

* Some new entries are duplicates of existing ones. And sometimes even the same word is added multiple times:
```
withdrawl->withdraw
withdrawl->withdrawal
withdrawl->withdrawal, withdraw
```

* Some corrections are misspelled. For example:
```
aaccomodaation->accomodation
cascade->casade
```

orbitcowboy:
@jwilk Good catch! 
I didn't know that codespell does not handle uppercase letters, or semicolons, or multi-word phrases.
I think it's better to close this PR for now. I'll refactor the dictionary and provide a new PR.

@lucasdemarchi: codespell should be extended to handle uppercase letters, or semicolons, or multi-word phrases.

PR #58, opened 2016-01-08
Add 28 more words (incl 2 suggestions) in Jan 2016

qyearsley:
This patch includes a list of words that I've seen misspelled when checking spelling in another project.

### DIFF
--- a/data/dictionary.txt
+++ b/data/dictionary.txt
@@ -478,6 +478,7 @@ asthetical->aesthetical
 asthetically->aesthetically
 asume->assume
 aswell->as well
+asychronously->asynchronously
 atain->attain
 atempting->attempting
 atheistical->atheistic
@@ -685,6 +686,7 @@ campagin->campaign
 campain->campaign
 campaing->campaign
 campains->campaigns
+cancelation->cancellation
 candadate->candidate
 candiate->candidate
 candidiate->candidate
@@ -869,6 +871,7 @@ comapany->company
 comback->comeback
 combanations->combinations
 combinatins->combinations
+combinded->combined
 combusion->combustion
 comdemnation->condemnation
 comemmorates->commemorates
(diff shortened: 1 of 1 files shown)


data/dictionary.txt
```diff
@@ -2230,6 +2247,7 @@ intenational->international
 intented->intended, indented,
 intepretation->interpretation
 intepretator->interpretor
+interal->internal
```
lucasdemarchi: I guess this should be a suggestion instead... could be "integral", particularly in SW dealing with math.

data/dictionary.txt
```diff
@@ -2396,6 +2414,7 @@ liuke->like
 livley->lively
 lmits->limits
 loev->love
+loger->logger
```
lucasdemarchi: suggestion, too... It could be "longer"

data/dictionary.txt
```diff
@@ -3875,6 +3903,7 @@ succeded->succeeded
 succeds->succeeds
 succesful->successful
 succesfully->successfully
+succesfully->sucessfully
```
lucasdemarchi: doesn't seem right.

lucasdemarchi: http://dictionary.reference.com/browse/successfully

qyearsley:
Thanks for the review :-) Does it look OK now?

data/dictionary.txt
```diff
@@ -1036,6 +1041,7 @@ constituant->constituent
 constituants->constituents
 constituion->constitution
 constituional->constitutional
+constructable->constructible
```
lucasdemarchi: http://www.collinsdictionary.com/dictionary/english/constructable

it seems to be acceptable, too.

qyearsley: Makes sense; removed this line in the latest update.

PR #68, opened 2016-04-27
ENH: Restructure for Travis

### LINKED ISSUE #60: Ignoring directories
Hello,

I would like to ignore everything that's in (or anywhere below) a specific directory. This currently doesn't seem to be supported, though.
Would you mind implementing it?

Thanks for the useful tool!

### LINKED ISSUE #63: please set returncode according to file(s) OK or not
no errors = 0 and errors found should give >0.

for now it always succeeds:
```
user@box:~# ./codespell.py -i 0 $FILE
(several mistakes)
user@box:~# echo $?
0
```
but i exspect e.g. '1'

### LINKED ISSUE #64: Python2 support
Would the devs be interested in me adding support for Python2? I've worked on a number of packages with support for Python2 and Python3, so I don't expect it would be too bad.

Eric89GXL:
I don't have permissions to add a `codecov` hook (https://codecov.io/gh) so that won't work, but hopefully Travis passes.

Closes #64.
Closes #63.
Closes #60.


Eric89GXL:
@lucasdemarchi split them up into smaller commits -- ready for review/merge from my end.

Tests get coverage to ~90%. It would be good to enable `codecov` so that any future PRs/commits can be compared.

codespell_lib/_codespell.py
```diff
@@ -188,10 +187,10 @@ def open_with_internal(self, filename):
                         pass
 
                 curr += 1
-
-            finally:
+            else:
+                lines = f.readlines()
                 f.close()
```
lucasdemarchi: this means that the file won't be closed if the exception above is raised, no?

lucasdemarchi:
Just a small comment on one of the commits that may break codespell. Anyway, looks good.

codespell_lib/_codespell.py
Eric89GXL: The only line in the try clause creates the variable f, so if it fails there is no variable f and thus nothing to close

PR #72, opened 2016-06-06
Remove duplicates from the dictionary

jwilk:


codespell_lib/data/dictionary.txt
```diff
@@ -2620,8 +2620,6 @@ moduels->modules
 moent->moment
 moeny->money
 mohammedans->muslims
-moil->mohel
-moil->soil
```
Eric89GXL: shouldn't these offer two alternatives, e.g. `moil->mohel,soil,`?

jwilk:
As I wrote in the commit message, [moil](https://en.wiktionary.org/wiki/moil#English) is an actual English word, and IMO not a plausible misspelling.
I don't mind keeping it, but I don't feel it's useful, and dropping it was simpler. :)

Eric89GXL:
Makes sense, +1 for merge from me then

PR #78, opened 2016-06-10
Use HTTPS for Wikipedia URL

jwilk:


jwilk:
Yup, but that seems unrelated to the change I proposed.
Can we do this later in a separate PR?

Eric89GXL:
Sure, I was just thinking since you were already in there you could make the change. I'll do it directly in `master` after merging instead

Eric89GXL:
Thanks for the fix @jwilk

PR #79, opened 2016-06-10
README: fix the GPL link

jwilk:


Eric89GXL:
Thanks for another fix

PR #86, opened 2016-06-17
FIX: Fix command

### LINKED ISSUE #84: codespell.py does nothng if [fileN] is not specified
Previously running `codespell` without file parameter starts the check. Now `codespell.py` does nothing. The behavior should stay the same as before - if file/dir argument is not specefied then current directory should be used as a default parameter.

Eric89GXL:
@anatol can you see if this fixes your problem?

Closes #84.

### DIFF
--- a/bin/codespell.py
+++ b/bin/codespell.py
@@ -4,4 +4,4 @@ import sys
 
 if __name__ == '__main__':
     import codespell_lib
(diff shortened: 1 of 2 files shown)


anatol:
Yes it fixes the problem. Thanks.

PR #89, opened 2016-08-18
FIX: Rename executable

### LINKED ISSUE #85: Makefile is broken
The makefile is no longer working, since there is no longer a codespell.py

Eric89GXL:
Closes #85.


anatol:
Hi Eric, a few feedback comments about the commit:
 * You removed Makefile and it is unrelated to executable renaming. Please separate logically independent things into separate commits
 * Do not use prefixes like `FIX: ` they are really useless
 * Add more context into the commit messages. Renaming files from what to what? And most important why did you rename it, add links to the spec documentations, mail discussion etc...

Eric89GXL:
> You removed Makefile and it is unrelated to executable renaming. Please separate logically independent things into separate commits

Yeah that would probably be better. To make it even cleaner I'll just revert the Makefile changes since they are orthogonal.

> Do not use prefixes like FIX: they are really useless

I'm used to [following SciPy's standard] (http://docs.scipy.org/doc/numpy/dev/gitwash/development_workflow.html#writing-the-commit-message) since it seems to be very common in the Python ecosystem, but I'll change it since they don't seem to be the norm here.

> Add more context into the commit messages. Renaming files from what to what? And most important why did you rename it, add links to the spec documentations, mail discussion etc...

Hmm... GitHub links this PR to the relevant issue where the discussion is, and when it gets merged, the merge commit will point to this PR. So it feels a bit unnecessary to me, but I'll add it since it does make the standalone commit history richer.

Eric89GXL:
Done

cinemast:
Any chance the PR can get merged soon? I am trying to package the latest release for debian, but this rename would be of great help.

PR #90, opened 2016-08-31
Change the line numbering in display to 1-based

trorornmn:
Codespell prints one less line numbers than the real one. This was caused by assigning loop counter 'i' to 'cline' directory, so I fixed it by assigning "'i' + 1" to cline.


lucasdemarchi:
I squashed the fixup done on top and applied. Thanks.

PR #91, opened 2016-08-31
Fix off-by-one error

trorornmn:
I don't know whether or not this is in the specifications.
When I used codespell with interactive mode, it printed one previous line of the line that includes misspelled words and then printed choices. I checked parse_file() and found that it's calling ask_for_word_fix() with argument "line[i - 1]", so I modified it to "line[i]".

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -420,7 +420,7 @@ def parse_file(filename, colors, summary):
                 fixword = fix_case(word, misspellings[lword].data)
 
                 if options.interactive and lword not in asked_for:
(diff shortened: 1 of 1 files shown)


tobiasbrunner:
This is broken since dc39dfcf53a3 ("ENH: Add Python2.7 support"). [enumerate()](https://docs.python.org/3/library/functions.html#enumerate) starts the counter at 0 (unless `start` is set explicitly). Before that commit `i` was manually initialized at 1.

Eric89GXL:
Thanks @tobiasbrunner

PR #95, opened 2017-01-02
Add new --regex option to set regular expression used to find words

thdot:
A new command line option `-r / --regex` is implemented to allow the user to set the regular expression which is used to find the words in the input lines. This is useful in projects where you want to overwrite the
default expression, for example to split words after an underscore like in `ID_ESTABLISED.`
The new regex option cannot be used together with the `--write-changes` functionality since it is not guarantied that this will not break code.
This PR is related to issue #6.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -38,7 +38,7 @@ options = None
 file_opener = None
 quiet_level = 0
 encodings = ['utf-8', 'iso-8859-1']
-regex = re.compile(r"[\w\-']+")
+word_regex = re.compile(r"[\w\-']+")
 # Users might want to link this file into /usr/local/bin, so we resolve the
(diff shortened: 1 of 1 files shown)


pabs3:
The patch contains whitespace changes to code unrelated to the commit. I suggest putting that in a separate pull request.

thdot:
I have updated the commit and removed the whitespace changes.

thdot:
Friendly reminder-ping...

thdot:
@lucasdemarchi: do you plan to integrate this feature in the future or is there a problem with this PR?
Some kind of feedback would be nice.

lucasdemarchi:
@thdot sorry for the delay. Applied.

thdot:
Great! Thank you.
I've created a new PR a few minutes ago :)

PR #107, opened 2017-06-14
Disable auto-correction of "seeked" -> "sought".

qyearsley:
See: https://developer.mozilla.org/en-US/docs/Web/Events/seeked


luzpaz:
@qyearsley you can apparently use --ignore-words now per https://github.com/lucasdemarchi/codespell/pull/100

qyearsley:
This isn't really an issue for me now, but I thought it would be a reasonable change to include in the dictionary because "seeked" isn't necessarily wrong. Feel free to close if you prefer not to merge.

luzpaz:
Devs, this issue could be closed.

PR #114, opened 2017-07-26
fix existing 'inbetween' in dictionary.txt

luzpaz:
inbetween->in between
src: https://ell.stackexchange.com/questions/63622/should-i-use-in-between-or-inbetween-in-the-following-sentence

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -2344,7 +2344,7 @@ inappropiate->inappropriate
 inaugures->inaugurates
 inbalance->imbalance
 inbalanced->imbalanced
-inbetween->between
+inbetween->between, in between,
 incarcirated->incarcerated
 incidentially->incidentally
 incidently->incidentally


luzpaz:
@Eric89GXL ping

codespell_lib/data/dictionary.txt
```diff
@@ -2344,7 +2344,7 @@ inappropiate->inappropriate
 inaugures->inaugurates
 inbalance->imbalance
 inbalanced->imbalanced
-inbetween->between
+inbetween->in between
```
Eric89GXL: perhaps both should be offered

lucasdemarchi: I'm not sure about this one. I've seen disagreements on dictionaries about it in the past. grammarly.com says it's incorrect, but it does appear on some dictionaries as "informally accepted".

Eric89GXL: My vote would be to recommend "between" (by listing it first) but allow "in between" as second option

luzpaz:
resubmit PR

luzpaz:
soft bump

PR #106, opened 2017-06-14
Add more words to dictionary

qyearsley:
This change includes 34 new words that I found misspelled in code comments elsewhere. I ran "make sort-dictionary" and "make check-dictionary" after adding the words.

PR #122, opened 2017-08-17
Improve skipping directories

pmhahn:
Stop traversing into sub directories, if a directory is matched by --skip.
Also allow --skip to match the directory name only (and not only the absolute path).
Fix two cases were the correction was the same as the wrong spelling.


larsoner:
LGTM +1 for merge

lucasdemarchi:
Nice! I think this fixes what I tried to do on #106 but failed. Thanks!

PR #134, opened 2017-09-18

luzpaz:
non->none
normales->normals 
te->the

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3035,6 +3035,7 @@ nonsence->nonsense
 nontheless->nonetheless
 noone->no one
 norhern->northern
+normales->normals
 normaly->normally
 northen->northern
 northereastern->northeastern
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -3029,12 +3029,14 @@ nkwo->know
 nmae->name
 nodel->model
 nodels->models
+non->none
```
larsoner: `non-` is standard in British English apparently, I worry this will cause problems

luzpaz: can we test if it's followed by a hyphen? In other words if it's `non-` then disregard?

larsoner: Not sure, you could test it.

A quick `git grep` tells me that `scipy` has 33 instances of ` non `. Even the Wikipedia article on [Non-negative matrix factorization](https://en.wikipedia.org/wiki/Non-negative_matrix_factorization) has a `... non negative.`, presumably because of the rule about not hyphenating compound adjectives if they don't precede a noun. So it might even be considered correct in some circles (?).

I suspect this change will create more problems for people than it solves.

luzpaz:
Ok, will be removed in resubmitted PR

luzpaz:
FYI, the offending commit has been removed. It should be good to merge now AFAICT

PR #142, opened 2017-10-03

luzpaz:
Also includes revision of
`differents->different` 
to 
`differents->difference` 
+ added `uniq->unique` (this may be contraversial since uniq is a unix command but I figured people could whitelist it. Thoughts?)

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -14,6 +14,7 @@ abotu->about
 abouta->about a
 aboutit->about it
 aboutthe->about the
+abouve->above
 abreviate->abbreviate
 abreviated->abbreviated
 abreviation->abbreviation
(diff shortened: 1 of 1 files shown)


larsoner:
I'd leave `uniq` as is, that should be an uncommon misspelling

luzpaz:
Ok, resubmit PR without `uniq`

luzpaz:
I'd be happy to address any concerns that are blocking this PR from merge.

codespell_lib/data/dictionary.txt
```diff
@@ -1509,7 +1513,7 @@ diferently->differently
 diferrent->different
 differance->difference
 differentiatiations->differentiations
-differents->different
+differents->difference
```
larsoner: it should probably have both, with the old correction as the first choice

luzpaz: done

PR #144, opened 2017-10-05
Add new --check-filenames option to check file names as well

thdot:
A new command line option `-f `/ `--check-filenames` is implemented which enables the check of the given file names.

What do you think?


luzpaz:
👍

larsoner:
Thanks @thdot

PR #170, opened 2017-10-23
hove->have, hover, love,

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -2340,6 +2340,7 @@ hospitible->hospitable
 hould->hold, should,
 hounour->honour
 housr->hours, house,
+hove->have, hover, love,
```
larsoner: turns out `hove` is the past tense of `heave`, which I didn't know until I checked

luzpaz:
added `hove` as a result. I think we should keep the entry though? thoughts?

PR #174, opened 2017-10-27
C escapes

thdot:
A new command line option `--c-esacpes` is implemented which treats files as if they contain C-style character escapes. So for example "\nHello" is parsed as "hello" instead of "nhello".

What do you think?

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -40,6 +40,7 @@ file_opener = None
 quiet_level = 0
 encodings = ['utf-8', 'iso-8859-1']
 word_regex = re.compile(r"[\w\-']+")
+c_escape_regex = re.compile(r'\\\w')
 # Users might want to link this file into /usr/local/bin, so we resolve the
 # symbolic link path to the real path if necessary.
 default_dictionary = os.path.join(os.path.dirname(os.path.realpath(__file__)),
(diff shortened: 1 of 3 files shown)


luzpaz:
I like the idea.

codespell_lib/_codespell.py
```diff
@@ -281,6 +282,12 @@ def parse_options(args):
                       action='store_true', default=False,
                       help='Check file names as well.')
 
+    parser.add_option('--c-escapes',
+                      action='store_true', default=False,
+                      help='Treats files as if they contain C-style character '
+                           'escapes. So for example "\\nHello" is parsed as '
+                           '"hello" instead of "nhallo".')
```
larsoner: nhallo -> nhello

also `\\nHello` -> `\\nhello` or capitalize the other two

And maybe `r"\nhello"` is clearer than `"\\nhello"`?

codespell_lib/_codespell.py
```diff
@@ -40,6 +40,7 @@
 quiet_level = 0
 encodings = ['utf-8', 'iso-8859-1']
 word_regex = re.compile(r"[\w\-']+")
+c_escape_regex = re.compile(r'\\\w')
```
larsoner: Would it be better to use the [standard set of escape sequences](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences)?

thdot:
Thank you @larsoner for the good comment.
The reason why I have not updated this PR so far is that I recognized that the new option doesn't work with the write-changes flag. See my comments for #196

codespell_lib/tests/test_basic.py
```diff
@@ -256,6 +256,15 @@ def test_check_filename():
         assert_equal(cs.main('-f', d), 1)
 
 
+def test_c_escapes():
+    """Test c-esacpes option"""
```
peternewman: Meta spelling error, it should be c-escapes! :smile:

bl-ue:
ping @thdot — any progress?  If not, we can close this PR.

PR #177, opened 2017-10-30
breef->brief

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -702,6 +702,7 @@ bracese->braces
 Brasillian->Brazilian
 breakthough->breakthrough
 breakthroughts->breakthroughs
+breef->brief
```
larsoner: given the proximity of `e` to `r` on a standard keyboard I guess `beef` should also be offered

larsoner: standard US keyboard I mean

luzpaz: point taken, revised commit

PR #182, opened 2017-11-05
Add lik->like

luzpaz:


luzpaz:
any objections?

larsoner:
This one could actually also be "lick"

luzpaz:
Resubmit

PR #189, opened 2017-11-08
Add donn->done, don,

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -1654,6 +1654,7 @@ dominaton->domination
 dominent->dominant
 dominiant->dominant
 donig->doing
+donn->done
```
larsoner: "don" is also valid

luzpaz:
donn :laughing:

PR #191, opened 2017-11-09
Add get's->gets

luzpaz:
I figured that if the person put an apostrophe perhaps they could also mean `gets us` even though that is totally gramitcally incorrect.


larsoner:
I'd prefer just `gets` since it's close enough to the `gets us` meaning

luzpaz:
Resubmit as just `get's->gets`

PR #195, opened 2017-11-09
Remove 'anymore'

luzpaz:
Per https://dictionary.cambridge.org/us/grammar/british-grammar/any-more-or-anymore 
both are valid.


larsoner:
Perhaps we should just get rid of the entry

luzpaz:
Forced pushed commit to remove 'anymore'

PR #199, opened 2017-11-10
Removed whitespaces->whitespace

luzpaz:
For similiar reason as mentioned in https://github.com/lucasdemarchi/codespell/pull/195#issuecomment-343151152


luzpaz:
I don't think this entry is necessary. Thoughts?

PR #213, opened 2017-11-15

luzpaz:


luzpaz:
Think it's the appveyor config that is broken and not the commit.

PR #252, opened 2017-12-09

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -1002,6 +1002,7 @@ clinicaly->clinically
 cloack->cloak
 cloacks->cloaks
 clos->close
+cmak->cmake
```
luzpaz: https://en.wikipedia.org/wiki/CMake

Issue #135, opened 2017-09-23
Integrate QGIS spelling.dat list entries in to dictionary.txt


luzpaz:
[QGIS](http://github.com/QGIS/QGIS) has a typo detection tool in their repo with a list that they have created for well known typos. It lives at: https://github.com/qgis/QGIS/blob/master/scripts/spell_check/spelling.dat
Notice though they do prefer British English spelling. 

I was trying to write a script that would diff out all non-duplicate entries so I could add them to codespell. But I've been hitting snags so I opened up this issue.

thdot:
That looks very interesting.
Do we have a licence problem when we integrate the database? QGIS is published under the GPL.

PR #266, opened 2017-12-19
Continue migrating #135 (letter 'h')

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -2876,12 +2921,17 @@ hopfully->hopefully
...
+horzontal->horizontal
+horzontally->horizontally
 hosited->hoisted
 hospitible->hospitable
 hould->hold, should,
 hounour->honour
 housr->hours, house,
 hove->hove, have, hover, love,
+hovewer->however
+how log->how long
```
larsoner: this one has legitimate uses

luzpaz: @larsoner I was trying to think of some. Care to give an example ?

larsoner: "We have to do this because of how log files work"

luzpaz:
Removed offending entry. Ready to merge.

PR #279, opened 2017-12-21
End migrating #135 (letter 'v' through 'y'')

luzpaz:
This is the final PR for this series.


larsoner:
Thanks for migrating these @luzpaz

Issue #103, opened 2017-03-31
allow codespell to differentiate between British and American spelling


luzpaz:
Some projects decide to use British spelling over American. Can there be a flag to choose one of them exclusively?

lucasdemarchi:
Basically it would mean to use a different dictionary, or to annotate the current one with markers for british/american corrections. Then add a flag to the tool to tell what "dialect" you want.  This doesn't exist right now in codespell but could be doable.

luzpaz:
we can start a list for british word list in this thread and then when the time comes create logic mentioned in https://github.com/lucasdemarchi/codespell/issues/103#issuecomment-290755564
What would be the best way to go about this? Creating a gist? or should we just start a new file in the repo?

```
catalogue
centre
initialise
```

peternewman:
Would something slightly more complex be more flexible, and allow for more language support.

It strikes me there are a few different cases (I'm ignoring the obvious color/colour as that has the challenge with e.g. HTML, where you can't pick, although perhaps that has to be solved using other codespell features):
In British mode, you want standardized->standardised
In American mode, you want standardised->standardized
standarddi[sz]ed and any other genuine misspellings wants to map to the appropriate localised version.

In terms of the functionality already in codespell (from my brief understanding), it feels like a lot of this could be done if it could handle multiple dictionary files. Indeed looking at https://github.com/lucasdemarchi/codespell/blob/master/codespell_lib/_codespell.py#L304 it's using a dict, so just loading multiple files in order should allow this.

Theory as follows, remove standarddi[sz]ed from the main dict, store standarddised->standardised and standarddized->standardised and standardized->standardised in BrE and the zed version in AmE. Use some magic to generate a generic English one too, which merges both versions (but doesn't include standardized->standardised and standardised->standardized I guess when there's one fix, and the fix is in the other dict as a reverse map).

People who aren't bothered load the base file (without standarddi[sz]ed), and the generated common one, so standarddised and standarddized would both suggest the s and z options (for bonus points, prioritise the most likely one (based on if the misspelling had an s or a z). People who want a specific flavour load the base (optionally the common), but crucially their flavour last, which overrides so standarddised and standarddized both correct to standardised.

[...]

PR #286, opened 2018-01-01
Remove a valid spelling that was being corrected (licenced->licensed)

peternewman:
Much as this pains me as a Brit, but both ced and sed are valid, e.g. http://www.dictionary.com/browse/licenced and https://en.oxforddictionaries.com/definition/licensed . Although also https://www.englishforums.com/English/LicencedOrLicensed/bgrjnb/post.htm .

My 2001 Oxford Dictionary includes both variants.


lucasdemarchi:
What about the last comment in the third link?

> In AmE, "license" is both a noun and a verb.
> 
> In BrE, "licence" is the noun, and "license" is the verb (despite any comments to the contrary in this message thread).
> 
> Therefore, "licensed" is the ONLY correct spelling throughout the English-speaking world.
> 
> That being said, Canadians (at a rate far more than in the British Isles or elsewhere in the Commonwealth) with astounding frequency misspell the word as "licenced" (almost as bad as the Americans with "accomodation"). Canadian dictionaries point out that "licenced" is an alternative, but nonstandard, spelling, even in Canada. My hunch is that Canadians are employing a bit of "hyper British" spelling to distinguish it from American spelling, which in this case leads to Canadians often using the nonstandard spelling of "licenced."

http://www.wordreference.com/definition/licensed  lists "licence" as as noun only, not a verb, as does other dictionaries. Random House (dictionary.com) is the only dictionary I've seen that does not list "licence" as nonstandard.

lucasdemarchi:
That been said, I don't think we should annoy people that argue it is correct and use it... So, I don't object if others think this should be applied.

peternewman:
Hmm, personally I'd trust a paper dictionary over a literal anonymous person on the Internet (although you obviously need to trust what I'm saying regarding what's written in the dictionary).

This also ties into #103 .

lucasdemarchi:
@peternewman agreed

peternewman:
Can it be merged then @lucasdemarchi ? :smile:

peternewman:
Thanks @larsoner

PR #288, opened 2018-01-08
README.rst: Remove obsolete comment about .py extension

neuschaefer:
The program is always called codespell, rather than codespell.py since commit
fb6f01d ("Move bin/codespell.py to bin/codespell"), and the "install" target
was removed in commit c5fe7f4 ("Remove borken parts of Makefile and repair
manpage creation").


larsoner:
Thanks @neuschaefer

Issue #255, opened 2017-12-13
How can codespell handle 2 word grammar issues


luzpaz:
How can codespell address: 
`has build`  
`is build`  
`was build`

sylvestre:
To add to this list:
```
bigger then->bigger than
less then->less than
```

bl-ue:
And
```
the the->the
and and->and
```
etc.

PR #289, opened 2018-01-09
Continue migrating #135 (letter 'n')

luzpaz:
It seems I forgot to open a PR for this one.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5264,6 +5264,7 @@ mysogyny->misogyny
 mysterous->mysterious
 Mythraic->Mithraic
 naieve->naive
+nams->names
 Naploeon->Napoleon
 Napolean->Napoleon
 Napoleonian->Napoleonic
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -5342,26 +5374,46 @@ notasion->notation
...
+notifiying->notifying
 nott->not
 notwhithstanding->notwithstanding
+not either->neither
+not quit->not quite
 noveau->nouveau
 Novermber->November
 nowdays->nowadays
 nowe->now
+no compliant->noncompliant
```
larsoner: this can be used legitimately

luzpaz: True. These are good examples of #255 
I wonder if there is a way we can consider entering them in to codespell somehow and use some regex look aheads to utilize them?

larsoner: Not in this PR

luzpaz: fair enough

luzpaz:
removed the offending entry. Ready to be merged.

PR #295, opened 2018-01-12
Variations of 'unregistered'

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -8458,10 +8458,36 @@ unqouting->unquoting
...
+unregistred->unregistered
+unregistrs->unregisters
+unregsiter->unregister
+unregsitered->unregistered
+unregsitering->unregistering
+unregsiters->unregisters
+unregster->unregister
+unregstered->unregistered
+unregstering->unregistering
+unregsters->unregisters
```
larsoner: Have you actually seen these in practice? I worry that if we add this many variants for each word the dictionary will eventually grow too large. We might want to limit to common misspellings.

luzpaz: I keep seeing variations. I'm not sure I've seen all of them. But what happens is that I've caught many typos this way. Also, I'm just wondering, in regards to the concern of the size of the dictionary: why is this a concern ?

larsoner: Avoiding file cruft is good practice in general. Any time a file size grows without a good reason there are potential downsides, often (but not always) having to do with time spent processing the file (copy/transfer, store, read, etc.). One specific one here is that to add new entries, users need to load this file in an editor (not all handle large files well), read it (usually by eye), and add new ones. Another is that it takes time for you to add them, open a PR, and time for me (or others) to review them as they come in. The more entries, the more time. Finally, as a shared project, we should only add entries that we would expect to help others. If there are corrections that are unlikely to be seen by other people, they can be handled by multiple dictionary files (if we support multiple) or by merging the standard dictionary file with a custom one (if we don't support multiple).

luzpaz: @larsoner i understand your points. May I share with you that if you look at my GH history you can see I've tested codespell extensively across the FOSS world and have submitted many PRs and followup PRs to the same projects. I have encountered many new typos (not within `dictionary.txt`) by accident as I've corrected typos that codespell found. These typos were usually variants of preexisting typos in `dictionary.txt` but weren't caught because they were a different tense etc... I noticed that if I varied them, I would encountered more typos as I continued my search. I would argue that there is a place for these variations. 

If you are concerned for file size, perhaps we can work on incorporating some sort of regex pattern which will lessen the need to explicitly put each variation on a separate line? 
For example:
`unregst[ed|er|ing|r|ration]` 
it's not easy on the eyes but it's an idea.

As for large files & dtext-editors, I don't know what to say to this, the folks that have this issue are probably encountering more problems as the future arrives, this will not be a solitary issue for them. 

As for my time and your time, I agree, this is tedious, but for the time being it is helpful, at least for me and many FOSS projects out there. Eventually the size and rhythm will slow down because I won't have time for this anymore.

luzpaz:
Thanks!

PR #317, opened 2018-01-27
Add more words from Lintian, Debian static analyzer 

sylvestre:


codespell_lib/data/dictionary.txt
```diff
@@ -2248,6 +2252,7 @@ deatches->detaches
 deatching->detaching
 debain->Debian
 debateable->debatable
+debiab->Debian
 Debiab->Debian
```
luzpaz: `debiab` already exists in the dictionary

codespell_lib/data/dictionary.txt
```diff
@@ -1255,6 +1257,7 @@ campagin->campaign
 campain->campaign
 campaing->campaign
 campains->campaigns
+cancelation->cancellation
```
luzpaz: `cancelation` and `cancellation` are both correct https://github.com/lucasdemarchi/codespell/issues/103#issuecomment-341706499

codespell_lib/data/dictionary.txt
```diff
@@ -3599,6 +3617,7 @@ foppy->floppy
 forbad->forbade
 forbiden->forbidden
 foreing->foreign
+forewarded->forwarded
```
luzpaz: should come after the next entry. 
Also since foreward is british spelling you can add it as such:
forewarded->forewarded, forwarded,
I'll link this to https://github.com/lucasdemarchi/codespell/issues/103#issuecomment-341706499 because eventually I'd love to have a british and american feature flag.

codespell_lib/data/dictionary.txt
sylvestre: I didn't realize that you weren't managing case

sylvestre:
@luzpaz Is that ok with you? Thanks

luzpaz:
@sylvestre we  sort the dictionary like this https://github.com/lucasdemarchi/codespell/blob/master/Makefile#L18
Please sort it that way. Thanks!

PR #329, opened 2018-02-06

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -3045,6 +3047,8 @@ emblamatic->emblematic
 eminate->emanate
 eminated->emanated
 emision->emission
+emiss->remiss, amass,
```
larsoner: can also be `amiss`

luzpaz: right! btw, didn't find that the word `amissed` exists, so I didn't touch the next line

codespell_lib/data/dictionary.txt
```diff
@@ -269,6 +269,8 @@ addtional->additional
 addtionally->additionally
 adecuate->adequate
 adequit->adequate
+adge->edge, badge,
+adges->edges, badges,
```
larsoner: sorry realized one more, `adage` and `adages`

larsoner: (probably at the end of these option lists)

luzpaz: Good catch! added and squashed commits

PR #351, opened 2018-02-16
Add fempto, mimimun and valiator

peternewman:


larsoner:
> only thing is that you should squash the commits down in to 1 commit.

This can be done while merging. Thanks for the fixes

peternewman:
Thanks. Yeah squashing is a pain/impossible if using the GitHub web UI, which is where I did these particular edits.

PR #360, opened 2018-02-22
Add a couple of spelling corrections

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -6781,6 +6781,8 @@ reacing->reaching
 reacll->recall
 readabilty->readability
 readapted->re-adapted
+readded->read
```
larsoner: http://www.yourdictionary.com/readd

luzpaz: hah, didn't read the word that way. Nice catch. How about:
```
readded->readded, read,
readed->read, readd, readded,
```

larsoner: sounds good

luzpaz: ok.. revised an push -f 'ed

PR #382, opened 2018-03-05
Be more defensive

arm-in:
Be more defensive. The wrong words are not unique enough to make an automatic replacement without knowledge of the context. We do not know what the author of the original text was thinking.


larsoner:
Agreed, thanks @arm-in

PR #386, opened 2018-03-07
Variations on suptitle->subtitle typo

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -8246,6 +8246,9 @@ suprize->surprise
 suprized->surprised
 suprizing->surprising
 suprizingly->surprisingly
+suptitle->subtitle
```
larsoner: `suptitle` though rare does exist in some toolboxes (e.g., `matplotlib`)

gheeshuggah: I am aware of that. I'm making a PR for matplotlib actually to change it to subtitle()

larsoner: ... and I think this is intentional, because "suptitle" is presumably short for "super-title" (over-title) as opposed to the more common "sub-title" (under-title)

luzpaz: yea..i guess that makes sense. I will retract PR.

PR #397, opened 2018-03-11
Add chech->check, czech,

luzpaz:


arm-in:
Might also be chech->czech

luzpaz:
Thanks @arm-in 
Revised and repushed

PR #406, opened 2018-03-20

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -7846,6 +7848,7 @@ simulataneously->simultaneously
 simultanous->simultaneous
 simultanously->simultaneously
 sincerley->sincerely
+sincs->synchs, sinks, since,
```
larsoner: Could also be `syncs`. Technically `sincs` could be considered correct in signal processing code, too, because of the `sinc` function

Issue #394, opened 2018-03-10
False positive with a unicode apostrophe


sylvestre:
with foo.xml containing:
`<!ENTITY unknownProtocolFound.title "The address wasn’t understood">
`
_$ codespell foo.xml_
`foo.xml:1: wasn  ==> wasn't, was`

peternewman:
I think this is closed in https://github.com/lucasdemarchi/codespell/pull/418 . Or is this the other issue where the apostrophe means wasn is always matched, and is effectively a false positive?

sylvestre:
Indeed, thanks!

PR #418, opened 2018-03-28
Add the Unicode apostrophe as a word separator

sylvestre:
Some projects like Firefox uses ’ as word separator.

This will fix issue #394

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -38,7 +38,7 @@ options = None
 file_opener = None
 quiet_level = 0
 encodings = ['utf-8', 'iso-8859-1']
-word_regex = re.compile(r"[\w\-']+")
+word_regex = re.compile(r"[\w\-'’]+")
 # Users might want to link this file into /usr/local/bin, so we resolve the
 # symbolic link path to the real path if necessary.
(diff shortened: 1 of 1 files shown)


sylvestre:
Failed with Python 2.7
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1532: ordinal not in range(128)

Maybe it is time to end 2.7 support?

codespell_lib/_codespell.py
```diff
@@ -38,7 +38,7 @@
 file_opener = None
 quiet_level = 0
 encodings = ['utf-8', 'iso-8859-1']
-word_regex = re.compile(r"[\w\-']+")
+word_regex = re.compile(r"[\w\-'’]+")
```
luzpaz: I suggest also adding: `

sylvestre:
Any idea on how to fix that with Python 2.7?

flodolo:
> Any idea on how to fix that with Python 2.7?

This might work (haven't fully tested it)

```python
word_regex = re.compile(ur"[\w\-'’`]+")
```

sylvestre:
@luzpaz thanks to @flodolo I think we are good!

larsoner:
Thanks @sylvestre

PR #420, opened 2018-03-29
Drop path in help output to make the build reproducible

### LINKED ISSUE #419: Making the build reproducible
Hi all,

There is an effort in Debian to ensure packages do build reproducible. Currently, codespell does fail in this regard, because it includes a [path in the help output](https://github.com/lucasdemarchi/codespell/blob/master/codespell_lib/_codespell.py#L223) (and therefore in the manpage). There is a proposed patch in Debian which just disables embedding the path in the output -- but before applying it, I wanted to ask people over here: what do you think?

Also, I'm much more in favor of getting this fixed here, upstream, instead of in Debian, downstream.

Thanks for your work,
cheers,
Georg

ge-fa:
Closes #419

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -220,9 +220,8 @@ def parse_options(args):
                       action='append', metavar='FILE',
                       help='Custom dictionary file that contains spelling '
                            'corrections. If this flag is not specified or '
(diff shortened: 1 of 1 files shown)


larsoner:
Thanks @ge-fa

ge-fa:
Thank you!

PR #429, opened 2018-04-02
Add leaded->led, lead, leaded,

luzpaz:
Added this even though `leaded` is a legit word, most likely it is used incorrectly.


codespell_lib/data/dictionary.txt
```diff
@@ -5176,6 +5176,7 @@ launhed->launched
 lavae->larvae
 layed->laid
 lazyness->laziness
+leaded->lead, leaded,
```
larsoner: also "led"

luzpaz: Of course. Added and repushed

PR #434, opened 2018-04-03
manoeuvering, manoeuvre and manoeuvres are valid

EdwardBetts:
These are valid spellings in British English.


luzpaz:
Should be added to #103

PR #437, opened 2018-04-03
Fix suggested mistakes

Gelma:
Based on script of @EdwardBetts here:
https://github.com/lucasdemarchi/codespell/issues/435

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5536,7 +5536,7 @@ mimatching->mismatching
 mimicing->mimicking
 mimick->mimic
 mimicks->mimics
-mimimise->minimise
+mimimise->minimize
 mimimum->minimum
 mimimun->minimum
 minature->miniature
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -5536,7 +5536,7 @@ mimatching->mismatching
 mimicing->mimicking
 mimick->mimic
 mimicks->mimics
-mimimise->minimise
+mimimise->minimize
```
luzpaz: `minimise` is the British spelling of `minimize`

luzpaz: I think there should be an entry for both.
```
mimimise->minimise
mimimize->minimize
```

luzpaz: @Gelma 
Please add the suggestion above https://github.com/lucasdemarchi/codespell/pull/437/files#r181580550
and repush. Thanks.

PR #443, opened 2018-04-09
These spellings are valid in British English.

EdwardBetts:


luzpaz:
Please add these to #103

PR #444, opened 2018-04-09
Add some more corrections.

EdwardBetts:
These are from https://github.com/Microsoft/winfile/pull/10


pborreli:
maybe you should check all my other PRs 😄

PR #450, opened 2018-04-15
Create a llvm toolchain exclude list

sylvestre:


larsoner:
IMO these lists should live with individual projects rather than in the `codespell` repo. Eventually we could consider getting rid of the `mozilla` and `linux-kernel` lists even.

sylvestre:
What is your concern?

larsoner:
1. Maintenance burden / distribution. `llvm` maintainers are more qualified to say what should and shouldn't be included in such a list than `codespell` maintainers, and (related to this)
2. Scalability / separability. Imagine if every package started doing this, there would be many PRs to `codespell` rather than many PRs to their respective repos.

In terms of which words to include/exclude, I think the job of `codespell` should be to provide good defaults, and each package should then tweak their repos to fit their use cases.

sylvestre:
I am a llvm maintainer too. I reused what has been used a few weeks ago on clang. @luzpaz 
but yeah, I understand your point.

sylvestre:
I closed it. I created this PR for Mozilla:
https://github.com/lucasdemarchi/codespell/pull/454

Do you want me to remove the kernel list too?

larsoner:
Nah, that one has been around for 2 years so it's more likely people are using it

PR #461, opened 2018-04-26
Fix suggested mistaked (revised)

luzpaz:
@Gelma hasn't responded to previous requests to fix PR #437 therefore I intervened and resubmit. Since we support both british and english mispellings when we encounter them. 
Acknowledgement goes to their work. Thanks for the original submission.

Edit: This PR closes #437

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5896,7 +5896,7 @@ mimatching->mismatching
 mimicing->mimicking
 mimick->mimic
 mimicks->mimics
-mimimise->minimise
+mimimise->minimize
 mimimum->minimum
 mimimun->minimum
 minature->miniature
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -7954,7 +7954,6 @@ shpae->shape
 shpaes->shapes
 shpapes->shapes
 shreak->shriek
-shrinked->shrunk
```
larsoner: Why? "shrinked" is nonstandard

luzpaz: I was on the fence about it. If you think it should stay then I'll re-add

larsoner: Yeah let's put it back for now

PR #471, opened 2018-04-27
Update from LO

Gelma:
I want to submit new words in small chunks.
They are all together because if I re-sort the
dictionary, I have a lot of words moved all around because
wrong order.

So, if you like this patch, then you can/have to re-sort.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -9964,3 +9964,23 @@ yur->your
 zeebra->zebra
 ziped->zipped
 zoon->zoom, zoon,
+colmn->column
+costructor->constructor
+fropm->from
+langeuage->language
+langeuagesection->languagesection
+noo->no
+reltive->relative
+requrired->required
+resuts->results
+retunr->return
+retunrs->returns
+seuence->sequence
+seuences->sequences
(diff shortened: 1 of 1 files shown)


luzpaz:
no please don't resubmit because it makes things so much more confusing. Just take a look at all the PRs that are open in relation to your original PR. 

Currently I've broken down said PR in to 4 mini-PRs based on the letters 'a', 'b', 'c', and 'd'. So far only 'a' has been merged. It's a long process that needs many eyeballs vetting what gets in to the dictionary. If you want to help then choose a set of letters and submit each one in it's separate PR linking back to the original PR.

Gelma:
Wait, it's not resubmitting.

These are words new added since my last PR.
So you don't have duplication. These are brand new.

luzpaz:
@Gelma I'm integrating the LO file into the dictionary.txt file. So just submit your small chunks to the codespell dictionary directly (not the LO dictionary since we won't merge that separately). That way you have no issue sorting etc..

Gelma:
@luzpaz: it's exactly what I have done.

luzpaz:
@larsoner yes, all PRs have been either merged or pending. We can close this ticket.

PR #480, opened 2018-05-02
Add Ardiuno->Arduino, add wether->weather as another option

peternewman:


codespell_lib/data/dictionary.txt
```diff
@@ -935,6 +935,7 @@ archtypes->archetypes
 archvie->archive
 archvies->archives
 arcticle->article
+Ardiuno->Arduino
```
larsoner: I think this will need to be `arduino` because caps aren't handled properly.

@luzpaz maybe we should add a `Travis` check that all `entry == entry.lower()`?

peternewman: I did wonder, but then found a few US states with capitals. Will they not work, or only match if upper case, or?

larsoner: They won't work. This is fixed in #482

peternewman:
Can this be merged now #482 has gone in then @larsoner ?

larsoner:
Yep, thanks for the reminder

PR #482, opened 2018-05-04
FIX: Caps and pytest

larsoner:
This PR:

1. Adds a test that all entries are `lower`
2. Adds a test that capitalized entries are checked
3. Fixes `dictionary.txt` entries that used to be non-lower
4. Transitions to using `pytest` instead of `nose` since `nose` is no longer maintained
5. Updates tests to use plain `assert` in `pytest` style

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ ld
 codespell.egg-info
 *.pyc
 .cache/
+.pytest_cache/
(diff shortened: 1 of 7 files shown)


larsoner:
@luzpaz @EdwardBetts feel free to have a look if you have time

larsoner:
Agreed @peternewman, feel free to look to see if this is what you had in mind

larsoner:
Thanks for the looks @peternewman, I'll merge after the CIs come back happy from this one

PR #485, opened 2018-05-06
Remove superfluous \n

luzpaz:
Found via running the `\s$` regex


peternewman:
Should this be part of the tests to catch them automagically in future.

larsoner:
I'm confused, this makes the entry invalid because there is no trailing comma

luzpaz:
Not sure but this PR removed trailing whitespaces from the dictionary. Don't know why I wrote `\n` and why the PR doesn't reflect that change. Sorry for the confusion. I'll close it, regroup and resubmit when necessary

PR #489, opened 2018-05-07
shortend->shortened

EdwardBetts:


codespell_lib/data/dictionary.txt
```diff
@@ -9211,6 +9211,7 @@ shiping->shipping
 shoild->should
 shopkeeepers->shopkeepers
 shorly->shortly
+shortend->shortened
```
larsoner: or "short end" (e.g., of the stick)

EdwardBetts: Good point. Github code search has [12 hits for "shortend of"](https://github.com/search?p=1&q=%22shortend+of%22&type=Code&utf8=%E2%9C%93).

PR #436, opened 2018-04-03
Fixes for "List of typos fixed in LibreOffice #428"

Gelma:
Merged suggestion from:

https://github.com/lucasdemarchi/codespell/issues/435
https://github.com/lucasdemarchi/codespell/pull/431
https://github.com/lucasdemarchi/codespell/pull/430

Thanks a lot to: EdwardBetts, luzpaz, Adolfo Jayme-Barrientos,
@luzpaz @fitojb @EdwardBetts

PR #428, opened 2018-04-01
List of typos fixed in LibreOffice

Gelma:
This is the list of words fixed in LibreOffice.
It's a specific list, so I guess it's better if
you check word by word.

I purged words already present in *all* the history
of CodeSpell dictionary (to avoid to re-insert words
removed before, for any reason).

PR #496, opened 2018-05-09
Letter 'o' from #436

luzpaz:
Continuing to separate ~#428~ #436 in to separate PRs for each letter. Also removed/corrected some entries that were not relevant/precise and augmented others. This PR has been sorted as well. 

Please review closely, some of the entries still may be incorrect or may need refinement.


codespell_lib/data/dictionary.txt
```diff
@@ -7461,11 +7461,17 @@ obay->obey
...
 obediance->obedience
 obediant->obedient
+obejct->object
 obeserve->observe
 obession->obsession
 obgect->object
+objectss->objects
 objekt->object
+objetcs->objects
+objext->object
```
peternewman: I wonder if this shouldn't be auto-corrected due to OBJEXT ( http://www.gnu.org/software/automake/manual/html_node/LIBOBJS.html ) in Automake, the same as some C stuff has a comment

PR #508, opened 2018-05-14
Add many spelling corrections

luzpaz:
A running PR of left-over entries post #436 mega merge.


pabs3:
I think it would be best to squash these down into one commit and
ensure that the file from that commit is sorted with `make sort`.

-- 
bye,
pabs

http://bonedaddy.net/pabs3/

larsoner:
@pabs3 it's okay, we can squash during merge

PR #511, opened 2018-05-16
Wrong words in LO since 2018-04-29

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -1350,6 +1350,7 @@ bakc->back
 bakup->backup
 bakups->backups
 balacing->balancing
+balck->black
```
larsoner: also "balk"

Gelma: Thanks a lot for the review, Eric.

Well, I would say "balk" is valid, but it's not so usual to find it in source code.
I like to keep one solution, where possible, so people don't have to edit file but cat use "-w".

larsoner: I'd rather err on the side of including it and letting the user pick. Plus by consistency, there are already a lot of words (and corrections) in here that don't necessarily show up a lot in source code, but seem like more general English corrections.

In any case, a GitHub code search for "balk at" (possibly the most common use of the term) returns almost 93k instances:

https://github.com/search?utf8=%E2%9C%93&q=%22balk+at%22&type=Code

Gelma: Ok, pushed updates.

larsoner:
Thanks @Gelma

PR #516, opened 2018-05-20
simplies can also mean implies

sylvestre:


codespell_lib/data/dictionary.txt
```diff
@@ -10340,7 +10340,7 @@ simlified->simplified
 simmilar->similar
 simpley->simply
 simplier->simpler
-simplies->simplifies
+simplies->implies, simplifies
```
luzpaz: forgot the `,` at the end of the 2nd option: `simplies->implies, simplifies,`

sylvestre:
... I did fix it locally but looks like I didn't push the fix ...

PR #523, opened 2018-05-23
re-mux->remux and reupload->re-upload

EdwardBetts:


EdwardBetts:
Changed my mind. I think reupload is correct.

Issue #531, opened 2018-06-03
Testsuite is failing on TypeError: test_case_handling() missing 1 required positional argument: 'reload_codespell_lib'


sylvestre:
Using a Debian testing

```
nosetests3 -e "test_command"
...EE.....E
======================================================================
ERROR: Test ignore dictionary functionality
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_ignore_dictionary() missing 1 required positional argument: 'reload_codespell_lib'

======================================================================
ERROR: Test custom word regex
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_custom_regex() missing 1 required positional argument: 'reload_codespell_lib'

======================================================================
ERROR: Test that capitalized entries get detected properly.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_case_handling() missing 1 required positional argument: 'reload_codespell_lib'

----------------------------------------------------------------------
Ran 11 tests in 3.081s

FAILED (errors=3)

```

larsoner:
`codespell` uses `pytest`. `nosetests` is no longer maintained.

PR #532, opened 2018-06-03
Fix the testsuite issue #531

sylvestre:
Hopefully will work!

### DIFF
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -193,7 +193,7 @@ def test_summary():
         os.remove(f.name)
 
 
-def test_ignore_dictionary(reload_codespell_lib):
+def test_ignore_dictionary(r=reload_codespell_lib):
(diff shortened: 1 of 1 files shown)


sylvestre:
The initial error is fixed.
However, it fails on a new error:
```
E           AssertionError: assert 2 == 1
E            +  where 2 = <function main at 0x7f132e9b5378>('-I', '/tmp/tmp4nnv_zhc', '/tmp/tmpahf0wky9')
E            +    where <function main at 0x7f132e9b5378> = cs.main
E            +    and   '/tmp/tmp4nnv_zhc' = <_io.TextIOWrapper name='/tmp/tmp4nnv_zhc' mode='w' encoding='UTF-8'>.name

```

https://travis-ci.org/codespell-project/codespell/jobs/387345350
not sure how to fix that

larsoner:
I don't think we need to support `nose`.

PR #536, opened 2018-06-04
Remove "tim" from dictionary.txt

### LINKED ISSUE #535: False positive for "tim" ("Tim")
dictionary.txt now has: `tim->time`

But "Tim" is a valid and common first name. Thus this correction should not be proposed. Alternatively, there should be some case sensitive flag, so that it can be proposed only for the lowercase version.

vinc17fr:
"Tim" is a valid first name.
This commit fixes #535.


larsoner:
Thanks @vinc17fr

PR #539, opened 2018-06-05
Disable som->some due to Start Of Message

peternewman:


larsoner:
How common is this? Perhaps this is a candidate for `--ignore-words` in specific repos that need it?

peternewman:
Well GitHub code search has lots of hits from a variety of sources:
https://github.com/search?q=som+eom&type=Code

I'd not heard of cmo or rela (already disabled), but they seem to have plenty of code hits too.

larsoner:
som is 85k (cmo/rela are >= 5M). This doesn't doesn't seem worth it relative to the misses I'd expect to get.

FWIW, if the `cmo/rela` were opened as PRs now I'd probably even argue for them that they shouldn't be exempted, either. However, since they're in already it would probably break people's code to change it

larsoner:
(I don't know if there is a standard, I am just trying to judge the prevalence of the error vs prevalence of correct usage, trying to err on the side of the former a bit since we provide an easy way to skip false alarms)

peternewman:
> However, since they're in already it would probably break people's code to change it

I wouldn't worry too much about it, I'd suggest there are a few scenarios, either people take the OS package and dictionary, so only get breakages very rarely, but then there may be new valid typos too that need fixing. Or people run bleeding edge from git (like we do with our CI), which means we're constantly breaking, but most of them are true rather than false positives.

I'd almost go as far as to suggest you could/should remove cmo/rela, the people that have it can use --ignore-words or --exclude-file when they update and while they're fixing any new genuine typos.

FWIW, I've been leaning towards --exclude-file for now, and skipping whole lines, I guess the balance is missing typos elsewhere in that line, versus missing that typo in other words elsewhere in the codebase.

I'll close this PR now anyway.

larsoner:
> I'd almost go as far as to suggest you could/should remove cmo/rela, the people that have it can use --ignore-words or --exclude-file when they update and while they're fixing any new genuine typos.

You've convinced me, feel free to open a PR to do this if you want and we can see if anyone objects

peternewman:
Looks like cmo was done by @lucasdemarchi :
https://github.com/codespell-project/codespell/commit/f118ddbc40afbeab8db82224f1668347d09a458c and https://github.com/codespell-project/codespell/commit/95bbf25c346fa3501bf94e5a29adba6bfd27f704

No more detail, aside from it was a long time ago (25 May 2011).

Rela was added and disabled from the Wikipedia dictionary, again by Lucas:
https://github.com/codespell-project/codespell/commit/19e0f49209f008290546a54525089301a716d30b

So it's probably over/up to him for starters.

PR #548, opened 2018-06-12
Removed \s from certain results

luzpaz:
Removing some whitespace in entries that were added incorrectly.


luzpaz:
I guess this would be another good check to add in to Travis. 
`^.*-\>\s`

larsoner:
Please do. The best way to test it (test-driven development) is to have one commit that adds the check, which fails, and then a commit on top of that correcting all the entries, which passes.

larsoner:
Done @luzpaz

PR #555, opened 2018-06-23
Removed misformed->malformed

Gelma:


luzpaz:
Agreed, misformed is a valid word.

PR #557, opened 2018-06-29
Typos from LO of June

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -6946,6 +6948,7 @@ listensers->listeners
 listernes->listeners
 listner->listener
 listners->listeners
+liszt->list
```
larsoner: This is a name of a (somewhat?) famous composer, do you think we should disable it by default? Although this is "codespell" so I am a bit torn on this one, I imagine the false alarms will be much rarer than the true positives for this one

codespell_lib/data/dictionary.txt
```diff
@@ -12494,6 +12499,7 @@ wirh->with
 wirte->write
 wirtes->writes
 wirting->writing
+wit hthe->with the
```
larsoner: I think currently we cannot find errors with spaces in them, can you check?

If we can't, are you up for modifying `test_dictionary_formatting` to ensure no whitespace characters are in `err`?

Gelma: You are right. No way to match word with space, right now.
I remove this. I'm going to add a few commits, so I can empty my LO queue.

Thanks a lot,
Andrea

codespell_lib/data/dictionary.txt
luzpaz: you can always just do 
`liszt->list, liszt,`

Gelma: Thanks a lot @luzpaz . I'm going to upload new version

PR #567, opened 2018-07-13
Import Linux Kernel typo history

Gelma:
Hi everybody, I finished to import the biggest part of "fix typo" commit of Linux Kernel.

I merged in batch of 20 words for each commit.

Tell me if you prefer in another way.

Dividing by first letter get huge and tiny commits. Maybe in this way is much more manageable.


codespell_lib/data/dictionary.txt
```diff
@@ -54,6 +54,7 @@ abscence->absence
 abselutely->absolutely
 abselutly->absolutely
 absense->absence
+absolete->absolute
```
luzpaz: can also be `obsolete`

codespell_lib/data/dictionary.txt
```diff
@@ -85,9 +86,12 @@ acccept->accept
 acccepted->accepted
 acccepting->accepting
 acccepts->accepts
+accces->access
 acccess->access
+acccessed->accessed
 acccesses->accesses
 acccessible->accessible
+acccessor->accessors
```
luzpaz: the plural here is wrong. `accessor` should be the solution

codespell_lib/data/dictionary.txt
```diff
@@ -205,6 +222,7 @@ achor->anchor
 achored->anchored
 achoring->anchoring
 achors->anchors
+aci->acpi
```
luzpaz: not sure about this one, may return false positives

codespell_lib/data/dictionary.txt
```diff
@@ -254,30 +273,41 @@ acqured->acquired
 acqures->acquires
 acquring->acquiring
 acrage->acreage
+acses->cases
```
luzpaz: can also be `access`

codespell_lib/data/dictionary.txt
```diff
@@ -254,30 +273,41 @@ acqured->acquired
...
 acssumed->assumed
+actally->actually
 actaul->actual
 actaully->actually
 actibity->activity
 actine->active
+actiual->actual
+activ->active
```
luzpaz: not sure, even though this makes sense, there may be code that purposefully spells the word this way. So... it's debatable

codespell_lib/data/dictionary.txt
```diff
@@ -353,11 +387,14 @@ adin->admin
 aditional->additional
 aditionally->additionally
 aditionaly->additionally
+adivce->advise
```
luzpaz: can also be `advice` (actually should be the first solution)

codespell_lib/data/dictionary.txt
```diff
@@ -386,25 +424,32 @@ adresable->addressable
...
 advizable->advisable
+adwances->advances
 aeactivate->deactivate, activate,
 aequidistant->equidistant
 aequivalent->equivalent
 aeriel->aerial
 aeriels->aerials
 afair->affair
+afaraid->afraid
+afect->effect
```
luzpaz: 1st solution should be `affect`
https://en.oxforddictionaries.com/usage/affect-or-effect

codespell_lib/data/dictionary.txt
```diff
@@ -497,14 +547,18 @@ alege->allege
 aleged->alleged
 alegience->allegiance
 alegorical->allegorical
+aleviating->alleviating
```
luzpaz: please also add `aleviate->alleviate`

codespell_lib/data/dictionary.txt
```diff
@@ -1195,32 +1316,40 @@ attendents->attendants
...
 attribtes->attribute
+attribtue->attribute
+attribude->attribute
+attribue->attribute
 attribues->attributes
 attribuite->attribute
 attribures->attributes
 attribut->attribute
 attributei->attribute
+attributen->attriburen
```
luzpaz: should be `attribute`
note: thi may throw false positives for any german spelling

codespell_lib/data/dictionary.txt
```diff
@@ -1290,15 +1423,19 @@ auxillary->auxiliary
...
 avaialble->available
 avaiblable->available
 avaible->available
 availabe->available
 availabel->available
 availabelity->availability
 availabilty->availability
+availabl->available
 availabled->available
+availablen->availbalen
```
luzpaz: should be `available`

luzpaz:
The preferred way is to submit 1 letter at a time alphabetically, for PRs this long. 
Some PRs can include several letters to make them large enough. I've gone through the letter 'a' so far and posted my feedback above. I think you should turn this PR in to 'Import Linux Kernel typo history: letter 'a'' and move the other commits in to their separate PRs. That's what I've done in the past for large commits like this.

codespell_lib/data/dictionary.txt
peternewman: I'd agree in this case, although accessors is also a valid spelling according to Wiktionary (although not Chrome's spellchecker) https://en.wiktionary.org/wiki/accessors#English

Gelma:
Hi @luzpaz ,
  I'm going to do PRs divided by first letter.
  I keep open this PR to migrate suggestions later.


Thanks a lot,
Andrea

codespell_lib/data/dictionary.txt
Gelma: thanks a lot both for the review. Tell me how to procede.

codespell_lib/data/dictionary.txt
Gelma: Well, anyway I don't see any valid meaning for "aci", in general pourpose.
So, I should let it in, but tell me.

codespell_lib/data/dictionary.txt
Gelma: You tell me.

codespell_lib/data/dictionary.txt
Gelma: Ok. Anyway, what I do is to report the case I see, not what it could be.

Gelma:
Thanks a lot @luzpaz  for your review.

I migrated the fixes for letter A in its own branch.

So we can continue there.

If you agree, you can close this pull request.

Thanks a lot,
Andrea

codespell_lib/data/dictionary.txt
peternewman: Surely put it in, as it's a typo, and they should blacklist it if they want to purposefully misspell it?

codespell_lib/data/dictionary.txt
peternewman: Assuming this is https://en.wikipedia.org/wiki/ACPI then can you capitalise the RHS please.

PR #576, opened 2018-07-19
Import Linux Kernel typos: letter R

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -9397,6 +9411,7 @@ recompence->recompense
 recomputaion->recomputation
 reconaissance->reconnaissance
 reconcilation->reconciliation
+reconfiged->reconfigured
```
luzpaz: reconfiged may be an intentional misspelling., I think we should either keep and add `reconfiged` or delete this entirely.

codespell_lib/data/dictionary.txt
```diff
@@ -9696,9 +9743,12 @@ reonly->read-only
 reopended->reopened
 reoport->report
 reord->record
+reorer->reorder
 reorganision->reorganisation
 reosurces->resources
 reounded->rounded
+reources->resources
+reourcesn->resourcesn
```
luzpaz: huh?

codespell_lib/data/dictionary.txt
Gelma: Removed.

codespell_lib/data/dictionary.txt
Gelma: Removed both.

luzpaz: No need to remove the 1st one. The 2nd one was the issue.

Gelma:
I close this and recreate a second versione to avoid conflicts.

PR #579, opened 2018-07-19
Import Linux Kernel typos: letter O

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8110,10 +8124,12 @@ otherwsie->otherwise
 othewice->otherwise
 othewise->otherwise
 othewize->otherwise
+otho->otoh
```
luzpaz: say what?

codespell_lib/data/dictionary.txt
```diff
@@ -8134,6 +8150,7 @@ ourselfs->ourselves
 ourselve->ourselves
 ourselvs->ourselves
 oustanding->outstanding
+out-of-bound->out-of-band
```
luzpaz: could also very well be `out-of-bounds`. I haven't encountered `out-of-band` before. Not sure about this one.

codespell_lib/data/dictionary.txt
larsoner: acronym, OTOH (on the other hand)

luzpaz: oh right. heh. I guess we should keep it

codespell_lib/data/dictionary.txt
Gelma: Kernel related.
Read here:
https://en.wikipedia.org/wiki/Out-of-band_management

Anyway, I'm going to remove it.

PR #582, opened 2018-07-19
Import Linux Kernel typos: letter L

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -6865,18 +6868,23 @@ lazyness->laziness
 leaast->least
 leace->leave
 leaded->led, lead, leaded,
+leafs->leaves
```
luzpaz: `leafs` is a valid word i.e. `leafs through`. I'd add it as a result as well since it's non-typical

codespell_lib/data/dictionary.txt
```diff
@@ -6899,9 +6907,11 @@ lenths->lengths
 leran->learn
 lerans->learns
 lesstiff->lesstif
+lest->least
```
luzpaz: `lest` is a valid word. It's old english but nevertheless still used. On the fence about adding it as a result though.

codespell_lib/data/dictionary.txt
```diff
@@ -6899,9 +6907,11 @@ lenths->lengths
...
 letgitimate->legitimate
 letmost->leftmost
 leutenant->lieutenant
+levae->leave
```
luzpaz: could also be `levee`

codespell_lib/data/dictionary.txt
larsoner: I use `lest` in code sometimes, so we should keep it as an option

luzpaz: Then I propose `lest->lest, least,`

codespell_lib/data/dictionary.txt
Gelma: Removed to avoid confusion.

codespell_lib/data/dictionary.txt
Gelma: Removed.

PR #602, opened 2018-07-29

mbiesek:
*reorded->reorder
*valgrind->Valgrind


mbiesek:
Thanks for information. I removed part with Valgrind. 
Are You planning in future that this kind of errors (upper/lower case ) will be supported or not by the codespell ?

PR #613, opened 2018-08-08

luzpaz:


luzpaz:
@EdwardBetts please review

codespell_lib/data/dictionary.txt
```diff
@@ -2556,6 +2563,7 @@ confids->confides
 configire->configure
 configration->configuration
 configrations->configurations
+configruated->configuration
```
larsoner: configured

codespell_lib/data/dictionary.txt
```diff
@@ -4281,6 +4293,7 @@ eigth->eighth, eight,
 eihter->either
 einstance->instance
 eiter->either
+either either->either
```
larsoner: I thought that we were unable to find such errors. Does this actually work? If not, we should add a check in the dict checking tests that there is no whitespace in the dict keys

peternewman: My experience with spellintian is that it's a bit of a nightmare of false positives anyway. It should probably just be added as another option which does duplicate word checking, but spellintian does it across line boundaries so it catches things like fi twice in bash scripts, but I suspect line boundaries are the most likely place for people to make duplicate word entries, so it's lose lose (which ironically is an intentional duplication).

larsoner:
We should merge #616 before this one

codespell_lib/data/dictionary.txt
luzpaz: sorry, so should I remove this? BTW.. will codespell still catch typos like this ?

larsoner: No it won't in its current form. Eventually it might be worth adding such two-word combos but someone would have to volunteer and be motivated to do it.

luzpaz: Well, said person should then remember to re-implement 
https://github.com/codespell-project/codespell/pull/616/files#diff-3bddc99f1f6dffc98ea11be5a5a31ae9

larsoner: Indeed, as noted in the description of that PR

peternewman: Although I'd say yes to the related words, but not the duplicates stripped out of that. However I imagine the default, and custom regexes, are looking for words, so it's more a case of comparing the current and following word and then probably only suggesting not auto-replacing. E.g. my regex matches underscore separated variable names.

peternewman: P.S. This still needs removing to fix the test @luzpaz .

luzpaz:
Ready for merge

PR #616, opened 2018-08-12
ENH: Check for whitespace

larsoner:
Inspired by #613. This PR:

1. Moves dict checks to dedicated file (cleaner).
2. Adds check for whitespace in `err`, which `codespell` currently cannot find. (Maybe someday we will add support for word pairs or something...?)
3. Removes `err` entries that have whitespace. (We can always dig through the history to this PR to get them back if we add support later.)

### DIFF
--- a/codespell_lib/tests/test_dictionary.py
+++ b/codespell_lib/tests/test_dictionary.py
@@ -1,0 +1,37 @@
+# -*- coding: utf-8 -*-
+
+import os.path as op
+import re
+
+
+def test_dictionary_formatting():
+    """Test that all dictionary entries are in lower case and non-empty."""
+    err_dict = dict()
+    ws = re.compile(r'.*\s.*')  # whitespace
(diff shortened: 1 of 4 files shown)


peternewman:
LGTM

larsoner:
Thanks for the quick look @peternewman

codespell_lib/tests/test_dictionary.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+
+import os.path as op
+import re
+
+
+def test_dictionary_formatting():
+    """Test that all dictionary entries are in lower case and non-empty."""
+    err_dict = dict()
+    ws = re.compile(r'.*\s.*')  # whitespace
```
peternewman: I was going to say this will match just a single space, but that's not a bad thing anyway I guess...

larsoner:
Maybe more general that's it's correct. The specifics are pretty clear in the test itself

PR #607, opened 2018-08-04
Add a load of typos from Buildbot

peternewman:
From https://github.com/buildbot/buildbot, mostly from https://github.com/buildbot/buildbot/blob/master/master/docs/spelling_wordlist.txt

PR #617, opened 2018-08-13
Removed nonexisting->nonexistent

Gelma:
LO people told me "nonexisting" is right.

Can someone check, please?

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -8029,7 +8029,6 @@ nomimal->nominal
 non-inreractive->non-interactive
 noncombatents->noncombatants
 nonexistant->nonexistent
-nonexisting->nonexistent
 nonsence->nonsense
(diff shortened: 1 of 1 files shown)


larsoner:
I'd rather add a `, disabled due to being a word` or change it to `nonexisting->nonexistent, nonexisting,`.

peternewman:
It was me who added this in https://github.com/codespell-project/codespell/pull/607 . Out of curiosity, where do they see it defined as a word? I can see a few hits for the hyphenated version, but nothing for non-hyphenated in a selection of dictionaries.

peternewman:
Ah, I see from https://gerrit.libreoffice.org/#/c/58809/ that they're referencing Wiktionary:
https://en.wiktionary.org/wiki/nonexisting

See also e.g. https://english.stackexchange.com/questions/444048/non-existing-or-nonexisting

PR #621, opened 2018-08-13
Import Linux Kernel typos: letter P

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -9209,6 +9257,7 @@ presentaion->presentation
 presentaions->presentations
 presernt->present
 preserv->preserve
+preset->present
```
peternewman: This is a perfectly valid and frequently used word, therefore I don't believe it should be in the dictionary:
https://en.oxforddictionaries.com/definition/preset

PR #626, opened 2018-08-13
Import Linux Kernel typos: letter D (v.2)

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -3339,21 +3355,25 @@ defferent->different, deferent,
...
 defiend->defined
 defin->define
 definance->defiance
 definate->definite
 definately->definitely
+defination->definition
 definatly->definitely
 defind->defined, defund,
 definded->defined, defunded,
+define'd->defined
```
peternewman: Can we handle non-alphanumerics by default? I've had two false positives where someone has written `#define'd` in a comment, which while technically a misspelling, is also entirely intentional. I'm ignoring for now, but I'm just wondering if this ever caught genuine typos?

larsoner: I'm happy with that solution

peternewman: https://github.com/codespell-project/codespell/pull/629

PR #640, opened 2018-08-30
MITRE Corporation (programming slang, from CVE and CWE numbers)

arm-in:


larsoner:
How common would you expect this to be? My naive guess is not very. Perhaps this one would be better as an `ignore-words` entry for repos that need it?

arm-in:
There are pros and cons. I think you are right, ignore-words is better.
(+) if we would add every IT product, brand, company name, there would be no end.
(-) "microsoft" is in the dictionary.

larsoner:
Okay let's close this then.

PR #644, opened 2018-08-31
Revised to whet->when, what, wet,

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -13748,7 +13748,7 @@ wherease->whereas
 whereever->wherever
 wherether->whether
 whery->where
-whet->when
+whet->when, what, wet,
```
larsoner: `whet` is also a word

luzpaz: I'll add it a result, is that OK?

larsoner: Sure

luzpaz:
rebased and ready for merge

PR #648, opened 2018-09-04
ENH: Avoid globals, simpler testing

larsoner:
I initially went to clean up the tests using standard pytest fixtures `capsys` and `tmpdir`, but then hit some nasty problems due to `codespell` using global vars. This PR thus gets rid of that practice, making it more explicit which variables are used / modified and where it happens.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -26,19 +26,13 @@ from optparse import OptionParser
 import os
 import fnmatch
 
+word_regex_def = u"[\w\-'’`]+"
+encodings = ('utf-8', 'iso-8859-1')
 USAGE = """
 \t%prog [OPTIONS] [file1 file2 ... fileN]
 """
 VERSION = '1.14.0.dev0'
 
-misspellings = {}
-ignore_words = set()
-exclude_lines = set()
-options = None
-file_opener = None
-quiet_level = 0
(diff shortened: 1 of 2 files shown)


codespell_lib/_codespell.py
```diff
@@ -567,35 +556,34 @@ def _script_main():
...
-        global word_regex
-        try:
-            word_regex = re.compile(options.regex)
-        except re.error as err:
-            print('ERROR: invalid regular expression "%s" (%s)' %
-                  (options.regex, err), file=sys.stderr)
-            parser.print_help()
-            return 1
+    if options.regex and options.write_changes:
+        print('ERROR: --write-changes cannot be used together with '
```
peternewman: Out of interest why have you switched away from .error?

larsoner: Ahh yes just missed that it was a `parser.error` before

larsoner: ... actually I realized this is a consistency thing. Everywhere else we `print('ERROR: ...); return 1` for errors.

peternewman: I wonder if it might be good to do different exit codes for different error states in future, but that's probably for another day!

larsoner:
Thanks for the reviews @peternewman

Issue #185, opened 2017-11-06
Pass specific words to ignore via command line as parameters instead of needing to use a file 


luzpaz:
Thanks to @thdot per #100 we can now use `-I <file name>` to ignore certain words. Each word needs to be on a separate line in the file. That is great when there is a place to save a whitelist file to so codespell can read it. 

But say we were in a [Buildbot](https://github.com/buildbot/buildbot) environment, an automated testing framework, where we've setup codespell as a unit test... and so there isn't an easy possibility to save a whitelist file somewhere? Or another scenario, we're trying to quickly filter out a ginormous list of false positives from a codespell result, one needs to open their favorite editor and add each word on a separate line and link said file. It's not the most efficient. 

It would be then useful to then have the the ability to pass the whitelisted words sequentially ex. `codespell -I 'nto,ans,pres,ot,amin'` 
Thoughts?

PR #650, opened 2018-09-08
Add -L/--ignore-words-list to whitelist words directly

flopp:
This PR adds a new command line option `-L WORDS` / `--ignore-words-list WORDS` (where `WORDS` is a comma-separated list of words) to whitelist words directly from the command line without the need for a separate "ignore file".

Other than proposed in #185, I've chosen to use a new command line option (`-L`) instead of extending the existing command line option `-I` in order to avoid ambiguities (is the parameter a filename or a word to whitelist?).

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -228,6 +228,12 @@ def parse_options(args):
                            'by codespell. File must contain 1 word per line. '
                            'Words are case sensitive based on how they are '
                            'written in codespell_lib/data/dictionary.txt')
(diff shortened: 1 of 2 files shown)


peternewman:
Should the dictionary tests be extended to ensure the misspellings don't have commas in? I'm guessing the corrections can't already.

larsoner:
Yes can you take a staff at it? Hopefully not too difficult given the existing testing code

peternewman:
See https://github.com/codespell-project/codespell/pull/669 .

PR #663, opened 2018-09-17
Add them selves->themselves

luzpaz:


peternewman:
The dictionary doesn't currently support whitespace in the misspelled words.

luzpaz:
we should at least tag these issues for the day that the logic for codespell will support whitespace

Issue #651, opened 2018-09-09
codespell falsely reports it fixed an entry


luzpaz:
Run: `codespell -i 3 -w` on the file below which contains the following lines:
```
  // 'first' doesnt' match, but 'second' matches.

```
[text.txt](https://github.com/codespell-project/codespell/files/2364219/text.txt)
Result:
Codespell shows it FIXED it but after reopening file, nothing has changed.

peternewman:
I'm confused luzpaz, what's the comment about (the first doesnt'... bit), or should that be part of your file too? text.txt contains:
`australia`

For me.

luzpaz:
@PeterHuewe I uploaded the wrong text file. Sorry. :blush: 

Here is the correct one:
[text.txt](https://github.com/codespell-project/codespell/files/2387020/text.txt)

PeterHuewe:
@luzpaz Did you mean @peternewman instead?

peternewman:
This also happens on 1.8 from Ubuntu's packages. Replacement in general is fine, it's just words with apostrophes at the end. i.e.
`cant'`
Also doesn't work.

This has a rather unexpected affect:
`cant cant'`

peternewman:
Okay, so the issue is here:
https://github.com/codespell-project/codespell/blob/master/codespell_lib/_codespell.py#L499

' isn't in `\w` so it treats it as a word boundary, so never matches. The regex needs some clever work done on it.

peternewman:
This is actually more involved than it first seems, consider `cta->cat`. The word boundary ensures we don't make a replacement in `spectator`, which would be a mistake.

Now if we look at your example file:
`// 'first' doesnt' match, but 'second' matches.`

For the general case, we want `'` to be a word boundary, otherwise it wouldn't correct `a tiger is 'cta' like` (I know this isn't great grammar, but it hopefully proves my point).

Thoughts @larsoner . The best idea I can come up with is some second replacement regex which adds ' to the word character set/modifies the word boundary (assuming we can) and is run when the misspelling includes an apostrophe. But edge cases like `the opposite of 'cant'' is can` would still trip it up. Likewise the order of applying fixes is important, or my `cant'` example doesn't work as intended, and indeed still won't when the dictionary has both of these in:
```
cant'->can't
cant->can't
```

larsoner:
[...]

PR #665, opened 2018-09-17
Add an apostrophe test to catch the bug in #651

peternewman:


bl-ue:
Is this still valid?

TheGiraffe3 (approved):
What's the status on this pull request?

PR #669, opened 2018-09-20
ENH: Check commas are not present in the typos in the dictionary

peternewman:
As discussed here:
https://github.com/codespell-project/codespell/pull/650#issuecomment-421967365


larsoner:
Thanks @peternewman

PR #674, opened 2018-09-23
Avoid 'invalid escape sequence \s' warning in Python 3.6

flopp:
Pytest on Python 3.6 reports a warning (`DeprecationWarning: invalid escape sequence \s`) for the expression `re.match('^\s.*', rep)` in `test_dictionary.py`, as `\s` is not a valid escape sequence any more.

This PR changes the expression to `re.match(r'^\s.*', rep)`, which avoids the warning. Note that an `r`-prefixed string literal is used.


flopp:
The warning is also visible on travis-ci, e.g. at the bottom of https://travis-ci.org/codespell-project/codespell/jobs/431989336

larsoner:
Thanks @flopp

PR #690, opened 2018-10-01
Add genererated->generated and friends

peternewman:


peternewman:
I've added another typo just to confuse things.

PR #694, opened 2018-10-04
Fixed incorrect correction and created new correction from it's result

luzpaz:


larsoner:
That would be a cool thing to add, feel free when you get a chance

PR #700, opened 2018-10-08
Add some spelling corrections.

flopp:
Some spelling corrections from a local project.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3281,6 +3281,8 @@ copyritghts->copyrights
 copys->copies
 copytight->copyright
 copywrite->copyright
+cordinate->coordinate
+cordinates->coordinates
 corecct->correct
 corecctly->correctly
 corect->correct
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -6531,7 +6533,7 @@ identifer->identifier
 identifers->identifiers
 identifictaion->identification
 identifyable->identifiable
-identiy->identity
+identiy->identify,identity,
```
luzpaz: @larsoner shouldn't there be a space between `identify,identity,` ?

larsoner: Not sure if it's required, can you check, and if so open a PR to fix and improve Travis?

flopp: @luzpaz Good catch! You're right - items with multiple corrections are indeed separated by comma+space in `dictionary.txt`. But this is neither enforced by any test not does it seem to have any effect (other that being more plesant to read in the error report).

PR #703, opened 2018-10-10
Extend dictionary.

flopp:
Another set of spelling errors

Notable change: I disabled "synopsys->synopsis" because Synopsys is a well known EDA company.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -64,6 +64,7 @@ abstact->abstract
 abstaction->abstraction
 abstactions->abstractions
 abstactly->abstractly
+absulute->absolute
 absymal->abysmal
 abtract->abstract
 abudance->abundance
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -12616,7 +12633,7 @@ syncting->syncing
 syndonic->syntonic
 synonomous->synonymous
 synonymns->synonyms
-synopsys->synopsis
+synopsys->synopsis, disabled due to being a company name
```
larsoner: We have to balance false positives with false negatives. My sense is that we are better off leaving this in, and allowing the subset of people/codebases that use this name to whitelist it themselves

flopp: Ok, I understand. I'll reenable 'synopsys->synopsis'.

Thank you for the review!

PR #704, opened 2018-10-10

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -6986,6 +6986,7 @@ informations->information
 informaton->information
 informtion->information
 inforrmation->information
+infos->info
```
larsoner: `info` itself is a bit informal, and I wouldn't be surprised if `infos` were a commonly used (informal) pluralization of the word

luzpaz: for the record: we do have `informations->informations`. Though I do agree with you, I'm not hardlined about it either way.

luzpaz: reverted in 36a825f

luzpaz:
PR should be good to go

Issue #435, opened 2018-04-03
Replacements that are themselves mistakes


EdwardBetts:
Some of the suggested replacements in dictionary.txt are listed as mistakes themselves.

Examples:
```
mimimise->minimise
minimise->minimize
orginal->orignal
orignal->original
passtime->pasttime
pasttime->pastime
shrinked->shrunk
srinkd->shrinked
srinked->shrinked
tranfering->transfering
transfering->transferring
```
Code to find these mistakes:
``` python
bad_words, new_words = {}, set()
lines = [line[:-1] for line in open('dictionary.txt')]

for line in lines:
    orig, replacements = line.split('->')
    bad_words[orig] = {word.strip() for word in replacements.split(',')}
    new_words.update(bad_words[orig])

highlight = {word for word in new_words if word in bad_words and word not in bad_words[word]}
for line in lines:
    if any(word in line for word in highlight):
        print(line)
```

Gelma:
Thanks a lot! Very smart.

Now I apply this also for my PR.

peternewman:
Is it worth adding this as a Travis check too?

EdwardBetts:
I think it would make sense to add this as a Travis check. I'm not sure how to make that happen.

larsoner:
You could adapt your code to work with `test_dictionary_formatting`, which added some sanity checks within the last couple of months

https://github.com/codespell-project/codespell/blob/master/codespell_lib/tests/test_basic.py#L389

peternewman:
The dictionary checks are now here @EdwardBetts :
https://github.com/codespell-project/codespell/blob/master/codespell_lib/tests/test_dictionary.py

Do you fancy giving it a go?

larsoner:
Should be fixed in #705

PR #705, opened 2018-10-11
Check if corrections are errors themselves

flopp:
This adds a pytest based test to ensure that corrections are not errors themselves - as requested in #435 and with a similar solution as suggested by @EdwardBetts 

Fortunately, the current dictionary is clean ;)


larsoner:
https://travis-ci.org/codespell-project/codespell/jobs/440115923#L681

PR #709, opened 2018-10-15
Add wit->with

arm-in:


arm-in:
OK, what to do? I've modified the rule. Add or drop it?

luzpaz:
I'm on the fence about this since it may create a lot of false positives...but lets do it and if we get complaints then we remove it.

larsoner:
Agreed, and we'll need to address the bigger issue in #712 at some point

arm-in:
Yes, right. Single words may not contain enough information.

More accuracy could be yielded with
* word combinations (word pairs containing white space) => There have been some good entries in the dictionary previously. I would like to see that feature. IMHO it seems to be feasible effort.
* probabilities of finding words in an English text corpus => Then, it would e.g. be possible to set a threshold for highly probable errors, or rather unlikely errors. The probability numbers could be evaluated more or less automatically, once there is a text corpus.
* context (user defined, e.g. math, physics, science, law, ...) => This seems to be lot of effort

PR #716, opened 2018-10-16
Add medias->media, mediums,

luzpaz:
ref: https://english.stackexchange.com/questions/314099/should-i-use-media-or-medias


luzpaz:
screwed up on this PR. now it's fixed

PR #722, opened 2018-10-24
Add context options: -A/-B/-C; Fixes #287

### LINKED ISSUE #287: Have a 'grep -C' like feature which shows the context in which the typo is in
See https://stackoverflow.com/questions/9081/grep-a-file-but-show-several-surrounding-lines#9083
This kind of functionality would be awesome in codespell since many times the context the typo is displayed in is not enough (1 sentence context is not enough sometimes to understand the problem fully). 
`grep -A <number>` returns the `<number>` of lines **before** the match
`grep -B <number>` returns the `<number>` of lines **after** the match
`grep -C <number>` returns the `<number>` of lines **before AND after** the match

flopp:
This PR adds the options `-A/--after-context`, `-B/--before-context`, and `-C/--context` that enable the display of context lines surrounding a detected spelling error. Using `-C LINES` results in the display of LINES many context lines before and after the erroneous line, with `-B` and `-A` the before and after context can be configured separately. The default is to not show any context at all (so there's no change in behaviour if no context options are used).

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -423,9 +423,9 @@ def print_context(lines, index, context_before, context_after):
     for i in range(index - context_before, index + context_after + 1):
         if 0 <= i < len(lines):
             if i == index:
-                print('> %s' % lines[i], end='')
+                print('> %s' % lines[i].rstrip())
             else:
-                print(': %s' % lines[i], end='')
+                print(': %s' % lines[i].rstrip())
 
 
 def parse_file(filename, colors, summary, misspellings, exclude_lines,
@@ -648,11 +648,11 @@ def main(*args):
         context_before = context_after = max(0, options.context)
     elif (options.before_context is not None) or \
          (options.after_context is not None):
+        context_before = context_after = 0
         if options.before_context is not None:
             context_before = max(0, options.before_context)
         if options.after_context is not None:
             context_after = max(0, options.after_context)
-        context_before = max(0, context_before)
 
     exclude_lines = set()
     if options.exclude_file:
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -640,7 +640,7 @@ def main(*args):
     context_after = -1
     if options.context is not None:
         if (options.before_context is not None) or \
-            (options.after_context is not None):
+                (options.after_context is not None):
             print('ERROR: --context/-C cannot be used together with '
                   '--context-before/-B or --context-after/-A')
             parser.print_help()
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -642,7 +642,8 @@ def main(*args):
             return 1
         context_both = max(0, options.context)
         context = (context_both, context_both)
-    elif (options.before_context is not None):
+    elif (options.before_context is not None) or \
+            (options.after_context is not None):
         context_before = 0
         context_after = 0
         if options.before_context is not None:
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -417,20 +417,15 @@ def ask_for_word_fix(line, wrongword, misspelling, interactivity):
     return misspelling.fix, fix_case(wrongword, misspelling.data)
 
 
-def print_context(lines, index, context_before, context_after):
-    if context_before < 0:
-        return
-    for i in range(index - context_before, index + context_after + 1):
+def print_context(lines, index, context):
+    # context = (context_before, context_after)
+    for i in range(index - context[0], index + context[1] + 1):
         if 0 <= i < len(lines):
-            if i == index:
-                print('> %s' % lines[i].rstrip())
-            else:
-                print(': %s' % lines[i].rstrip())
+            print('%s %s' % ('>' if i == index else ':', lines[i].rstrip()))
 
 
 def parse_file(filename, colors, summary, misspellings, exclude_lines,
-               file_opener, word_regex, context_before, context_after,
-               options):
+               file_opener, word_regex, context, options):
     bad_count = 0
     lines = None
     changed = False
@@ -502,8 +497,9 @@ def parse_file(filename, colors, summary, misspellings, exclude_lines,
                 fixword = fix_case(word, misspellings[lword].data)
 
                 if options.interactive and lword not in asked_for:
-                    context_shown = True
-                    print_context(lines, i, context_before, context_after)
+                    if context is not None:
+                        context_shown = True
+                        print_context(lines, i, context)
                     fix, fixword = ask_for_word_fix(
                         lines[i], word, misspellings[lword],
                         options.interactive)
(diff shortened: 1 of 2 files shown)


luzpaz:
ooh I can't wait to test this! Thanks!

larsoner:
> why is flake8 suddenly complaining about the (invalid) word_regex_def that's not touched by this commit?

As the upstream libraries get updated, they can become more strict.

> `word_regex_def = u"[\\w\\-'’`]+"`

I would go with this one because it might need to be `unicode` in Python2.7 to be concatenated. Though I agree the `r""` version looks better so feel free to try it and we can see if the CIs are happy.

Can you add a few simple tests, including that things like `-A 1 -B1` with a one-line file does not break everything?

flopp:
Ok, I've added some tests for various combinations of the context options (including the case that more context is requested than available in the file). The tests actually revealed an error when only a "before context" is requested - I fixed that.

larsoner:
Horray unit testing!

@luzpaz can you give this PR a try?

luzpaz:
> @luzpaz can you give this PR a try?

So, I'm triyng to test it.. is it not enough to checkout the PR and run codespell ? Do i need to install it ? 
I'm running `codespell -h` and don't see the new flags 
```
$ codespell -A
Usage: 
	codespell [OPTIONS] [file1 file2 ... fileN]


codespell: error: no such option: -A
```

larsoner:
It depends on how you installed. If you did `python setup.py develop` or `pip install -e .` then yes checking out should be enough.

luzpaz:
@larsoner I did neither :blush: 
I just checked out the branch. 
I install codespell via pip generally 
`pip install --user --upgrade git+https://github.com/codespell-project/codespell.git`
So how do you recommend I go about testing this PR ?

larsoner:
Checkout the repo to your local machine, install in editable mode, check out this PR with some commands like (only some of which you probably need):
```
$ git checkout git@github.com:/codespell-project/codespell
$ cd codespell
$ pip install --user -e .
$ git fetch origin pull/722/head:722
$ which codespell
$ codespell -A 2 codespell_lib/tests/
```
Then whatever you check out at the time will be what `codespell` runs.

luzpaz:
It looks like it works!
Nice!
Green light from me. 

@larsoner Do i need to uninstall anything after I'm done testing ? Or just remove the 722 branch ?

luzpaz:
Actually @flopp it looks like you've diverted from `grep` behavior. 
As specified in #287 in `grep`:
> `grep -A <number>` returns the `<number>` of lines **before** the match
> `grep -B <number>` returns the `<number>` of lines **after** the match
> `grep -C <number>` returns the `<number>` of lines **before AND after** the match

In this PR 
> `-A/--after-context`, `-B/--before-context`, and `-C/--context`

So `-A` should really be **before** 
and `-B` should be **after**

larsoner:
> Do i need to uninstall anything after I'm done testing ? Or just remove the 722 branch ?

No you can just do:
```
git checkout master
```
and to update later do
```
git pull origin master
```

flopp:
@luzpaz I'm confident that #287 is wrong - and the implementation in this PR is consistent with grep. Grep's manual (https://www.gnu.org/software/grep/manual/grep.html#Context-Line-Control) clearly  states that '-A' is the 'after context' and '-B' is the 'before context'.

luzpaz:
@flopp oops.. that was prob. a moment of dyslexia. Cancel clear. You are correct. Green light from me

codespell_lib/_codespell.py
```diff
@@ -411,8 +417,20 @@ def ask_for_word_fix(line, wrongword, misspelling, interactivity):
...
 
+def print_context(lines, index, context_before, context_after):
+    if context_before < 0:
+        return
+    for i in range(index - context_before, index + context_after + 1):
+        if 0 <= i < len(lines):
+            if i == index:
+                print('> %s' % lines[i].rstrip())
+            else:
+                print(': %s' % lines[i].rstrip())
```
larsoner: Perhaps could be slightly more Pythonic?
```
    for i in range(max(index - context_before, 0),
                   min(index + context_after + 1, len(lines))):
        print('%s %s' % ('>' if i == index else ':', lines[i].rstrip())
```

codespell_lib/_codespell.py
```diff
@@ -411,8 +417,20 @@ def ask_for_word_fix(line, wrongword, misspelling, interactivity):
     return misspelling.fix, fix_case(wrongword, misspelling.data)
 
 
+def print_context(lines, index, context_before, context_after):
+    if context_before < 0:
+        return
```
larsoner: How can `context_before` ever be `< 0`? This seems like it should not happen. If so, inputs should be sanity checked / fixed so this conditional is no longer necessary.

codespell_lib/tests/test_basic.py
```diff
@@ -302,6 +302,68 @@ def test_case_handling(tmpdir, capsys):
...
+        cs.main('-C', '2', '-A', '1', d)
+        lines = capsys.readouterr()[0].split('\n')
+        assert 'ERROR' in lines[0]
+
+        # both '-C' and '-B' on the command line
+        cs.main('-C', '2', '-B', '1', d)
+        lines = capsys.readouterr()[0].split('\n')
+        assert 'ERROR' in lines[0]
+    finally:
+        os.remove(f.name)
```
larsoner: The beauty of `tmpdir` fixture is that you do not need to do this `try/except` business, `pytest` should take care of removal for you

codespell_lib/_codespell.py
flopp: `context_before` is negative if no context option is used at all. But I see that this may be confusing. I'll try to refactor the code a bit.

codespell_lib/_codespell.py
flopp: You're right. Let me refactor the code.

codespell_lib/tests/test_basic.py
flopp: Ok, I see. I just followed the other test functions, that all include the `try/finally` pattern. So, I don't need to call `os.remove(f.name)` explicitly?

peternewman:
> > why is flake8 suddenly complaining about the (invalid) word_regex_def that's not touched by this commit?
> 
> As the upstream libraries get updated, they can become more strict.
> 
> > `word_regex_def = u"[\\w\\-'’`]+"`
> 
> I would go with this one because it might need to be `unicode` in Python2.7 to be concatenated. Though I agree the `r""` version looks better so feel free to try it and we can see if the CIs are happy.
> 

I hit this on another project recently (due to the same Python update). This says you can use ur for the best of both worlds:
https://docs.python.org/2/reference/lexical_analysis.html#string-literals

`word_regex_def = ur"[\w\-'’`]+"`

larsoner:
My sense is that we should merge this and then tackle #726, WDYT @flopp ?

flopp:
@larsoner there seems to be a problem with `word_regex_def = ur"..."` and Python 3.X (see travis output for #735).

Other than that I'm ok with merging the current version of the PR.

larsoner:
Thanks @flopp

luzpaz:
Thank you @flopp and everyone else!

PR #724, opened 2018-10-24
Use proper escaping in 'word_regex_def'.

flopp:
This PR changes

```
word_regex_def = u"[\w\-'’`]+"
```

to the properly escaped

```
```

Which makes `flake8` happy ;)


larsoner:
Indeed it does, thanks!

PR #725, opened 2018-10-24

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -9086,6 +9086,8 @@ orgamise->organise
 organim->organism
 organistion->organisation
 organiztion->organization
+organsiation->organization
```
larsoner: `organisation` is the British spelling, maybe we should include it, too, especially since it was probably the intended word in this case

PR #726, opened 2018-10-28
Add a --no-context option when parsing stdin

mathieui:
I am writing a linter plugin integrating codespell, and it helps a lot
if there is only one line per warning, with only the relevant message
(the line content is already known by the editor).

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -257,6 +257,11 @@ def parse_options(args):
                            'choose one fix when more than one is '
                            'available; 3 applies both 1 and 2')
 
+    parser.add_option('-n', '--no-context',
+                      action='store_true', default=False,
+                      help='(when reading stdin) Do not print the line '
+                           'above the correction')
+
     parser.add_option('-q', '--quiet-level',
                       action='store', type='int', default=0,
                       help='Bitmask that allows codespell to run quietly. '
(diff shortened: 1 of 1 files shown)


peternewman:
Have you seen https://github.com/codespell-project/codespell/pull/722 @mathieui ?

Given that uses context to refer to the surrounding lines, which ties up with grep, I wonder if a different word would be better. I'm not really sure what to suggest though, the closest grep has is "Suppress normal output" from it's `--count` option. Maybe `--no-line` or `--no-source` something?

larsoner:
@mathieui now that #722 is merged can you see how this fits into the API

mathieui:
Thanks, I’ll look into it.

flopp:
Interesting. codespell in "stdin mode" behaves almost exactly as codespell in "file mode" with `-C0` (=> only show the faulty line but no other context):

```
$ cat codespell_lib/tests/test_basic.py | codespell -
64: f.write('abandonned abandonned\n')
	abandonned ==> abandoned
169: f.write('1 abandonned 1\n2 abandonned 2\nabondon\n')
	abandonned ==> abandoned
...
```

```
$ codespell -C0 codespell_lib/tests/test_basic.py 
>         f.write('abandonned abandonned\n')
codespell_lib/tests/test_basic.py:64: abandonned  ==> abandoned
>         f.write('1 abandonned 1\n2 abandonned 2\nabondon\n')
codespell_lib/tests/test_basic.py:169: abandonned  ==> abandoned
...
```

So, one could change the "stdin mode" to show no context by default, and use `-C0` to show the faulty line (or other `-A`, `-B`, `-C` combinations for more context lines).

PR #731, opened 2018-10-29
Add several spelling corrections for install*

cfi-gb:


codespell_lib/data/dictionary.txt
```diff
@@ -7207,6 +7207,7 @@ inpuut->input
 inreractive->interactive
 inro->into
 ins't->isn't
+insallation->installation
```
luzpaz: maybe `insulation` too?

larsoner: I doubt that would come up very often

PR #734, opened 2018-10-29
Add several spelling corrections for Microsoft

cfi-gb:
I'm seeing those quite often. Not sure if those could cause some false positives for other similar names so any feedback is very welcome.

The PR is changing the existing microsoft spelling and capitalise it based on the suggestion in https://github.com/codespell-project/codespell/pull/734#issuecomment-434089546

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -8272,10 +8272,16 @@ mial->mail
 mices->mice
 Michagan->Michigan
 micorcode->microcode
+Micorsoft->Microsoft
 micoscopy->microscopy
+Micosoft->Microsoft
 microprocesspr->microprocessor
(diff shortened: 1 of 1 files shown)


cfi-gb:
@peternewman Thanks for the hint, will to this tomorrow. Are there any guidelines for such upper/lowercase names? I just had followed the existing "microsoft" line in https://github.com/codespell-project/codespell/pull/734/files#diff-3bddc99f1f6dffc98ea11be5a5a31ae9R8277 which i havn't touched yet (but could do that in the scope of this PR as well).

peternewman:
Oops, yes please. The system doesn't do upper/lower properly yet, but the theory is to get the right data in the dictionary, then it "just works" when that gets enabled in future!

PR #735, opened 2018-10-29
Switch to raw string unicode rather than adding extra slashes

peternewman:
As per my comment here
https://github.com/codespell-project/codespell/pull/722#issuecomment-433771852


peternewman:
Hmm, WIP, needs more work, or a slightly different syntax. :disappointed:

PR #742, opened 2018-11-01
Add several spelling corrections + extra results for preexisting entries

luzpaz:
FYI, `wight` is a legit although archaic word. I kept it in there.


luzpaz:
thanks :wink:

PR #762, opened 2018-11-20
Added results to preexisting entries

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -7657,7 +7657,7 @@ irregardless->regardless
 irrelavent->irrelevant
 irrelevent->irrelevant
 irrelvant->irrelevant
-irreplacable->irreplaceable
+irreplacable->irreplaceable, irreparable,
```
larsoner: This does not look like a spelling error, but rather an English usage / word choice error. I don't think we should start adding those.

PR #787, opened 2019-01-02

arm-in:


peternewman:
I'm opening a PR to disable `byteorder->byte order` due to:
https://docs.oracle.com/javase/7/docs/api/java/nio/ByteOrder.html

PR #797, opened 2019-01-10
Add contine->continue, contain

cfi-gb:


luzpaz:
missing comma after `contain`

cfi-gb:
Ah, nice catch. Seems the CI caught it as well, fixed now.

PR #808, opened 2019-01-16
README: refines -S flag entry + add note about not using '/'

luzpaz:
@larsoner I'm having trouble with .rst formatting. How to I remove the bold from the example bullets ?

View the README: https://github.com/luzpaz/codespell/tree/README-refinement

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -53,13 +53,15 @@ The -S (--skip) flag is used to skip over comma-separated files and/or directori
 
 Examples:
 
-* Skip .eps & .txt files, invoke:
-    ``codespell --skip="*.eps,*.txt"`` OR
-    ``codespell -S *.eps,*.txt``
+* Skip .eps & .txt files, invoke one of::
 
-* Skip directories, invoke:
-    ``codespell --skip="./src/3rd-Party,./src/Test"`` OR
-    ``codespell -S /src/3rd-Party,./src/Test``
+    codespell --skip="*.eps,*.txt"
+    codespell -S *.eps,*.txt
+
+* Skip directories, invoke one of::
+
+      codespell --skip="./src/3rd-Party,./src/Test"
+      codespell -S /src/3rd-Party,./src/Test
 
 Useful commands::
 
--- a/README.rst
+++ b/README.rst
@@ -71,7 +71,7 @@ Useful commands::
 
 List all typos found except translation files and some directories.
 Display the results without terminal colors (``-d``). 
-The ``-q 3`` means we're running codespell with a quiet level of 3, which represents not to notify us that it's skipping binary files ::
+The ``-q 3`` means we're running codespell with a quiet level of 3, which represents not to notify us that it's skipping binary files. ::
 
     codespell -i 3 -w
 
--- a/README.rst
+++ b/README.rst
@@ -49,13 +49,17 @@ The -L flag can be used to whitelist certain words that are comma-separated plac
 
     codespell -S, --skip=
 
-Comma-separated list of files to skip. It accepts globs as well.  Examples:
+The -S (--skip) flag is used to skip over comma-separated files and/or directories. It accepts globs as well. **Important note:** do not append a '/' to a directory you want to skip. This currently renders the entry ineffective.
 
-* to skip .eps & .txt files, invoke ``codespell --skip="*.eps,*.txt"`` OR ``codespell -S *.eps,*.txt``
+Examples:
 
-* to skip directories, invoke ``codespell --skip="./src/3rd-Party,./src/Test"`` OR ``codespell -S /src/3rd-Party,./src/Test ::
+* Skip .eps & .txt files, invoke:
+    ``codespell --skip="*.eps,*.txt"`` OR
+    ``codespell -S *.eps,*.txt``
 
-   Note: do not append a '/' to a directory you want to skip. This currently renders the entry ineffective. 
+* Skip directories, invoke:
+    ``codespell --skip="./src/3rd-Party,./src/Test"`` OR
+    ``codespell -S /src/3rd-Party,./src/Test``
 
 Useful commands::
 
--- a/README.rst
+++ b/README.rst
@@ -53,11 +53,11 @@ The -S (--skip) flag is used to skip over comma-separated files and/or directori
 
 Examples:
 
-1. Skip .eps & .txt files, invoke:
+* Skip .eps & .txt files, invoke::
     ``codespell --skip="*.eps,*.txt"`` OR
     ``codespell -S *.eps,*.txt``
 
-2. Skip directories, invoke:
+* Skip directories, invoke::
     ``codespell --skip="./src/3rd-Party,./src/Test"`` OR
     ``codespell -S /src/3rd-Party,./src/Test``
 
--- a/README.rst
+++ b/README.rst
@@ -37,28 +37,30 @@ Some noteworthy flags::
 
     codespell -w, --write-changes
 
-The -w flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
+The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
 
     codespell -I FILE, --ignore-words=FILE
 
-The -I flag can be used to whitelist certain words that are in the ``codespell_lib/data/dictionary.txt``. The format of the whitelist file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said whitelist. **Important note:** The whitelist passed to ``-I`` is case-sensitive based on how it is listed in ``dictionary.txt``. ::
+The ``-I`` flag can be used to whitelist certain words that are in the ``codespell_lib/data/dictionary.txt``. The format of the whitelist file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said whitelist. **Important note:** The whitelist passed to ``-I`` is case-sensitive based on how it is listed in ``dictionary.txt``. ::
 
     codespell -L word1,word2,word3,word4
 
-The -L flag can be used to whitelist certain words that are comma-separated placed immediately after it. ::
+The ``-L`` flag can be used to whitelist certain words that are comma-separated placed immediately after it. ::
 
     codespell -S, --skip=
 
-The -S (--skip) flag is used to skip over a comma-separated list of files and/or directories. It accepts globs as well. **Important notes:** Do not append a '/' to a directory you want to skip. This renders the entry for that particular directory ineffective. Also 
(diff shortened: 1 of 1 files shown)


larsoner:
Your spacing is off so it's going to render like a definition list

larsoner:
See if that's more what you had in mind

README.rst
```diff
@@ -49,12 +49,19 @@ The -L flag can be used to whitelist certain words that are comma-separated plac
 
     codespell -S, --skip=
 
-Comma-separated list of files to skip. It accepts globs as well.  Examples:
+The -S (--skip) flag is used to skip over comma-separated files and/or directories. It accepts globs as well. **Important note:** do not append a '/' to a directory you want to skip. This currently renders the entry ineffective.
```
arm-in: I find the following passage misleading:
"skip over comma-separated files"
Suggestion:
"skip over comma-separated list of files"

README.rst
```diff
@@ -49,12 +49,19 @@ The -L flag can be used to whitelist certain words that are comma-separated plac
 
     codespell -S, --skip=
 
-Comma-separated list of files to skip. It accepts globs as well.  Examples:
+The -S (--skip) flag is used to skip over comma-separated list of files and/or directories. It accepts globs as well. **Important note:** do not append a '/' to a directory you want to skip. This currently renders the entry ineffective.
```
arm-in: Ooops! I forgot the article.
"skip over comma-separated"
should read
"skip a over comma-separated"

luzpaz: huh? `skip a over comma-separated` that makes no sense to me :confused:

larsoner: "skip over a comma-separated"

README.rst
```diff
@@ -49,12 +49,19 @@ The -L flag can be used to whitelist certain words that are comma-separated plac
 
     codespell -S, --skip=
 
-Comma-separated list of files to skip. It accepts globs as well.  Examples:
+The -S (--skip) flag is used to skip over a comma-separated list of files and/or directories. It accepts globs as well. **Important note:** Do not append a '/' to a directory you want to skip. This currently renders the entry ineffective.
```
peternewman: "This currently renders the entry ineffective." does this just affect that one entry in the list, or the whole list? It would help to clarify either way, perhaps something like "This renders the entry for that particular directory ineffective". I'd suggest the currently is redundant, if it's intended to fix it, linking to the issue might be more useful. Either way it's irrelevant unless someone remembers to update the docs, if you want to leave it in, adding a version number and/or date would be more helpful. Both your example directories begin with ./, if that's also required it's not mentioned and I'd suggest is much more important.

README.rst
```diff
@@ -49,12 +49,19 @@ The -L flag can be used to whitelist certain words that are comma-separated plac
...
 
-* to skip .eps & .txt files, invoke ``codespell --skip="*.eps,*.txt"``
+Examples:
 
-* to skip directories, invoke ``codespell --skip="./src/3rd-Party,./src/Test"`` ::
+* Skip .eps & .txt files, invoke one of::
 
+    codespell --skip="*.eps,*.txt"
+    codespell -S *.eps,*.txt
```
peternewman: I'm not sure listing the short and long version as an example adds much. Power users will understand the relationship between them without needing an example, and normal users just need something that works. Having a few more examples would be more helpful if anything (e.g. .txt files and all files in a particular folder).

README.rst
luzpaz: I've changed it based on your advice to `This renders the entry for that particular directory ineffective`.

re: the `./` in the examples.  
This is a quirk that I utilize in orrder to benefit form bash auto-completion but I'm not quite sure explaining that would be helpful as it would make this section overly-verbose. So I guess, removing this would be more appropriate.

README.rst
luzpaz: I didn't understand this at first but learned after usage and contributing to the project. I look at this as a way to turn amateurs in to power users. I vote to keep this.

luzpaz:
Made some refinements, please review

luzpaz:
Ready for merge unless someone has more feedback?

README.rst
```diff
@@ -49,16 +49,23 @@ The -L flag can be used to whitelist certain words that are comma-separated plac
 
     codespell -S, --skip=
 
-Comma-separated list of files to skip. It accepts globs as well.  Examples:
+The -S (--skip) flag is used to skip over a comma-separated list of files and/or directories. It accepts globs as well. **Important notes:** Do not append a '/' to a directory you want to skip. This renders the entry for that particular directory ineffective. Also
```
larsoner: There is a trailing `Also`

README.rst
peternewman: Personally I'd worry it just confuses them, especially given you've also changed the short mode syntax to drop the equals and quote. If you did the same thing with the long one, it's only four more characters to type "-kip" and it's more obvious what you're doing on the command line. Whereas I'd worry it's a lot more noise if each example is listed twice. If you really want to do it, personally I'd think a line at the end explaining how to use short arguments would be more effective (or probably link to somewhere else that does so already, as e.g. short flag chaining "grep -iIR" is where the benefit can really start to kick in. You haven't actually explained the other usage, you've just written something that looks similar, it wants a comment saying, this can also be done using the short flags as...

README.rst
peternewman: It wants to be "to skip, this renders". Assuming there aren't more words to go after the trailing also, then it should be "Important note:" (singular)

README.rst
```diff
@@ -49,16 +49,23 @@ The -L flag can be used to whitelist certain words that are comma-separated plac
...
+
+      codespell --skip="src/3rd-Party,src/Test"
+      codespell -S src/3rd-Party,src/Test
 
 Useful commands::
 
-    codespell -d -q 3 --skip="*.po,*.ts,./src/3rdParty,./src/Test"
+    codespell -d -q 3 --skip="*.po,*.ts,src/3rdParty,src/Test"
 
 | List all typos found except translation files and some directories.
```
peternewman: I would argue here again that if it was `codespell -d -q 3 --skip="*.po,*.ts,src/3rdParty,src/Test"` then the note about colours and quiet level becomes mostly irrelevant, or at least you could more usefully spend the words explaining what a quiet level 3 means (especially as it involves binary flags).

luzpaz:
Please review changes: https://github.com/luzpaz/codespell/tree/README-refinement

README.rst
luzpaz: Addressed in 90cf99f

README.rst
luzpaz: Addressed in 90cf99f

luzpaz: Addessed in 90cf99f

README.rst
```diff
@@ -37,35 +37,45 @@ Some noteworthy flags::
 
     codespell -w, --write-changes
 
-The -w flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
+The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
```
peternewman: Should be "same as doing a dry run"

peternewman: Why is the -i flag recommended? "... as this will prompt in ambiguous cases?"?

README.rst
```diff
@@ -37,35 +37,45 @@ Some noteworthy flags::
...
 
     codespell -I FILE, --ignore-words=FILE
 
-The -I flag can be used to whitelist certain words that are in the ``codespell_lib/data/dictionary.txt``. The format of the whitelist file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said whitelist. **Important note:** The whitelist passed to ``-I`` is case-sensitive based on how it is listed in ``dictionary.txt``. ::
+The ``-I`` flag can be used to whitelist certain words that are in the ``codespell_lib/data/dictionary.txt``. The format of the whitelist file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said whitelist. **Important note:** The whitelist passed to ``-I`` is case-sensitive based on how it is listed in ``dictionary.txt``. ::
```
peternewman: Given none of our matching is currently cases-sensitive, isn't this a bug we should fix rather than document?

peternewman: "execute codespell referencing said whitelist" might be better written as "execute codespell using the specified whitelist"

README.rst
```diff
@@ -37,35 +37,45 @@ Some noteworthy flags::
...
 
     codespell -L word1,word2,word3,word4
 
-The -L flag can be used to whitelist certain words that are comma-separated placed immediately after it. ::
+The ``-L`` flag can be used to whitelist certain words that are comma-separated placed immediately after it. ::
 
     codespell -S, --skip=
 
-Comma-separated list of files to skip. It accepts globs as well.  Examples:
+The ``-S`` (``--skip``) flag is used to skip over a comma-separated list of files and/or directories. It accepts globs as well. **Important note:** Do not append a ``/`` (forward-slash) to a directory you want to skip, this renders the entry for that particular directory ineffective. ::
```
peternewman: Ineffective how? I assume it matches the entire path, so if you were codespelling full paths from root, it wouldn't be ineffective?

README.rst
```diff
@@ -37,35 +37,45 @@ Some noteworthy flags::
...
 
-* to skip .eps & .txt files, invoke ``codespell --skip="*.eps,*.txt"``
+The ``-i <number>`` flag is used to invoke 'interactive mode' when writing changes. 0: no interactivity. 1: ask for confirmation. 2: ask user to choose one fix when more than one is available. 3: will combine both options 1 and 2 together.
```
peternewman: Move this up with the bit that recommends interactive mode (or add a note there saying see below?

README.rst
peternewman: I still personally disagree with cluttering the examples like this.

README.rst
```diff
@@ -37,35 +37,45 @@ Some noteworthy flags::
...
 
 List all typos found except translation files and some directories.
-Display them without terminal colors and with a quiet level of 3. ::
+Display the results without terminal colors (``-d``). 
+The ``-q 3`` means we're running codespell with a quiet level of 3, which represents not to notify us that it's skipping binary files. ::
 
     codespell -i 3 -w
 
-Run interactive mode level 3 and write changes to file.
+Run ``-i 3`` is interactive mode level 3 (see explanation above) and ``-w`` will write changes to file. **Note**: Omitting the ``-w`` flag will mean that your changes will not be saved!
```
peternewman: Again this feels like a bug, this system should at least log a message and perhaps just exit if you select interactive without write (I guess the only valid exception would be --dry-run).

README.rst
bl-ue:
```suggestion
The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag. ::
```

bl-ue:
At some point, I'll review the "bugs" that @peternewman pointed out in this PR, but for right now, I'll close this PR since it's over two years old, has conflicts, and may be out-of-date.

PR #820, opened 2019-01-30
Typos from VisualSharp Microsoft (part 3)

Fourchaux:
(Part 3 - Final...)
cf. https://github.com/Microsoft/visualfsharp/pull/2789/files


arm-in:
This should build now without error.

PR #821, opened 2019-02-04

arm-in:


arm-in:
I'm not sure how to interpret the error message:
"713 E               AssertionError: error u'lient' already exists"

Actually, the "lient" is from the latest commit on master, not from me.
750d3917236f5514317045b0cf10e1e5fa1df4e1

larsoner:
@pabs3 it looks like you are committing directly to `master`, and this has caused a failure. Could you open PRs from now on to add words? And also a PR to fix this error?

PR #829, opened 2019-02-10
Add surgestion(s)->suggestion(s)

luzpaz:


You95l:
Network all phone features and whatever you can think of

PR #832, opened 2019-02-13
Add opertions->operations

peternewman:


peternewman:
Failed randomly on 3.4, just needs the test re-running please.

PR #844, opened 2019-02-27

pabs3:
Also fix some confusion with American vs International English spelling.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -447,6 +447,8 @@ adresse->address
 adressed->addressed
 adresses->addresses
 adressing->addressing
+adresss->address
+adressses->addresses
 adrress->address
 adusted->adjusted
 advanatage->advantage
@@ -3810,6 +3812,8 @@ decendant->descendant
 decendants->descendants
 decendend->descendent, descendant, descended,
 decendent->descendent, descendant,
+decendentant->descendant
+decendentants->descendants
 decendents->descendents, descendants,
 deciaml->decimal
 decidate->dedicate
@@ -4240,6 +4244,8 @@ desintegrated->disintegrated
 desintegration->disintegration
 desipite->despite
 desireable->desirable
+desision->decision
+desisions->decisions
 desitned->destined
 deskop->desktop
 deskops->desktops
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -7830,12 +7841,15 @@ intiailizing->initializing
 intial->initial
 intialisation->initialisation
 intialise->initialise
-intialised->initialized
-intialises->initializes
+intialised->initialised
+intialiser->initialiser
+intialises->initialises
+intialises->initialises
```
larsoner: From Travis

u'intialises' already exists

pabs3:
> u'intialises' already exists

Yes but it is broken, it changes the misspelled International English
word into a correctly spelled but American English word. Same for the
other updates of similar words.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

codespell_lib/data/dictionary.txt
larsoner: That's not sure the problem -- it's complaining that there is now a duplicate entry. It's in there twice if you look at the diff (here).

larsoner: Lines 7846 and 7847 are identical in this case, delete one and you should be good

pabs3:
Woops forgot to run `make sort-dictionary` after fixing the commit.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

pabs3:
Sorting the dictionary also removes duplicates.

Anyway, updated the commit.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

larsoner:
Now Travis says

AssertionError: error diabler: correction disabler is an error itself

So if disabler is to be taken as valid, the correction for it should be removed

pabs3:
I found the diabler misspelling in the wild and it seems to indeed be a
misspelling for disabler not disabled. In addition disabler seems to be
a valid word so I'm inclined to drop the disabler->disabled check.
Thoughts?

https://codesearch.debian.net/search?q=diabler+package:gforth
https://www.thefreedictionary.com/disabler

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

larsoner:
Yeah I'm okay with removing `disabler` as an error

PR #848, opened 2019-03-02
Add spelling corrections for prepent->prepend

cfi-gb:
See e.g.

https://github.com/search?q=%22prependet%22&type=Code
https://github.com/search?q=%22prepented%22&type=Code

For https://github.com/search?q=%22prepent%22&type=Code there is e.g.:

https://www.urbandictionary.com/define.php?term=prepent

so not sure if this should be dropped?

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -10807,6 +10807,7 @@ prepaired->prepared
 prepartion->preparation
 prepartions->preparations
 prepatory->preparatory
+prepent->prepend, disabled due being a programming term
 preperation->preparation
 preperations->preparations
 prepresent->represent


peternewman:
Prepent does seem to be used as a programming term too:
https://github.com/lukexi/infinitybox/blob/56f383d09fd0b5bef4e89e70bf6ac7af7080e5c3/patches/kit/list-abs/prepent-help.pd#L13

You can always add with the "disabled due to..." which might be the best bet, as it's something people can, or will be able to tune out. There do appear to be genuine typos of it too.

cfi-gb:
Good idea, updated the PR accordingly.

codespell_lib/data/dictionary.txt
```diff
@@ -10556,8 +10556,16 @@ prepaired->prepared
...
 prepatory->preparatory
+prependet->prepended
+prepent->prepend, disabled due being a programming term
+prepented->prepended
 preperation->preparation
 preperations->preparations
+preppend->prepend
+preppended->prepended
+preppendet->prepended
+preppent->prepend
```
peternewman: This should probably suggest preprent then, although I don't know what Travis will think of that.

cfi-gb: What about using both, preprent and prepend like:

``
preppent->prepend, preprent,
``

peternewman: Yes sorry, I think that's what I meant. I think someone needs to decide if it having lots of use as a programming term means its a valid spelling, in which case we should treat it as a valid word and suggest it here as the closest spelling. Or if it's a typo primarily, we should suggest prepend as the closest valid word (although curiously Firefox dictionary doesn't like it!).

larsoner: I have never seen `prepent`, where does it get used?

larsoner: Oh, I see one example above. I'm not sure it's prevalent enough to justify the misses it'll cause, I'd rather make people whitelist it. If it's very common in some language I'm less familiar with (of which there are many -- mostly I'm Python/C/C++!) then I'm okay allowing it. FF dictionary not allowing also makes me think more that we shouldn't.

cfi-gb: I'm currently also unsure which way to go so i have split the other spelling corrections into a separate PR https://github.com/codespell-project/codespell/pull/863

larsoner:
I guess we don't need this, then?

cfi-gb:
@larsoner IIUC it could be still needed and its just open for discussion if:

```
prepent->prepend, disabled due being a programming term
```

or:

```
prepent->prepend
```

should be used.

larsoner:
Or neither, which is also an option. (That's the option I was going for.) People can still discuss, and if we decide on one or the other, reopen.

PR #865, opened 2019-03-12
Add notse->note(s)

cfi-gb:
See e.g. https://github.com/search?q=notse&type=Code

There seems some valid use like "NotSE" but in most cased this should be a spelling mistake.


larsoner:
Yes, agreed people should whitelist such sufficiently rare valid uses

PR #875, opened 2019-03-16
Import typos from mdadm history

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -3740,6 +3743,7 @@ coutners->counters
 covere->cover
 coveres->covers
 coverred->covered
+covert->convert
```
larsoner: covert is a word

PR #878, opened 2019-03-18
Add "inherets" to the dictionary

elopio:


codespell_lib/data/dictionary.txt
```diff
@@ -7713,6 +7713,7 @@ ingration->integration
 ingreediants->ingredients
 inhabitans->inhabitants
 inherantly->inherently
+inherets->inherits
```
luzpaz: Please also add `inheret->inherit`

PR #883, opened 2019-03-19

Gelma:


Gelma:
> looks good but there is a conflict

Fixed.

PR #888, opened 2019-03-19

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -3886,6 +3895,7 @@ dafaults->defaults
 dafaut->default
 dake->take
 dalmation->dalmatian
+dam->damn
```
larsoner: dam is fine

Gelma: rmeoved

PR #903, opened 2019-03-19

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -10176,6 +10177,7 @@ overwritren->overwritten
 overwrittes->overwrites
 overwrittin->overwriting
 overwritting->overwriting
+ovrridables->overridables
```
peternewman: Can you add the singular please

codespell_lib/data/dictionary.txt
```diff
@@ -11286,6 +11294,7 @@ propertly->properly, property,
 propertys->properties
 propertyst->properties
 propery->property, properly,
+properyvalue->propertyvalue
```
peternewman: This seems specific.

codespell_lib/data/dictionary.txt

codespell_lib/data/dictionary.txt
Gelma: Removed

larsoner:
Whenever you're happy feel free to merge. I think one set of eyes on each PR is probably good enough

peternewman:
> Whenever you're happy feel free to merge. I think one set of eyes on each PR is probably good enough

I can't guarantee I won't miss stuff/someone else won't spot other bits, but I guess we ought to be practical to get through this backlog, and users will presumably find it in the real world. 

Is this configured in GitHub too then, that PRs need to pass Travis and Appveyor and have an approved review?

larsoner:
No it's not configured that way, I always just look. But it might be possible to add this step if you want

peternewman:
> No it's not configured that way, I always just look. But it might be possible to add this step if you want

Yeah you can via protected branches ( https://help.github.com/en/articles/defining-the-mergeability-of-pull-requests ), we use it in another repo I work on, even for admins, and while it can occasionally be a small faff (and I've turned it off briefly for admins once or twice to get something through), I'd say it's generally a good idea. I think it also gives some UI hints to contributors to tell them Travis isn't working and it can't be merged.

larsoner:
Okay, I think I added the protection

peternewman:
> Okay, I think I added the protection

It looks like statuses are required, but not approving reviews, see e.g. this one: https://github.com/codespell-project/codespell/pull/1134

larsoner:
I didn't enable that one because it seems like unnecessary burden (extra clicks to start and approve a review) given the unlikelihood of errant merge

peternewman:
> I didn't enable that one because it seems like unnecessary burden (extra clicks to start and approve a review) given the unlikelihood of errant merge

Okay, fair enough.

peternewman:
Did you also tick "Require branches to be up to date before merging", we get a button to update on another repo, and will persuade people to resync?

peternewman:
Actually that may be irrelevant apart from new dictionary tests, as conflicts will still block.

larsoner:
Travis builds the merged version of each PR so it shouldn't be necessary

PR #909, opened 2019-03-19

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -1997,6 +1999,7 @@ braodcasted->broadcasted
 Brasillian->Brazilian
 brazeer->brassiere
 bre->be
+breaked->broken
```
luzpaz: break in past tense should be changed to `broke` not 'broken'

Gelma: Fixed. Removed the commit, and added the good one.

PR #938, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -5490,13 +5498,15 @@ enrties->entries
 enrty->entry
 ensconsed->ensconced
 entaglements->entanglements
+entent->intent
```
arm-in: also Diplomat slang
https://en.wikipedia.org/wiki/Entente

PR #956, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -2383,7 +2391,9 @@ challange->challenge
 challanged->challenged
 challanges->challenges
 challege->challenge
+champain->champagne
```
arm-in: also "champaign"

codespell_lib/data/dictionary.txt
```diff
@@ -2317,11 +2317,14 @@ ceated->created
...
 ceating->creating
 ceation->creation
+ceding->cedeing
 cehck->check
 cehcked->checked
 cehcking->checking
 cehcks->checks
 Celcius->Celsius
+cellabration->celebration
+celler->cellar
```
arm-in: also "seller"

codespell_lib/data/dictionary.txt
Gelma: the city?

arm-in: You are right, there is only the city. The drink is Champagne. I mixed it up.

PR #964, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -6601,6 +6605,8 @@ fysisist->physicist
 fysisit->physicist
 gadged->gadget, gauged,
 gae->game, Gael, gale,
+gage->gauge
```
arm-in: https://en.wikipedia.org/wiki/Gage
"variant spelling of the word gauge"

PR #974, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -7377,6 +7377,7 @@ imporing->importing
 importamt->important
 importat->important
 importd->imported
+importent->important
```
arm-in: also "impotent" ;-)

PR #975, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -9541,6 +9548,7 @@ inro->into
 ins't->isn't
 insallation->installation
 insalled->installed
+insance->instance
```
peternewman: Suggest insane too

PR #980, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8399,20 +8399,30 @@ keywork->keyword
...
 klicked->clicked
 klicks->clicks
 knifes->knives
 knive->knife
+knockous->noxious
+knowlegable->knowledgeable
 knowlegde->knowledge
 knowlege->knowledge
 knowlegeable->knowledgeable
+knowns->knows
```
arm-in: https://en.wikipedia.org/wiki/There_are_known_knowns

PR #981, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8430,8 +8437,11 @@ labled->labeled, labelled,
 lablels->labels
 lables->labels
 labling->labeling, labelling,
+labour->labor
```
arm-in: e.g. Labour Party (British English)

PR #984, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8521,9 +8521,13 @@ lentgh->length
 lentghs->lengths
 lenth->length
 lenths->lengths
+lepracan->leprechaun
+leprachan->leprechaun
+lepracy->leporsy
 leran->learn
 lerans->learns
 lesstiff->lesstif
+lest->least
```
arm-in: https://en.wikipedia.org/wiki/Lest_We_Forget

PR #985, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8615,12 +8620,14 @@ listner->listener
 listners->listeners
 liszt->list, liszt,
 litature->literature
+lite->light
```
arm-in: Tech speak, e.g. "SQLite", "Maglite", ...

PR #986, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8723,10 +8728,12 @@ magnatiude->magnitude
 magnatude->magnitude
 magnificient->magnificent
 magolia->magnolia
+mailinglist->mailing
```
arm-in: Also "mailing list"?

PR #987, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8811,6 +8816,7 @@ mantanine->maintain
 mantanined->maintained
 mantatory->mandatory
 mantenance->maintenance
+mantisses->mantisse
```
arm-in: plural of mantisse

PR #992, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -1386,9 +1391,13 @@ asthetical->aesthetical
 asthetically->aesthetically
 astiimate->estimate
 astiimation->estimation
+astrix->asterisk
```
arm-in: Asterix?

Gelma: added

PR #998, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -9077,11 +9078,15 @@ millenial->millennial
 millenialism->millennialism
 millenium->millennium
 millepede->millipede
+millimetre->millimeter
```
arm-in: British English OK

PR #1007, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -10064,6 +10071,7 @@ oterwice->otherwise
 oterwise->otherwise
 oterwize->otherwise
 othe->other
+other wise->otherwise
```
arm-in: whitespace

codespell_lib/data/dictionary.txt
```diff
@@ -10114,6 +10114,7 @@ ourselve->ourselves
 ourselvs->ourselves
 ouside->outside
 oustanding->outstanding
+out-of-bound->out-of-band
```
luzpaz: https://en.wikipedia.org/wiki/Out_of_bounds

codespell_lib/data/dictionary.txt
luzpaz: @Gelma the commit text needs to be amended as well 
image

codespell_lib/data/dictionary.txt
luzpaz: image

PR #1008, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -10226,6 +10229,7 @@ paintin->painting
 paitience->patience
 paiting->painting
 pakage->package
+pakageimpl->packageimpl
```
luzpaz: This is highly specialized. Thought I don't see a major problem having it in the dictionary.

PR #1016, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -11012,9 +11020,11 @@ presentaion->presentation
 presentaions->presentations
 presernt->present
 preserv->preserve
+preset->present
```
arm-in: exists

PR #1031, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -12939,8 +12939,11 @@ sheduled->scheduled
 shedules->schedules
 sheduling->scheduling
 sheild->shield
+sheme->scheme
+shepard->shepherd
```
arm-in: https://en.wikipedia.org/wiki/Shepard

Gelma:
@peternewman @larsoner : can you check this out, please?

Well, if it makes it easier for you, I could transport all my patches in a wiki page. Then, we can edit together all the entries, and when done, we can commit.

codespell_lib/data/dictionary.txt
larsoner: it's still part of this PR

Gelma: You are right.... uhm... maybe something wrong in the rebase?

Now I fix it and then check.

codespell_lib/data/dictionary.txt
```diff
@@ -16420,6 +16423,8 @@ shouls->should
 shoult->should
 shouw->show
 shouws->shows
+showfer->chauffeur
```
peternewman: Suggest shower

peternewman:
> Well, if it makes it easier for you, I could transport all my patches in a wiki page. Then, we can edit together all the entries, and when done, we can commit.

I think the workflow would still be almost as challenging. I don't know about @larsoner , but half my restriction is time rather than workflow anyway.

larsoner:
Yes time is the limitation. Though it would help to close all PRs that are not ready for review and reopen them when they are ready for review. And do this one or two at a time to avoid conflicts. Currently looking at the PR list we can't immediately know which are ready and which are not, we have to rely on emails (harder to track) or actually clicking through the PR to see if there is a conflict (cumbersome).

codespell_lib/data/dictionary.txt

Gelma:
> Yes time is the limitation. Though it would help to close all PRs that are not ready for review and reopen them when they are ready for review. And do this one or two at a time to avoid conflicts. Currently looking at the PR list we can't immediately know which are ready and which are not, we have to rely on emails (harder to track) or actually clicking through the PR to see if there is a conflict (cumbersome).

Last days I wrote a script that keep updated all the branches as soon as someone merge in master.

Picking randomly instead to go in order, for me is a mistake.

Anyway, if time is problem, I can't see advantages in closing all the PRs, open one, wait one month for merge, open another one, wait another month.

But, because you insist on this, I have no problem to close everything.

larsoner:
> Picking randomly instead to go in order, for me is a mistake.

You can go in any order you like, I'm not trying to control the order in which you want us to review ...

> Anyway, if time is problem, I can't see advantages in closing all the PRs, open one, wait one month for merge, open another one, wait another month.

... The problem I'm describing is the discoverability of the order for us maintainers. From my reviewing perspective, following email pings for when to review is much less convenient than always just clicking the "Pull Requests" tab on GitHub to see what needs to be done. So if you can indicate "I would like this to be reviewed and merged next" by having one or two PRs open, rather than email-pinging via one or two PR comments (with many many open PRs) it's much easier for us.

Gelma:
> You can go in any order you like, I'm not trying to control the order in which you want us to review ...

It's about the ordering in the merge/commit, not in the evaluation. Anyway, now my script manage this, so no problem.

> ... The problem I'm describing is the discoverability of the order for us maintainers. From my reviewing perspective, following email pings for when to review is much less convenient than always just clicking the "Pull Requests" tab on GitHub to see what needs to be done. So if you can indicate "I would like this to be reviewed and merged next" by having one or two PRs open, rather than email-pinging via one or two PR comments (with many many open PRs) it's much easier for us.

Well, I see two ways:
a) I create a webpage with links on the few working PR;
b) I can find a way to create one PR for every single line. This way, you can close immediately the good one.

If you find it manageable, just tell me.
Anyway, I don't want to bother the project, so I can close them and avoid my noise.

larsoner:
> Well, I see two ways:
> a) I create a webpage with links on the few working PR;

My proposal is to do this the standard way, i.e. with the GitHub Pull Requests page having the small number of PRs that are ready for feedback and merge. It has the lowest maintainer overhead. So 1) close existing PRs, 2) open one (maybe two) at a time, when one is merged, feel free to open another. Okay?

> b) I can find a way to create one PR for every single line. This way, you can close immediately the good one.

I'm not sure why you would need to do this. Your PRs can have ~10-100 changes in it and they are not so bad. One PR per changed line would actually be worse to review.

PR #1032, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -12997,7 +12997,10 @@ siezed->seized, sized,
 siezing->seizing, sizing,
 siezure->seizure
 siezures->seizures
+sift->shift
```
arm-in: sift = to use a sieve

codespell_lib/data/dictionary.txt
```diff
@@ -13030,8 +13033,14 @@ signto->sign to
 signul->signal
 signular->singular
 signularity->singularity
+siguret->cigarette
+silabus->syllabus
 silentely->silently
 silenty->silently
+sill->still
```
arm-in: exists

codespell_lib/data/dictionary.txt

codespell_lib/data/dictionary.txt

PR #1040, opened 2019-03-20

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -14105,6 +14107,7 @@ tanget->tangent
 tangetial->tangential
 tangetially->tangentially
 tangets->tangents
+tanks->thanks
```
arm-in: Plural of e.g. storage tank or battle tank

PR #1058, opened 2019-04-02
Add underlaying->underlying

Gelma:
From LibreOffice


codespell_lib/data/dictionary.txt
```diff
@@ -15194,6 +15195,7 @@ undefiend->undefined
 undeflows->underflows
 undefuned->undefined
 undependend->independent, nondependent,
+underlaying->underlying
```
larsoner: FYI underlay is a verb, so underlaying is okay, but I think it's going to be very rare

PR #1069, opened 2019-04-10
From LibreOffice today Git

Gelma:


Gelma:
@peternewman @larsoner 

Please, check this out. If it is good for you, merge it.
So, one by one, we can sort out all my queue.

Thanks a lot,
Gelma

Gelma:
Thanks a lot, I move to next one!

PR #1074, opened 2019-04-11
Add spelling fixes, primarily triple letters when double is correct, recuse->recurse and hmdi->hdmi  

warmsocks:
Fixes from some Linux-based embedded video-centric systems projects.


codespell_lib/data/dictionary.txt
```diff
@@ -12368,6 +12406,7 @@ recurively->recursively
 recurrance->recurrence
 recursivion->recursion
 recursivly->recursively
+recuse->recuse, recurse,
```
luzpaz: this will return false positives but that's OK IMO

codespell_lib/data/dictionary.txt
```diff
@@ -10935,6 +10964,7 @@ parsin->parsing
 parstree->parse tree
 partaining->pertaining
 partialy->partially
+partical->partial
```
luzpaz: can also be `particle`

codespell_lib/data/dictionary.txt
```diff
@@ -5098,6 +5117,7 @@ discourrage->discourage
 discourraged->discouraged
 discovereability->discoverability
 discovr->discover
+discreet->discrete, discreet,
```
luzpaz: will return false positives but since this is a common misspelling LGTM

codespell_lib/data/dictionary.txt
warmsocks: @luzpaz Good point.  Fixed.

codespell_lib/data/dictionary.txt
warmsocks: @luzpaz Agreed.  I found several occurrences of `recuse` in the codebases I was correcting, and they were all misspellings of recurse.  I was tempted to make it the first option since `recurse` is used much more frequently in code, but decided since it is a real word to leave it as the first option.  If you think they should be switched, PLMK.

codespell_lib/data/dictionary.txt
warmsocks: Agreed.

codespell_lib/data/dictionary.txt
```diff
@@ -10964,7 +10964,7 @@ parsin->parsing
 parstree->parse tree
 partaining->pertaining
 partialy->partially
-partical->partial
+partical->partial, particle,
```
luzpaz: remove the trailing whitespace :wink:

Makefile
```diff
@@ -24,7 +24,7 @@ sort-dictionary:
 	LC_ALL=C sort ${SORT_ARGS} -u -o ${DICTIONARY} ${DICTIONARY}
 
 trim-dictionary:
-	sed -E -i.bak -e 's/^[[:space:]]+//; s/[[:space:]]+$$//; /^$$/d' ${DICTIONARY} && rm ${DICTIONARY}.bak
```
peternewman: What's wrong with just sed -E -i -e ...?
https://linux.die.net/man/1/sed

codespell_lib/data/dictionary.txt
peternewman: I don't know what @larsoner thinks, but personally I feel there shouldn't be valid spellings in the primary dictionary. To take a not hugely contrived example, `passwords must be stored in a discreet way with each one in a discrete file`.

codespell_lib/data/dictionary.txt
```diff
@@ -11114,6 +11141,7 @@ neeeded->needed
 neeeding->needing
 neeedn't->needn't
 neeeds->needs
+neees->needs
```
peternewman: Add knees.

codespell_lib/data/dictionary.txt
peternewman: Again, personally I don't think we should have it in the main file

Makefile
warmsocks: > What's wrong with just sed -E -i -e ...?

That won't work on MacOS.  Mac has a BSD-derived sed, while Linux distros use the GNU version derived from ATT.   There seems to be no portable way to do sed in-place editing without making backup files.  Mac users could install gnu sed, but that doesn't seem like a reasonable dependency.

The change in commit 2d83ab7 switches from sed to ex, which should work on both Linux and MacOS (and others).  It does in-place editing properly, not creating a backup file in the first place.

peternewman: Ah yes, now you mention it, I think I've hit this myself somewhere.

You think `ex` is preferable or more cross-platform to `perl -pie`? I don't know, but I suspect `perl -pie` won't fail on no hits. I'd be rather wary about running in quiet mode AND having `|| true` on the end, that still feels like a recipe for trouble.

Otherwise is $$ or $RANDOM cross-platform enough to use as the extension with sed?

warmsocks: @peternewman `$$` and `$RANDOM` are definitely cross-platform enough.  I think they would be fine in practice, giving a very small chance of clobbering an existing backup file, but as long as I was going to fix the issue, I thought I would do it right, i.e. by not specifying one at all, completely eliminating the risk. This also parallels the behavior of the `sort-dictionary` target.

I don't know Perl well enough to say whether `perl -pie` is preferable to `ex`, or which versions of `perl` are installed by default on which versions of different OSes.  The Perl substitution command would be simple enough, but (as usual) it is the error handling and rare conditions that are the tricky bit, and where I'm unfamiliar with any `perl` pecularities.  `ex` is specified by POSIX, and is part of the Single Unix Specification, so it's common and uniform.  Most importantly, I'm familiar with it, so it is what I went with.   Perl may well be a better option; if you are comfortable enough with it, please feel free to submit a separate PR for the `Makefile` that uses it instead.  I won't be offended, and the maintainer can choose what he/she deems best.

Ask `n` engineers to solve a problem, and you'll get `3*n` different solutions, on average.  :-)

luzpaz:
This PR needs to be split up. One should be a pure dictionary diff and the other is the makefile trim-dictionary proposal. @warmsocks are you OK with that?

warmsocks:
> This PR needs to be split up. One should be a pure dictionary diff and the other is the makefile trim-dictionary proposal. @warmsocks are you OK with that?

@luzpaz Yes, absolutely.

codespell_lib/data/dictionary.txt
```diff
@@ -15582,6 +15610,7 @@ probabaly->probably
 probabilaty->probability
 probabilty->probability
 probabily->probability
+probabily->probability, probably,
```
luzpaz: dupe

luzpaz:
this PR is very cluttered. hard to comprehend where we're at with it

codespell_lib/data/dictionary.txt
```diff
@@ -19633,6 +19670,8 @@ suppported->supported
 suppporting->supporting
 suppports->supports
 suppres->suppress
+suppresion->suppression
+suppresions->suppression
```
c72578: Missing plural s at the end. Should be:
`suppresions->suppressions`

codespell_lib/data/dictionary.txt
peternewman: This correction should now be in rare if it goes in at all:
https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary_rare.txt

luzpaz:
@warmsocks please split PR

bl-ue:
Resolved conflicts.  I removed the `ex` stuff and went back to the old `sed` solution because, at least currently, it's working for me on macOS _without_ GNU `sed` (i.e. _with_ BSD `sed`).

Makefile
bl-ue: I removed the `ex` stuff and went back to the old `sed` solution because, at least currently, it's working for me on macOS _without_ GNU `sed` (i.e. _with_ BSD `sed`).

bl-ue (approved):

MarcoGorelli:
recuse is a real word https://www.merriam-webster.com/dictionary/recuse

would you consider a PR to remove the `recuse-recurse` rewrite?

PR #1080, opened 2019-04-13
Typos history from electron

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -8086,6 +8109,7 @@ ilogical->illogical
 ilustrate->illustrate
 ilustrated->illustrated
 ilustration->illustration
+imag->image
```
larsoner: This is likely to lead to a lot of false alarms as this is a shorthand for `imaginary` in mathematical domains

Issue #1088, opened 2019-04-19
Blank lines in dictionary aren't flagged by check-dictionary, cause crash


warmsocks:
Seen in codespell 1.15.0.dev0

If a blank line is introduced into the dictionary, `make check-dictionary` doesn't flag it as a problem, but when codespell is invoked, it crashes:

```
$ codespell
Traceback (most recent call last):
  File "/usr/bin/codespell", line 9, in <module>
    load_entry_point('codespell==1.15.0.dev0', 'console_scripts', 'codespell')()
  File "/usr/lib/python2.7/site-packages/codespell_lib/_codespell.py", line 579, in _script_main
    return main(*sys.argv[1:])
  File "/usr/lib/python2.7/site-packages/codespell_lib/_codespell.py", line 625, in main
    build_dict(dictionary, misspellings, ignore_words)
  File "/usr/lib/python2.7/site-packages/codespell_lib/_codespell.py", line 316, in build_dict
    [key, data] = line.split('->')
ValueError: need more than 1 value to unpack
```
I would expect 

1. The check-dictionary target to flag blank lines as an error
2. A make target that would strip blank lines out, along with any leading or trailing whitespace.

I have a Makefile patch ready; I'll submit a PR.

warmsocks:
#1089

PR #1089, opened 2019-04-19
Fix Blank lines in dictionary aren't flagged by check-dictionary, cause crash #1088

warmsocks:
The -b argument added to sort ignores leading whitespace, so sort and trim can be done in either order.  Without it, if sorting is done first, all lines with leading whitespace are bunched together, so after the whitespace is removed sort needs to be done a second time. 

-b sorts everything into the proper order regardless of leading whitespace, so it is still in the correct order after trimming.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
-SORT_ARGS := -f
+SORT_ARGS := -f -b
 
-PHONY := all check check-dictionary sort-dictionary clean
+DICTIONARY := codespell_lib/data/dictionary.txt
+
+PHONY := all check check-dictionary sort-dictionary trim-dictionary clean
 
 all: check-dictionary codespell.1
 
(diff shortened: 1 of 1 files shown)


larsoner:
Thanks @warmsocks

Makefile
```diff
@@ -9,13 +11,20 @@ codespell.1: codespell.1.include bin/codespell
...
+		echo "Dictionary contains leading/trailing whitespace and/or blank lines.  Trim with 'make trim-dictionary'"; \
+		exit 1; \
+	fi
 
 sort-dictionary:
-	LC_ALL=C sort ${SORT_ARGS} -u -o codespell_lib/data/dictionary.txt codespell_lib/data/dictionary.txt
+	LC_ALL=C sort ${SORT_ARGS} -u -o ${DICTIONARY} ${DICTIONARY}
+
+trim-dictionary:
+	sed -E -i.bak -e 's/^[[:space:]]+//; s/[[:space:]]+$$//; /^$$/d' ${DICTIONARY} && rm ${DICTIONARY}.bak
```
peternewman: I don't know about you, but I regularly use .bak as a backup extension when hacking around with stuff, won't sed overwrite that file, and the rm remove it. Something more unique, like adding the PID would be safer @warmsocks .

warmsocks: I'll submit a PR that uses ex instead of sed, so no tempfile is required at all.

PR #1090, opened 2019-04-20
Add escpae->escape and escpaed->escaped

EdwardBetts:


larsoner:
Can you fix the merge conflict?

codespell_lib/data/dictionary.txt
```diff
@@ -5978,6 +5978,8 @@ esacpe->escape
 esacpes->escapes
 escapeable->escapable
 escapemant->escapement
+escpae->escape
+escpaed->escaped
```
luzpaz: not highly likely but may also be `escapade` (link)](https://www.merriam-webster.com/dictionary/escapade)

larsoner:
Needs merge/rebase

codespell_lib/data/dictionary.txt
larsoner: I think this is rare enough we should be okay either way

PR #1094, opened 2019-04-29
Added spelling corrections for impact.

cfi-gb:


cfi-gb:
@luzpaz Good idea, thanks. Added in 83c6ce8cf896fb97d0553fa68828bab6973920f0

PR #1099, opened 2019-04-30
Typos from Perl5 (2012-2013)

Fourchaux:
Some typos from https://perl5.git.perl.org/perl.git/search/HEAD?s=typo;st=free


codespell_lib/data/dictionary.txt
```diff
@@ -9318,6 +9358,7 @@ istead->instead
 istener->listeners
 ists->its, lists,
 isue->issue
+isused->is used
```
luzpaz: I think `issued` should be the first result here.

Fourchaux: Right. Fixed.

PR #1106, opened 2019-05-07
Typos from Perl5 (2014-2019) Grand Finale

Fourchaux:
Some typos from https://perl5.git.perl.org/perl.git/search/HEAD?s=typo;st=free

**Note:**

I would like to thank the Perl5 community for all these misspellings (800+ between 1989 and 2019!)
**AS WELL AS** their corrections.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -672,6 +672,8 @@ alghoritm->algorithm
 alghoritmically->algorithmically
 alghoritms->algorithms
 algined->aligned
+alginment->alignment
+alginments->alignments
 algorhitm->algorithm
 algorhitms->algorithms
 algorightm->algorithm
@@ -1044,6 +1046,7 @@ anticpate->anticipate
 anual->annual
 anually->annually
 anulled->annulled
+anumber->a number
 anway->anyway
 anways->anyway
 anwser->answer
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -11180,6 +11207,7 @@ obtaiend->obtained
 obtaion->obtain
 obtrain->obtain
 obtrained->obtained
+obverse->observe
```
larsoner: This is a word

codespell_lib/data/dictionary.txt
```diff
@@ -11931,6 +11965,9 @@ paticular->particular
 paticularly->particularly
 patrent->parent, patent, patron,
 pattented->patented
+patter->pattern
```
larsoner: This is a word, but maybe we should keep since there will presumably be way more true positives than false positives from including this

peternewman: It feels like a slippery slope to me; if you're going to do that, can they at least be pulled into a separate dictionary file instead and add the ability to use multiple dictionaries (if not already present). Then there's just the challenge of deciding what typos are sufficiently popular to be in the second file.

codespell_lib/data/dictionary.txt
Fourchaux: OK. Deleted.

codespell_lib/data/dictionary.txt
Fourchaux: What should I do? 
Delete the word?

larsoner: Yeah

Fourchaux: Okay. 
'patter' & 'patters' deleted.

PR #1116, opened 2019-05-13

pabs3:


codespell_lib/data/dictionary.txt
```diff
@@ -1408,7 +1417,12 @@ arrtibute->attribute
 arry->array, carry,
 arrys->arrays
 artcile->article
+articaft->artefact
+articafts->artefacts
```
larsoner: Seems like these should have the American spelling (maybe only since these seems like attempts at the American spelling?)

codespell_lib/data/dictionary.txt
```diff
@@ -1408,7 +1417,12 @@ arrtibute->attribute
...
 artical->article
+articals->articles
+articat->artefact
+articats->artefacts
```
larsoner: same

Issue #712, opened 2018-10-16
Readd(ed) corrects to itself


cfi-gb:
The dictionary.txt currently contains the following:

```
readd->readd, re-add, read,
readded->readded, read,
```

which corrects to itself:

```
$ cat test.txt
readd
readded

$ codespell test.txt
test.txt:1: readd  ==> readd, re-add, read
test.txt:2: readded  ==> readded, read
```

If the correct spelling of "readded" would be "re-added" then there is also the following found a few lines below:

```
readed->read, readd, readded,
```

larsoner:
Related discussion in #410, it's not clear how to deal with this case yet

peternewman:
As I mentioned in #410, I think something needs to be done. Personally if things like `wit->wit, with,` are being included, these ones need to be excluded by default. Only some sort of magic context based AI can decide if you wanted wit or with in a particular sentence.

larsoner:
Maybe what we should do is have support for multiple dictionary files, which will use the union of all entries (probably just taking the entries from ones listed later if there are overlaps). Then the default can be `dictionary.txt` and `dictionary_rare.txt` or so, where the latter holds all of these self-corrections that are likely mistakes but could be correct.

peternewman:
> Maybe what we should do is have support for multiple dictionary files, which will use the union of all entries (probably just taking the entries from ones listed later if there are overlaps). Then the default can be `dictionary.txt` and `dictionary_rare.txt` or so, where the latter holds all of these self-corrections that are likely mistakes but could be correct.

Sounds good to me, multiple files should give some much better options for validation during CI. The only real question is how they get merged or overwritten regards suggestions etc.

larsoner:
> The only real question is how they get merged or overwritten regards suggestions etc.

I would try to keep it simple and let the later entries take precendence (overwriting any exitsing entries). From a user perspective it's clear what happens, and if we need some other behavior let's add it later and not over-engineer (assume YAGNI for now).

@peternewman do you want to try to implement this?

peternewman:
Yes, I guess I should given I keep banging on about it! I've got some other stuff I need to sort first, but I'll try and take a look at some point soon. In the meantime, I guess people can start preparing additional dictionaries...

peternewman:
I think I've unfortunately found a scenario where suggestions may need merging (or maybe I just need a working camel case filter).

[...]

PR #1118, opened 2019-05-16
Add spelling corrections for "construct" variants.

cfi-gb:
See https://github.com/search?q=consrtuct&type=Code

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3890,6 +3890,11 @@ consonents->consonants
 consorcium->consortium
 conspiracys->conspiracies
 conspiriator->conspirator
+consrtuct->construct
+consrtucted->constructed
+consrtuctor->constructor
+consrtuctors->constructors
+consrtucts->constructs
 consructor->constructor
 constained->constrained
 constains->constrains


codespell_lib/data/dictionary.txt
```diff
@@ -17826,6 +17831,7 @@ varients->variants
 varierty->variety
 variey->variety
 varified->verified
+varify->verify
```
peternewman: Seems this is also a real word:
https://www.dictionary.com/browse/varify

cfi-gb: Mhhh, neither Mozilla/Firefox nor LibreOffice spelling corrections are accepting this as a correct word.

There seems to be also already a "varifying->verifying" in the dictionary.txt one line below. 

On the other hand https://en.wiktionary.org/wiki/varify is available as well, so any opinions to remove varifying->verifying instead?

peternewman: Probably also `varified->verified`.

I accept it's a rarely used word, but it is a valid word. See also discussion in #712 .

cfi-gb: I have removed the ``varify->verify`` from this PR again and created a separate PR #1127 for the other variants.

PR #1123, opened 2019-05-18
Removing 'webp' - reason: too many false positives

luzpaz:


gpotter2:
It would be nice if we could have a release with this fix soon 😄 
Thanks for this great project !

PR #1129, opened 2019-05-22
Add spelling corrections

Fourchaux:


codespell_lib/data/dictionary.txt
```diff
@@ -14261,7 +14266,7 @@ rennovation->renovation
 rentime->runtime
 rentors->renters
 reoccurrence->recurrence
-reocurring->recurring
+reocurring->recurring, reoccurring,
```
luzpaz: Nice catch. I think `reoccurring` should be the first result

codespell_lib/data/dictionary.txt
```diff
@@ -1558,7 +1560,8 @@ assumme->assume
 assummed->assumed
 assummes->assumes
 assumpation->assumption
-assumtions->assumption
```
luzpaz: Nice catch.

codespell_lib/data/dictionary.txt
Fourchaux: OK. Done.

PR #1132, opened 2019-05-24
Add directort -> directory

jcfr:
Discovered in auditwheel package. See https://github.com/pypa/auditwheel/pull/165


jcfr:
Thanks for integrating :+1:

PR #1134, opened 2019-05-25
Add several spelling corrections and refinements

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -12201,7 +12201,7 @@ overwite->overwrite
 overwites->overwrites
 overwitten->overwritten
 overwize->otherwise
-overwridden->overwritten
+overwridden->overwritten, overridden,
```
peternewman: I think overridden should be the first suggestion!

codespell_lib/data/dictionary.txt
```diff
@@ -16355,6 +16355,8 @@ subsytem->subsystem
 subsytems->subsystems
 subtabels->subtables
 subtances->substances
+subtarger->subtarget
```
peternewman: Perhaps also add the hyphenated variants of this and the plural?

luzpaz: 2982c26

codespell_lib/data/dictionary.txt
luzpaz: 2982c26

PR #898, opened 2019-03-19
Misc queue from 2018 (sorry, lost sources)

Gelma:

PR #1138, opened 2019-05-28
Check for self-replacements

peternewman:
i.e. foo->foo


peternewman:
As found in https://github.com/codespell-project/codespell/pull/898/files#diff-3bddc99f1f6dffc98ea11be5a5a31ae9R5740 .

larsoner:
Thanks @peternewman

PR #1139, opened 2019-05-28

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -6871,7 +6880,7 @@ entired->entered, entire,
 entireity->entirety
 entires->entries
 entirey->entirely
-entirity->entirely
+entirity->entirety, entirely,
```
peternewman: I'd agree with your suggestion and be tempted to drop entirely, it seems like it was influenced by the ones either side of it.

codespell_lib/data/dictionary.txt
```diff
@@ -17144,6 +17162,8 @@ sytntax->syntax
 szenario->scenario
 szenarios->scenarios
 szes->sizes
+szie->size
```
peternewman: Add szied equivalent too.

codespell_lib/data/dictionary.txt
luzpaz: I'll drop it

luzpaz:
If this is approved please squash it down to one commit. If you want me to to so...then I'd be happy to.

luzpaz:
I added 72c096d last minute. Please also review it. Thanks!

peternewman:
> If this is approved please squash it down to one commit. If you want me to to so...then I'd be happy to.

Squashed and merged as requested. Certainly I've not generally bothered with other commits.

PR #1150, opened 2019-06-06
Add variations on iml[i,e]ment->implement

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -9065,9 +9065,18 @@ imigrated->emigrated, immigrated,
 imigration->emigration, immigration,
 imilar->similar
 iminent->eminent, imminent, immanent,
```
peternewman: I know it's not your change, but can we swap this so imminent is the first suggestion, given it's only one letter missing.

codespell_lib/data/dictionary.txt
```diff
@@ -9065,9 +9065,18 @@ imigrated->emigrated, immigrated,
 imigration->emigration, immigration,
```
peternewman: Likewise here, I assume they've got their e/i right and more likely to have missed a letter.

luzpaz: 0ce5399

codespell_lib/data/dictionary.txt
luzpaz: 0ce5399

PR #1153, opened 2019-06-08
New typo (found in llvm) unavaliable->unavailable

sylvestre:


sylvestre:
see: https://reviews.llvm.org/rG375297f38fe8

PR #1155, opened 2019-06-11
Remove bodgy->body as it's a valid word

peternewman:
https://www.lexico.com/en/definition/bodgie

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -2506,6 +2506,7 @@ boads->boards
 boaut->bout, boat, about,
 bobard->board, bombard,
 boddy->body
+bodgy->body, disabled because one might want to allow informal spelling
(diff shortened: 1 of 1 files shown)


luzpaz:
on the fence about this one.

peternewman:
> on the fence about this one.

How so, best case it should be disabled the same way as gonna is?

See also:
https://www.lexico.com/en/definition/bodge and https://www.collinsdictionary.com/dictionary/english/bodge

peternewman:
Thoughts @larsoner ?

larsoner:
Maybe for now disable how `gonna` is disabled until we get around to separating dictionaries into strict and loose?

peternewman:
> Maybe for now disable how `gonna` is disabled until we get around to separating dictionaries into strict and loose?

Done

peternewman:
Thanks @larsoner .

PR #1158, opened 2019-06-13
Added variation on auto-detet typo

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -2061,6 +2061,9 @@ autimatically->automatically
 autmatically->automatically
 auto-dependancies->auto-dependencies
 auto-destrcut->auto-destruct
+auto-detet->auto-detect
+auto-deteted->auto-detected
```
peternewman: At least this and detection should also suggest deletion variants.

luzpaz: please elaborate

peternewman: Sorry `auto-deteted->auto-detected, auto-deleted,`

codespell_lib/data/dictionary.txt
```diff
@@ -2061,6 +2061,11 @@ autimatically->automatically
 autmatically->automatically
 auto-dependancies->auto-dependencies
 auto-destrcut->auto-destruct
+auto-detet->auto-detect, auto-delete,
+auto-deteted->auto-detected, auto-deleted,
+auto-detets->auto-detects, auto-deletes,
+auto-detetes->auto-detects, auto-deletes,
```
peternewman: I wonder if this one should have deletes before detects as the suggestions.

luzpaz: @peternewman i found this typos intending to be `auto-detect`. IMO we should keep it this way for now unless we get reports showing the contrary...?

luzpaz:
FIxed travis error by sorting dictionary (40f7894)

codespell_lib/data/dictionary.txt
peternewman: Which specific variant @luzpaz ? I'm only suggesting swapping the order of the suggestions on `auto-detetes->auto-detects, auto-deletes,`, given it's one mistyped letter (t/l), rather than a missing letter (c) and an extra letter (e).

luzpaz: Ok, I'll acquiesce. Of course it's not a big deal. IMO 'detect' is the higher probable typo.  Made the change in 6ecd300
While I was at it, I also added f0afbc9

peternewman: Can you put back most of those please. It was only the specific `detetes` variants where I feel the other one is closer, not every single variant.

luzpaz: oh, i misunderstood.

luzpaz: @peternewman im confused, can you make the change instead :stuck_out_tongue:  
I just don't want to repeatedly make mistakes that need to be revised again.

peternewman: I've done it but made a duplicate entry I need to delete. Essentially my thinking is missing letter, substitution, additional letter as order of likelihood.

luzpaz:
LGTM!

PR #1161, opened 2019-06-16
Remove 'parameterset->parameter set' ; too many false positives

luzpaz:
Because of codespell's lack of differentiating camelcase functions names in conjunction with this being a common programming term - I'm proposing to delete this entry from the dictionary.


peternewman:
Agreed, I just got hit by ErrorString, which I think you or someone else has subsequently removed again.

peternewman:
I guess the only question is should they be left in as "Disabled due to too many false positives to stop someone adding them back in again next time?

luzpaz:
Or we could have a list of words that we restrict when someone performs a PR (this would help with british/american spelling issues as well

PR #1163, opened 2019-06-16

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -2552,6 +2552,7 @@ bottlneck->bottleneck
 bottomborde->bottomborder
 bottome->bottom
 botton->bottom, button,
+bottons->buttons
```
peternewman: Given the above suggests bottom, I think this should suggest bottoms too.

codespell_lib/data/dictionary.txt
```diff
@@ -6215,6 +6222,8 @@ dirctly->directly
 dirctories->directories
 dirctory->directory
 direccion->direction
+direcctory->directory
```
peternewman: It would probably be good to add `direcctorys->directories` too.

codespell_lib/data/dictionary.txt
```diff
@@ -8933,8 +8945,13 @@ halp->help
 halpoints->halfpoints
 hammmer->hammer
 hampster->hamster
+handel->handle
```
peternewman: The composer? https://en.wikipedia.org/wiki/George_Frideric_Handel

codespell_lib/data/dictionary.txt
```diff
@@ -8933,8 +8945,13 @@ halp->help
...
 handeld->handled, handheld,
 handeled->handled, handheld,
+handeler->handler
+handeles->handles
+handeling->handling
+handels->handles
```
peternewman: His music (Handel's)?

codespell_lib/data/dictionary.txt
luzpaz: we're not doing names AFAIR, so I think this is valid

codespell_lib/data/dictionary.txt
luzpaz: This is a usecase for using the `--ignore` flag.

codespell_lib/data/dictionary.txt
peternewman: Okay fair enough.

codespell_lib/data/dictionary.txt
```diff
@@ -19118,6 +19142,15 @@ unsable->unusable, usable, unstable,
...
+unselcted->unselected
+unselctes->unselects
+unselcting->unselecting
+unselcts->unselects
+unselecgt->unselect
+unselecgted->unselected
+unselecgtes->unselects
+unselecgting->unselecting
+unselecgts->unselects
 unselect->deselect
```
peternewman: Travis is failing because of this. See also here:
https://english.stackexchange.com/questions/18465/unselect-or-deselect

luzpaz: I see this come up very often and i have learned to ignore it. Not sure what to do here.

peternewman: Well we either need to remove `unselect->deselect` as a correction, or make all the ones that suggest unselect actually suggest deselect.

luzpaz: I'd say get rid of `unselect->deselect` since it's context sensitive grammar. All the others are typos.

peternewman: Sounds good to me, do you want to do the honours. That thread is a bit vague either way, but there seem to be enough instances of it, and reasons for it, that it doesn't seem unreasonable.

PR #1172, opened 2019-06-25
Add failsave->failsafe

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -8078,6 +8078,7 @@ failes->fails
 faill->fail
 failled->failed
 failre->failure
+failsave->failsafe
```
peternewman: Wikipedia seems conflicted, but fail-safe looks to be the most used version, so probably suggest that first, and then maybe the non-hyphenated one.

luzpaz: be1b33e

luzpaz:
pluralized in 93b92b9

Issue #737, opened 2018-10-30
Add note on the capitalisation of words in dictionary to readme


cfi-gb:
Hi *,

based on https://github.com/codespell-project/codespell/pull/734#issuecomment-434089546 it would be great to add a short note about that for new contributors to e.g.:

https://github.com/codespell-project/codespell/blob/master/README.rst#dictionary-format

or

https://github.com/codespell-project/codespell/blob/master/README.rst#sending-pull-requests

larsoner:
Feel free to make a PR to add something like this

https://github.com/mne-tools/mne-python/blob/master/.github/CONTRIBUTING.md

larsoner:
Actually we should probably move the sending-pull-requests section of `README.rst` into a new `CONTRIBUTING.md`; then also update the README.rst about the dict format.

cfi-gb:
Hi,

and thanks for your replies.

> Feel free to make a PR to add something like this

Creating a PR for something like this is out of my knowledge as i don't know the rules which applies for this repository. That's why i had created the issue here instead of sending a PR directly.

But a CONTRIBUTING.md as well as a PULL_REQUEST_TEMPLATE.md could be a real good improvement for new contributors. :+1:

PR #1178, opened 2019-07-02
Add vulnerabuility and variants

cfi-gb:
See e.g.:

https://github.com/search?q=Vulnerabuility&type=Code
https://github.com/search?q=%22vulnerabule%22&type=Code

I took the chance to make all existing "vulnerabilit(y|ies)" corrections uppercase.


luzpaz:
> I took the chance to make all existing "vulnerabilit(y|ies)" corrections uppercase.

Please don't. `codespell` corrects words in the case it finds them. There are plenty of usecases where they don't need to capitalized and it could be very irritating that codespell makes that decision arbitrarily. Please revert the uppercasing.

cfi-gb:
@luzpaz Thanks for the remark. This seems to me somehow in contradiction to what @peternewman had requested in https://github.com/codespell-project/codespell/pull/734 in the past.

That's why i had created https://github.com/codespell-project/codespell/issues/737 to have a fixed guideline on this topic.

luzpaz:
either myself or you have misunderstood something here. `Microsoft` is a name and therefore falls in to that gray area of capitalization (I say gray because sometimes people don't want such words capitalized but nvm). You current PR for variations on the word 'vulnerable' doesn't fit that naming classification, AFAICT. Unless I'm missing something

cfi-gb:
> Microsoft is a name and therefore falls in to that gray area of capitalization (I say gray because sometimes people don't want such words capitalized but nvm)

Understood. So that could be perfectly as part of the guideline requested in https://github.com/codespell-project/codespell/issues/737.

cfi-gb:
Reverted the capitalization in 5561d1cb4e63fe82fb2ef732940fbdfd96e6fc4e

peternewman:
Yeah I'd only asked for Microsoft to be capitalised because it's a name. If we have that bit of the dictionary correct, we can in future add an option to use or ignore the case of the dictionary (depending on if you're writing e.g. code or documentation.

PR #1189, opened 2019-07-15
Remove 'byteorder'

luzpaz:
Too many false positives.


luzpaz:
@peternewman please review

PR #1198, opened 2019-07-20
Remove heuristical->heuristic

luzpaz:
`heuristical` is a [legit](https://www.lexico.com/en/definition/heuristical) word.


luzpaz:
@larsoner please merge

PR #1199, opened 2019-07-21
Add wee->we

luzpaz:


luzpaz:
Not sure why Travis is failing: 
```
  File "/home/travis/venv/lib/python2.7/site-packages/docutils/utils/__init__.py", line 21, in <module>
    import docutils.io
  File "/home/travis/venv/lib/python2.7/site-packages/docutils/io.py", line 348
    (self.destination.mode, mode)), file=self._stderr)
                                        ^
SyntaxError: invalid syntax
The command "python setup.py check --restructuredtext --strict" exited with 1.
```
Source: https://travis-ci.org/codespell-project/codespell/jobs/561689610#L467-L473

larsoner:
`wee` is a legit word, @peternewman did we decide to include these or no?

peternewman:
> `wee` is a legit word, @peternewman did we decide to include these or no?

I really don't think we should in the main dictionary. To the extent I've started a Travis check in https://github.com/codespell-project/codespell/pull/1142 to catch such things. I proposed there should be an additional "strict" dictionary which you could optionally load which has things like this.

peternewman:
Surely this should be in rare as part of #1142 ?

peternewman:
It is now.

PR #1218, opened 2019-08-09
Change from optparse to argparse

qyearsley:
Reason: optparse is deprecated; argparse is the mostly-similar replacement for optparse. There should be no significant changes in behavior, although the usage string printed when you use the --help flag is more detailed.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -19,12 +19,12 @@ Copyright (C) 2011  ProFUSION embedded systems
 
 from __future__ import print_function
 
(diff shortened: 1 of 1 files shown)


qyearsley:
Sorry for not updating this immediately after making the PR; there were two issues that needed to be fixed after the initial PR:

 - wrapped long (81 and 82 char) lines
 - argparse in python2 doesn't properly handle non-ascii help strings, changed one help string slightly

larsoner:
Thanks @qyearsley

PR #1220, opened 2019-08-10
Add aminat{or|ion|ions}->animat{or|ion|ions}

luzpaz:
`aminator` is a legit word but not frequently used in comparison to it being a typo. IMO this PR is merited.


larsoner:
I've never heard of it and it wasn't in the dictionary I checked

codespell_lib/data/dictionary.txt
```diff
@@ -1185,6 +1185,9 @@ amgle->angle
 amgles->angles
 amiguous->ambiguous
 amin->main, disabled because of var names
+amination->animation, amination,
+aminations->animations, aminations,
```
larsoner: I would remove the self-suggestion in this and the one above

luzpaz: https://en.wikipedia.org/wiki/Amination (sorry should of linked to this before)

larsoner:
Okay, just have to convince @peternewman then :)

peternewman:
> Okay, just have to convince @peternewman then :)

I still don't see why they can't all go in a secondary dictionary (aside from that capability not currently being present, but it can't be too hard).

I know it's called codespell, but it strikes me one of the fairly unique features of codespell is the curated nature of the dictionary, it doesn't find words not in it's dictionary, the dictionary is a list of genuine or likely typos and suggested corrections. If you're writing about something niche like enzyme catalysation, you may still want to use codespell to avoid getting flooded with false positive words that a normal dictionary doesn't know about.

Given that, to keep the core dictionary pure, we need to have some CI to keep on top of it, otherwise we get more and more of https://github.com/codespell-project/codespell/pull/1181/files and https://travis-ci.org/codespell-project/codespell/jobs/562631674 . I'm sure everyone could find at least one word in #1181 they would have or have already legitimately used in a comment.

codespell_lib/data/dictionary.txt
arm-in: also "lamination"

larsoner:
@luzpaz this (and possibly other) PR(s) can now be rebased and refactored given that we have multiple dictionaries

luzpaz:
> @luzpaz this (and possibly other) PR(s) can now be rebased and refactored given that we have multiple dictionaries

I'm a little bit turned around. How should I proceed?

larsoner:
Rebase or merge your PR, then make use of new dictionaries. Or close the PRs and start over if it's easier

peternewman:
If it's correcting to itself, i.e. it's a real word, put it in the rare dictionary ( https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary_rare.txt ), otherwise you can proceed as normal (unless it's a name etc).

luzpaz:
Revised commit, please see 1fe6501

codespell_lib/data/dictionary_rare.txt
```diff
@@ -1,3 +1,5 @@
+amination->animation, amination, lamination,
+aminations->animations, aminations, laminations,
```
peternewman, line 2: Remove the self-corrections

codespell_lib/data/dictionary.txt
```diff
@@ -1450,6 +1450,7 @@ amerliorate->ameliorate
 amgle->angle
 amgles->angles
 amiguous->ambiguous
+aminator->animator, laminator,
```
peternewman, line 1453: You say "aminator is a legit word but not frequently used in comparison to it being a typo." So this should be moved into rare too.

Ideally with a reference to where it's a real word.

sebweb3r: Aminator is used by netflix to create amazon machine images:

https://github.com/Netflix/aminator
https://netflixtechblog.com/ami-creation-with-aminator-98d627ca37b0

peternewman: Hmm, perhaps code dictionary then, good find @sebweb3r .

codespell_lib/data/dictionary_rare.txt
bl-ue:
```suggestion
amination->animation, lamination,
aminations->animations, laminations,
```

codespell_lib/data/dictionary.txt
bl-ue: ping @luzpaz.  If you don't want to do this I'd be happy to :)

bl-ue (approved):

Issue #410, opened 2018-03-22
"to and fro" is correct


EdwardBetts:
codespell suggests replacing "fro" with "for". Can we have an exception for the phrase "to and fro"?

https://en.wiktionary.org/wiki/to_and_fro

larsoner:
Sure, feel free to add it

EdwardBetts:
Thanks @larsoner. I couldn't find the an existing mechanism to specify multi word exceptions like this. Does it exist and I've missed it, or would it need to be added to handle this case?

larsoner:
Usually people just add the word to the list like:
```
fro->for, fro
```
Right @luzpaz ?

larsoner:
(I don't think there is a multi-word way in particular, I think it's okay to just use this single-word one in the meantime)

luzpaz:
Actually, you're missing a comma at the end:  
`fro->for, fro,`

peternewman:
So I've just sort of hit another of these, preform->perform. Well actually in this case it was a typo, but preform is a valid word too. But there are others like Cristal.

`cristal->crystal, cristal,`

Essentially we currently have some entries where the misspelling is also listed as a valid correction. There are two trains of thought here I guess, firstly it's nonsense and we should remove the entry, or more considered, that cristal is most likely a misspelling, but in some (rare) circumstances you may really mean it (see also fro, until we support multi-word corrections #255).

What's the general feeling? I've not used interactive mode, but there a "did you really mean?", and in manual/automatic the option to skip/ignore all potential ones seems like it might be sensible. Essentially treat likely typos differently from definite typos.

larsoner:
> I've not used interactive mode, but there a "did you really mean?",

That is more or less what these entries seem to do. I agree we could add a parameter to be more suggestive (default) or less suggestive (if the "error" is in the list of corrections, don't prompt or report)

peternewman:
I guess one of the things I've always liked about Codespell is the fact the dictionary is curated, rather than a list of valid words, and hence doesn't normally trip up on valid but obscure/technical words. It sort of feels to me that it goes against that ethos when words are added to the dictionary which are valid (although admittedly mostly rarely used). Even if the other variant (i.e. the "typo") is listed, but even more so when it's not.

larsoner:
We probably need a new argument for this, that (I agree) should disable these by default.

[...]

PR #1240, opened 2019-08-31
Disable 'implicity' due to false positives

luzpaz:
Implicity is a [legit word](https://www.merriam-webster.com/dictionary/implicity)

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -10292,7 +10292,7 @@ impletment->implement
 implicite->implicit, implicitly,
 implicitely->implicitly
 implicitley->implicitly
-implicity->implicitly
+implicity->implicitly, disabled due to common misspelling
 implict->implicit
 implictly->implicitly
 impliment->implement


peternewman:
Can we just remove it from the dictionary then? If we're reluctant to do that, please can you reword the disabled message, `disabled due to common misspelling` implies it is a typo, but it's not, `disabled due to being a valid, but infrequently used word` would be more accurate, but seems daft to me.

waldyrious:
Looking at the existing dictionary, it seems like the usual approach for such potential misspellings is to list both options:

    implicity->implicity, implicitly,

Could that be a solution here?

larsoner:
See also #712, #410, but I guess so

peternewman:
> Could that be a solution here?

I'm personally against that as a user as they still flag up and I have to whitelist them, or disable all of them.

waldyrious:
I see what you mean. My view is that, just like we don't have entries for all possible misspellings of all English words, so we should only add self-correction entries for valid words that are both rare and common misspellings of other words. Since the second criterion is (by definition) more likely to occur than the first one, it sounds like an acceptable trade-off to me.

peternewman:
Surely this should be in rare as part of #1142 ?

larsoner:
Yes I merged these then updated the PR

PR #1246, opened 2019-09-09
Add several spelling corrections + refinements

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -21923,6 +21958,7 @@ youseff->yourself, yousef,
 youself->yourself
 ypes->types
 yrea->year
+yse->yes, use,
```
arm-in: also "NYSE"

luzpaz: c88e38c

PR #1258, opened 2019-09-19
Disable some words that have led to false positives

qyearsley:
I've been using this dictionary for a spellchecker for the chromium codebase for a while, and there are a few entries that seem to often lead to false positives, and may lead to false positives for others.

I'm happy to omit any of these particular changes, these are just some suggestions.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1722,7 +1722,7 @@ arithemtic->arithmetic
 arithmatic->arithmetic
 arithmentic->arithmetic
 arithmethic->arithmetic
-arithmetics->arithmetic
+arithmetics->arithmetic, arithmetics,
 arithmitic->arithmetic
 aritmetic->arithmetic
 aritst->artist
@@ -2694,9 +2694,9 @@ blindy->blindly
 Blitzkreig->Blitzkrieg
 bload->bloat
 bloaded->bloated
-bloc->block
+bloc->block, bloc,
 bloccks->blocks
-blocs->blocks
+blocs->blocks, blocs,
 blohted->bloated
 blokcer->blocker
 bloking->blocking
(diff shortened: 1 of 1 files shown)


luzpaz:
I'd also add `uint`, `sinc`, `dof` as well

qyearsley:
Also disabled sinc and dof based on comment from luzpaz@, uint already had a "disabled" reason

codespell_lib/data/dictionary.txt
```diff
@@ -11674,7 +11674,7 @@ iterm->term, intern,
 iterpreter->interpreter
 iterrupt->interrupt
 iteself->itself
-ith->with
+ith->with, disabled because of ordinal form of i, like nth
```
luzpaz: travis doesn't like that a 2nd comma was used:  
```
line       = b'ith->with, disabled because of ordinal form of i, like nth\n'
msg        = 'error %s: correction %r contains a comma *not* followed by a space'
```

Suggested edit would be `ith->with, disabled because of ordinal form of 'i' (like 'nth')`

qyearsley: It's good that it caught that! Now changed

luzpaz:
also `arithmetics->arithmetic` or should we just remove this one altogether ?

qyearsley:
Honestly, I think that if we `arithmetics->arithmetic` is removed, then there are some others that could be disabled or removed, where the word may be rare but is a legitimate accepted (possibly informal or uncommon) spelling, e.g. cant, coo, crated, followings, fro, gird, leaded, lightening, planed, process', rime, revered, therefrom, theses, tho, thru, vertexes, whet, wont.

Following the current pattern, I'd say we could perhaps just add the word itself as one of the suggested words? e.g. `wit->wit, with,`, if you wrote wit you might mean with, but you might have meant wit. Similarly if you wrote arithmetics, you probably meant arithmetic, but also you could have itentionally meant arithmetics with an "s"...

Now updated the commit to add the original possible spelling for arithmetics, coo, bloc, rime, tho, thru, wounder; what do you think?

luzpaz:
@qyearsley forget `arithmetics` for now. I say just merge this as it is and we can continue the conversation in the issue tracker.

PR #1277, opened 2019-10-08
Add spelling correction for fixe->fixed, fixes, fix

cfi-gb:
Have seen this "live" from time to time but i'm not sure if this is too generic as it might be e.g. a variable or an abbreviation:

https://github.com/search?q=%22fixe%22

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -8928,6 +8928,7 @@ fitering->filtering
 fitler->filter
 fitlers->filters
 fivety->fifty
+fixe->fixed, fixes, fix,
 fixeme->fixme
 fizeek->physique
 flage->flags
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -8928,6 +8928,7 @@ fitering->filtering
 fitler->filter
 fitlers->filters
 fivety->fifty
+fixe->fixed, fixes, fix,
```
peternewman: Suggest fixme.

arm-in: also "fixer"

cfi-gb: Good points, added both in e5f687a3da1f89e3069adba7dcf0f5d765a930bb

PR #1281, opened 2019-10-09
Fix "cahgne->cahgne"

Gelma:
Introduced in commit:
https://github.com/codespell-project/codespell/commit/ea85402b7b017b5a418a9bb2171bbeaf54de6692

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3017,7 +3017,7 @@ cahdidate->candidate
 cahdidates->candidates
 cahe->cache
 cahes->caches
-cahgne->cahgne
+cahgne->change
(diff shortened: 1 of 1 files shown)


peternewman:
Before we merge this we should do some TDD to catch when a replacement is identical to it's change (and there aren't others).

peternewman:
Hmm, I did this in https://github.com/codespell-project/codespell/pull/1138 so why isn't it working?

peternewman:
Understood and fixed in https://github.com/codespell-project/codespell/pull/1284 . So I've pulled in this change in https://github.com/codespell-project/codespell/commit/a911c7728047c496c223a2a5f1c65f7526f392dc , hence closing this.

PR #1290, opened 2019-10-13
Add variation on the typo 'chrash'

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -3585,7 +3585,11 @@ chracter->character
 chracters->characters
 chractor->character
 chractors->characters
-chrashed->crashed
+chrash->crash, trash,
```
peternewman: Suggest `crash, thrash, trash,` and the same for all the others.

PR #1314, opened 2019-10-25
Refine preexisting results

luzpaz:


waldyrious:
> It's failing as it says unuseful is an error.

It [seems to be valid](https://en.wiktionary.org/wiki/unuseful) though.

peternewman:
> > It's failing as it says unuseful is an error.
> 
> It [seems to be valid](https://en.wiktionary.org/wiki/unuseful) though.

Sorry what I didn't explain very well is the system thinks its an error because there is a later correction of it to something else.
https://github.com/codespell-project/codespell/blob/c065c6c33d6389cb022bd965381c614e8abc5764/codespell_lib/data/dictionary.txt#L21827

sebweb3r:
push

Maybe one can move 
```
 unuseful->useless 
```
to rare?

peternewman:
> push
> 
> Maybe one can move
> 
> ```
>  unuseful->useless 
> ```
> 
> to rare?

I think it will currently still complain across dictionaries (as it normally should).

I suspect these should perhaps all go in informal as they don't seem to be typos per-say, just usage.

luzpaz:
Is there consensus ?

sebweb3r:
unuseful is not in aspell, thus it would fail the tests?!

peternewman:
So does someone have a more authoritative source than Wiktionary for it being a real word?

waldyrious:
> So does someone have a more authoritative source than Wiktionary for it being a real word?

Maybe [Google Ngram Viewer](https://books.google.com/ngrams/graph?content=unuseful)?

bl-ue:
So I guess the general consensu is that this PR should be closed.  If we decide otherwise, feel free to reopen it.

PR #1322, opened 2019-11-06
Remove comma from simple entries w/o given reason

c72578:
- Remove the comma at the end of simple entries without given reason.
  Use the following command to find these entries:
```console
  awk -F , 'NF == 2' < ./codespell_lib/data/dictionary.txt | \
  grep --invert disabled
```
- Output was:
```console
  inteligable->intelligible,
  inturpratasion->interpretation,
  invokable->invocable,
  lightsensor->light sensor,
  won;t->won't,
```

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -11742,7 +11742,7 @@ intelegence->intelligence
 intelegent->intelligent
 intelegently->intelligently
 inteligability->intelligibility
-inteligable->intelligible,
+inteligable->intelligible
 inteligance->intelligence
 inteligantly->intelligently
 inteligence->intelligence
@@ -12057,7 +12057,7 @@ intruments->instruments
 intrusted->entrusted
 intstead->instead
 intuative->intuitive
-inturpratasion->interpretation,
+inturpratasion->interpretation
 inturpratation->interpretation
 inturprett->interpret
 intutive->intuitive
@@ -12099,7 +12099,7 @@ invlove->involve
 invloved->involved
 invloves->involves
 invocaition->invocation
-invokable->invocable,
+invokable->invocable
 invokation->invocation
 invokations->invocations
 invokee->invoked, invoke,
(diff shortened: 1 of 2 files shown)


larsoner:
@c72578 are you up for modifying our tests to catch future errors like this? Basically here:

https://github.com/codespell-project/codespell/blob/master/codespell_lib/tests/test_dictionary.py#L38-L43

You could probably do something like:
```
if rep.count(',') == 1 and rep.split(',')[-1].strip() == '':
    raise AssertionError(...)
```

c72578:
Good point.
I have updated **`test_dictionary.py`** now according to your recommendation.

larsoner:
@c72578 I realized that the check could be matched with the existing code better by writing it as a regex, so I did that and pushed. Will merge once CIs come back happy

larsoner:
Thanks @c72578

c72578:
I just did some testing with single whitspaces in between and also get the assert for the trailing comma, even if there is no trailing comma, e.g.:
```console
line       = 'instansiate-> instantiate\n'
msg        = '%s has a single entry but contains a trailing comma'

line       = 'instansiate ->instantiate\n'
msg        = '%s has a single entry but contains a trailing comma'
```
However, these errors are detected, but the message is not the correct one.

larsoner:
@c72578 I do not get this error on `master` / with my changes. If I change the first line in `codespell_lib/data/dictionary.txt` to be:
```
1nd ->1st
```
Then `pytest codespell_lib/tests/test_dictionary.py` gives:
```
AssertionError: error '1nd ' has whitespace
```
If I move the space to the correction I get:
```
AssertionError: error 1nd: correction ' 1st' cannot start with whitespace
```
So I can't replicate your problem. What modifications are you making and code are you running to see errant behavior?

c72578:
Please try one of the subsequent ones with current master, not the first one.
e.g.
`aaccessibility-> accessibility`
or
`instansiate-> instantiate`

larsoner:
```
aaccessibility-> accessibility
```
yields:
```
E               AssertionError: error aaccessibility: correction ' accessibility' cannot start with whitespace
```

c72578:
And later on in the output:
`...`
`line       = 'aaccessibility-> accessibility\n'`
`msg        = '%s has a single entry but contains a trailing comma'`
`...`

c72578:
What do you think about clearing `msg` at the beginning of the loop?
```diff
         for line in fid:
+            msg=''
             err, rep = line.decode('utf-8').split('->')
```

PR #1325, opened 2019-11-07
Delete msg after loop

c72578:
The last value of `msg` will not be carried to the next line and a
misleading value of `msg` for a different assert can be avoided.

### DIFF
--- a/codespell_lib/tests/test_dictionary.py
+++ b/codespell_lib/tests/test_dictionary.py
@@ -41,6 +41,7 @@ def test_dictionary_formatting():
                  '%s has a single entry but contains a trailing comma'),
             ]:
                 assert not re.search(r, rep), (msg % (prefix,))
+            del msg
             rep_count = rep.count(',')
             if rep_count and not rep.endswith(','):
                 assert 'disabled' in rep.split(',')[-1], \


codespell_lib/tests/test_dictionary.py
```diff
@@ -12,6 +12,7 @@ def test_dictionary_formatting():
     with open(op.join(op.dirname(__file__), '..', 'data',
                       'dictionary.txt'), 'rb') as fid:
         for line in fid:
+            msg = ''
```
larsoner: I don't see how this could affect the output of running `pytest`

c72578:
Please have a look at the attached outputs of running pytest, especially:
`msg        = '%s has a single entry but contains a trailing comma' `
->
`msg        = '' `

[accessibility_1_before.txt](https://github.com/codespell-project/codespell/files/3821391/accessibility_1_before.txt)
[accessibility_2_fixed_PR_1325.txt](https://github.com/codespell-project/codespell/files/3821390/accessibility_2_fixed_PR_1325.txt)

larsoner:
Oh in the locals. Sure that can be junk. I always look at the `AssertionError` line, it's more informative.

A cleaner fix would be to `del msg` after the loop. Then it won't even show up in the locals. It's clearer why it's being done, as looking at the code there is no clear reason to have `msg = ''` at the top of the  loop.

c72578:
OK, thanks for the feedback.
I have amended the commit.

PR #1332, opened 2019-11-13
Add variation typos of 'enhance'

luzpaz:


albert-github:
I agree that these typos (`enhabce`) are typos and it is good to catch them, but this brings also the `enhamce` versions on the table.
Actually maybe a more flexible checking with like `enha.ce` or `enha[a-mo-z]ce` should be implemented.

larsoner:
> Actually maybe a more flexible checking with like enha.ce or enha[a-mo-z]ce should be implemented.

This would require changing the guts of how things work. Also it might be (very) bad for performance as it would require regex parsing rather than simple `dict` lookup.

PR #1344, opened 2019-11-24
Add meida->media

peternewman:


codespell_lib/data/dictionary.txt
```diff
@@ -14823,6 +14828,7 @@ optmizations->optimizations
 optmize->optimize
 optmized->optimized
 optomism->optimism
+Orcale->Oracle
```
luzpaz: vote to lowercase these, oracle doesn't only refer to the company

peternewman: I did consider it, but given our environment it's probably most likely.

I don't feel too strongly if the consensus is it should be lower case.

PR #1350, opened 2019-11-27
Gauss spellings and more

banesullivan:
Gauss should be capitalized as it's referring to a human (proper noun).

Note: what I did doesn't seem to work. Is the dictionary case insensitive?

Output from a sample project:

```
./section5.tex:102: gauss'  ==> gauss'
./section1.tex:669: Gauss'  ==> Gauss'
./section3.tex:377: Gauss'  ==> Gauss'
./section3.tex:385: Gauss'  ==> Gauss'
```

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -22070,7 +22070,7 @@ therby->thereby
 thereads->threads
 therefor->therefore, therefor,
 therefrom->there from
-therem->there
+therem->there, theorem,
 thereom->theorem
 thererin->therein
 theres->there's
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -10953,6 +10953,7 @@ guarunteing->guaranteeing
 guaruntes->guarantees
 guarunty->guaranty
 guas->Gauss
+guass->Gauss
 guassian->gaussian
 Guatamala->Guatemala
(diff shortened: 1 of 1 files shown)


larsoner:
> Is the dictionary case insensitive?

Yes at this point it just maintains the case of whatever the original spelling was IIRC. At some point we might add better case handling, though.

You have an error:

> AssertionError: currently corrections must end with trailing "," (if  multiple corrections are available) or have "disabled" in the comment

banesullivan:
Okay I made this so that it only corrects misspellings Gauss and suggests a correction with a capital G. Thus "gauss" would be ignored by codespell because it isn't case-sensitive

banesullivan:
and `becausw->because` which is a really common one for me

banesullivan:
I've been writing a lot of lecture slides lately and I added all the mispellings I made that `codespell` didn't recognize.

I also added `,easure->measure`. Will that mess up the script because it parses the dictionary using commas?

banesullivan:
Welp, travis answered that question

larsoner:
Thanks @banesullivan

lurch:
Not that it really matters, but FYI https://en.wikipedia.org/wiki/Gauss_(unit)#Name,_symbol,_and_metric_prefixes says

> When the unit is spelled out, it is written in lowercase ("gauss"), unless it begins a sentence.

PR #1351, opened 2019-11-27
Allow to use --skip multiple times

dseight:
Make it possible to add custom system-wide ignored directories with an alias in shell rc script, e.g.:
```bash
alias codespell="codespell --skip=".git,node_modules"
```
and allow to extend this list by executing codespell alias with extra `--skip` argument.


larsoner:
Thanks @dseight

PR #1362, opened 2019-12-05

c72578:
The corrections are based on the following real world examples:
autoconpleting, momemtarily

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -2565,6 +2565,11 @@ auto-realease->auto-release
 auto-realeased->auto-released
 autochtonous->autochthonous
 autocommiting->autocommitting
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -7104,6 +7114,7 @@ depricates->deprecates
 depricating->deprecating
 dequed->dequeued
 derageable->dirigible
+derails->derails, details,
```
larsoner: For now can you remove this? See for example #1142

c72578: OK, `derails->derails, details,` has been removed

luzpaz: IMO, we should have a separate dictionary or perhaps flag that lets the user search these type of typos.

PR #1387, opened 2019-12-18
Fix underware with underwear

Gelma:


luzpaz:
Nice catch. In that case can you also add `underware->underwear` please?

Gelma:
Done. I'm going to push.

peternewman:
> Done. I'm going to push.

We can't see it yet @Gelma ?

Gelma:
> > Done. I'm going to push.
> 
> We can't see it yet @Gelma ?

The Luzpaz request is already pushed. I guess I pushed before write "I am going to push", so we have commit before commet.

Please check the commits. For me is complete.

peternewman:
> The Luzpaz request is already pushed.

Apologies I misread it as a plural.

PR #1388, opened 2019-12-18
Fix entry intiialize->initiialize

Gelma:


Gelma:
Done. I'm going to push.

PR #1410, opened 2020-01-15
GitHub Action to automate the identification of common misspellings

cclauss:
GitHub Action to automate the identification of common misspellings in text files.
https://github.com/codespell-project/codespell

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -10,4 +10,4 @@ jobs:
       - uses: actions/setup-python@v1
       - run: pip install codespell
       # When using this Action in other repos, the --skip option below can be removed
-      - run: codespell --skip=./.*,./codespell_lib,./example/code.c
+      - run: codespell --skip=./.*,./codespell_lib/data,./example/code.c,./test_basic.py,*.pyc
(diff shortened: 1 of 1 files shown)


cclauss:
This can be used by other projects to rapidly add __codespell__ to their repos.

larsoner:
Rather than actually add this as an action for this repo, we could add this to the `README`. What you have here will fail because we have:
```
larsoner@bunk:~/python/codespell$ codespell | wc -l
...
186
```
So if I merge this PR it will fail the check. If you want to make it pass, then you'll need to tweak the call. For us maybe `codespell codespell_lib/_codespell.py` is good enough. That passes for me locally.

.github/workflows/codespell.yml
```diff
@@ -9,4 +9,5 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v1
       - run: pip install codespell
-      - run: codespell
+      # When using this Action in other repos, the --skip option below can be removed
+      - run: codespell --skip=./.*,./codespell_lib,./example/code.c
```
larsoner: This now skis `codespell_lib` which is where the library code actually lives, which is not so helpful. Maybe:
```
codespell --skip=./.*,./example/code.c,*.pyc,./codespell_lib/data,test_basic.py
```

larsoner:
Thanks @cclauss

PR #1415, opened 2020-01-23
Remove two non-typos from the list

waldyrious:
"memorise" and "minimise" are note misspellings; they're just British spelling variants of the words.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -14654,7 +14654,6 @@ memmory->memory
 memoery->memory
 memomry->memory
 memor->memory
-memorise->memorise, memorize,
 mempry->memory
 memroy->memory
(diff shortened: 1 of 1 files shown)


larsoner:
See also #103, but it seems like we have been removing these things so let's merge

peternewman:
> See also #103, but it seems like we have been removing these things so let's merge

My feeling (as a Brit), is that they shouldn't be present in the core list (given they're valid spellings), but hopefully someone (possibly not me the way things are going), will deal with the multiple dictionary stuff soon.

waldyrious:
As an aside — [Muphry's law](https://en.wikipedia.org/wiki/Muphry's_law
) strikes again! There's a typo in my commit message 🤦‍

> ...are **note** misspellings

Issue #233, opened 2017-11-27
escaped characters should be recognized 


luzpaz:
The term `wasn\'t` is being recognized as `wasn` by codespell. There should be logic to distinguish this better.

peternewman:
https://github.com/lucasdemarchi/codespell/pull/201

This currently causes false positives due to not matching wasn\'t, but I guess that would be the case even without the PR above, it would just be suggesting was instead of wasn't.

arm-in:
+1

"\nTo terminate ..."
leads to
"nto->not, disabled due to \n"

yarikoptic:
IMHO the original issue of escaped `\'` (where result would constitute just removal of `\`) is different from `\n` problem (where result should constitute of completely removing `\n` and indicating word boundary there), the oldest instance of which is mentioned in the oldest still open issue #17 where I just provided an extra example.

PR #1422, opened 2020-02-06
🚧 Ignore escaped characters

akaszynski:
This take a stab at ignoring escaped characters.  It's a basic (and probably inefficient) implementation, so if there's a better way to get this done, please let me know.

Also, the escape dictionary should be modifiable by the user.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -33,8 +33,6 @@ USAGE = """
 """
 VERSION = '1.17.0.dev0'
 
-SUB_DICT = {'\\n': ' ', r"\'": "'"}
-
 # Users might want to link this file into /usr/local/bin, so we resolve the
 # symbolic link path to the real path if necessary.
 default_dictionary = os.path.join(os.path.dirname(os.path.realpath(__file__)),
@@ -234,6 +232,10 @@ def parse_options(args):
                         help='Comma separated list of words to be ignored '
                              'by codespell. Words are case sensitive based on '
                              'how they are written in the dictionary file')
+    parser.add_argument('-P', '--sub-pairs', type=str, metavar='FILE',
+                        help='Custom substitution text file that contains '
+                             'substituions key value pairs.  Can be used to '
+                             'substitute escaped characters.')
     parser.add_argument('-r', '--regex',
                         action='store', type=str,
                         help='Regular expression which is used to find words. '
@@ -346,6 +348,25 @@ def build_dict(filename, misspellings, ignore_words):
             misspellings[key] = Misspelling(data, fix, reason)
 
 
+def build_sub_pairs(filename):
+    """Parse substitution pairs from a text file.
+
+    Notes
+    -----
+    File expected to be in the following format
+    tobesubstituted->substituted
+    """
+    if not os.path.isfile(filename):
+        raise Exception('Unable to find sub pair file "{}"'.format(filename))
+
+    sub_pairs = {}
+    with codecs.open(filename, mode='r', encoding='utf-8') as f:
+        for line in f:
+            [key, data] = line.split('->')
+            sub_pairs[key] = data
+    return sub_pairs
+
+
 def is_hidden(filename, check_hidden):
     bfilename = os.path.basename(filename)
 
@@ -372,7 +393,7 @@ def fix_case(word, fixword):
 
 
 def multiple_replace(find_dict, text):
-    """Multiple find and replace based on a dictionary
+    """Multiple find and replace based on a dictionary.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


akaszynski:
Same problem here.  I've noticed that ``regex`` is almost always slower than the built-in ``replace``.

akaszynski:
We're now using replace.  As for making the substitution dictionary customizable, would you prefer to have a file containing the key/value pairs, or a string that they read in?

larsoner:
> As for making the substitution dictionary customizable, would you prefer to have a file containing the key/value pairs, or a string that they read in?

I find putting characters like `\n` in the command line to be annoying (usually takes me multiple tries to get it right), so I would lean toward a file

akaszynski:
Agreed, I was encountering that issue as well when testing it out.  I'm thinking simple key value pairs in a csv, where the pairs are separated by line breaks.

```
"\n", " "
"\'", "'"
```

larsoner:
Might make more sense to match the dictionary format

thing->replacement

akaszynski:
As we’re going to have to use spaces, are quotes permitted in your dictionary format?
```
"\n"->” "
"\'"->"'"
```

larsoner:
I would just make it that anything to the left of `->` is the substring to look for, and anything to the right is the replacement. So:
```
\n-> 
\'->
```

akaszynski:
Should be good to review now.

codespell_lib/_codespell.py
```diff
@@ -232,6 +232,10 @@ def parse_options(args):
                         help='Comma separated list of words to be ignored '
                              'by codespell. Words are case sensitive based on '
                              'how they are written in the dictionary file')
+    parser.add_argument('-P', '--sub-pairs', type=str, metavar='FILE',
+                        help='Custom substitution text file that contains '
+                             'substituions key value pairs.  Can be used to '
```
c72578: substituions->substitutions

codespell_lib/tests/test_basic.py
```diff
@@ -73,6 +75,28 @@ def test_basic(tmpdir, capsys):
...
+def test_escaped(tmpdir, capsys):
+    """Test escaping characters"""
+    d = str(tmpdir)
+    with open(op.join(d, 'escaped_char.txt'), 'w') as f:
+        f.write(r"\n\nWe can")
+    assert cs.main(d) == 0
+
+
+def test_escaped_sub_file(tmpdir, capsys):
+    """Test escaping characters using substituion file"""
```
c72578: substituion->substitution

codespell_lib/_codespell.py
akaszynski: Funny, because this made it through the codespell check.

larsoner:
Looks good. One last thing: we probably want to provide and use a default file, just like we do for the dictionary. Maybe it should just contain `\n-> ` to start.

akaszynski:
> Looks good. One last thing: we probably want to provide and use a default file, just like we do for the dictionary. Maybe it should just contain `\n-> ` to start.

Perhaps `\'->'` as well.

larsoner:
Fine with me

codespell_lib/_codespell.py
```diff
@@ -232,6 +232,10 @@ def parse_options(args):
                         help='Comma separated list of words to be ignored '
                              'by codespell. Words are case sensitive based on '
                              'how they are written in the dictionary file')
+    parser.add_argument('-P', '--sub-pairs', type=str, metavar='FILE',
+                        help='Custom substitution text file that contains '
```
peternewman: I'm a bit unclear from the help detail what this is for, is it to "fix up" the dictionary to deal with it matching escape sequences? To actually do sed type runs on my codebase or something else?

peternewman: Is this linked to https://github.com/codespell-project/codespell/issues/233 ?

larsoner:
Also in the related PR #174 it was noted that the `write-changes` flag was problematic when dealing with escapes, if we come back to this we should make sure that `write-changes` works even with these substitutions

bl-ue:
Why the close?  I se this is quite old 🤔

akaszynski:
> Why the close? I see this is quite old thinking

Just cleaning up old PRs.  I'd like to work on this, but this project hasn't been updated in quite some time (last release 6 months ago).

peternewman:
Hi @akaszynski ,

Despite the lack of releases, coding is still happening (although mostly in the dictionary), also see #1923:
https://github.com/codespell-project/codespell/compare/v2.0.0...master

You didn't seem to have respond to the review comments myself and @larsoner had left if you were expecting it to have been merged.

akaszynski:
> You didn't seem to have respond to the review comments myself and @larsoner had left if you were expecting it to have been merged.

True, there's still work to be done.  I'll work on this.

PR #1425, opened 2020-02-25
Update dictionary with more misspellings and disabled names

qyearsley:
This PR:
- Disables a few more person names: Amin, Ang, Anny, Bae, Chang, Que, Sargent, Wen
- Also disables Synopsys (company), thru (informal).
- Adds a few misspellings I've seen: callabck, cleand, notificaiton, seprated, serailze.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1429,7 +1429,7 @@ amerliorate->ameliorate
 amgle->angle
 amgles->angles
 amiguous->ambiguous
-amin->main, disabled because of var names
+amin->main, disabled because amin might be a var name
 amke->make
 amking->making
 ammend->amend
@@ -3658,6 +3658,7 @@ calilng->calling
 caliming->claiming
 caling->calling, scaling, culling,
 callabck->callback
+callabcks->callbacks
 callack->callback
 callbacl->callback
 callbacsk->callback
@@ -15957,6 +15958,8 @@ notifed->notified
 notifer->notifier
 notifes->notifies
 notificaction->notification
+notificaiton->notification
+notificaitons->notifications
 notificaton->notification
 notificiation->notification
 notifiy->notify
@@ -21073,6 +21076,7 @@ seporate->separate
 sepperate->separate
 seprarate->separate
 seprate->separate
+seprated->separated
 seprator->separator
 seprators->separators
 Septemer->September
@@ -21118,6 +21122,8 @@ serached->searched
 serailisation->serialisation
 serailization->serialization
 serailized->serialized
+serailze->serialize
+serailzed->serialized
 sergent->sergeant
 serialiazation->serialization
 serie->series


codespell_lib/data/dictionary.txt
```diff
@@ -2914,7 +2914,7 @@ backwords->backwards, backwords,
 backwrad->backward
 bactracking->backtracking
 bacup->backup
-bae->base
+bae->base, disabled due to being a person's surname
```
peternewman: Personally surnames seem like a slippery slope to me. I can just about see first names, but the range of surnames is far greater. TBH both first and surnames feel like local whitelists per project really. Or again a separate dictionary for them.

codespell_lib/data/dictionary.txt
```diff
@@ -22779,7 +22786,7 @@ synomym->synonym
 synonim->synonym
 synonomous->synonymous
 synonymns->synonyms
-synopsys->synopsis
+synopsys->synopsis, disabled since this is also a company name
```
peternewman: Similar to surnames, but probably worse, companies have a terrible habit of misspelling stuff, as someone has already taken the correct spelling, or because they want to look trendy.

codespell_lib/data/dictionary.txt
qyearsley: That's a good point.

My argument for including these names is:
 - The number of common misspellings that are also common names is actually not too large, and is much smaller than the list of all names.
 - It can be pretty annoying if your name gets *automatically* "corrected" after being flagged as a misspelling. Furthermore, if it was a traditional English name like Johnson or Smith it would never be auto-corrected, but if you have the surname Chang, Wen, Bae etc. (each not too rare in different parts of Asia) then you'd get flagged, which seems slightly unfair.

codespell_lib/data/dictionary.txt
qyearsley: That's true, although again I'd argue that the number of large tech companies that may be mentioned in source code with names that are also misspellings is not too large.

qyearsley:
Thanks for review -- I still think it might be a good idea to consider including the names (and the dictionary already contains `"tim->time, Tim, disabled due to being a person's name"`.

But if you feel that it's better to remove all mention of names and include misspellings that are also people's names, I could remove the names from this PR.

codespell_lib/data/dictionary.txt
peternewman: I don't think I'd ever personally heard of this company.

I feel companies are a weaker reasoning as they don't pop up unexpectedly/the chances of "them" contributing is probably significantly lower, and therefore white-listing seems more appropriate.

codespell_lib/data/dictionary.txt
peternewman: Fair point, I'd suggest in this particular case, bae could perhaps be disabled due to being a slang term of endearment, and the name of a large company (BAE Systems; although as below I'd prefer to avoid that), primarily before you hit the surname aspect.

If you take "variants" of my name, then there are Peter, Pete and Petr that I'm immediately aware of; Petr could be a typo of Petri (dish).

Off the top of my head, and perhaps not a great example, Smyth ( https://en.wikipedia.org/wiki/Smyth ) is a surname, but also possibly a typo of Smith (the keys are even in similar places to be a plausible typo).

I take your point though, but to me this feels like another strong use-case for the multi-dictionary approach I mentioned in https://github.com/codespell-project/codespell/issues/103#issuecomment-354913920 .

codespell_lib/data/dictionary.txt
```diff
@@ -23291,7 +23298,7 @@ throuh->through
 throuth->through
 throwed->threw, thrown,
 throwgh->through
-thru->through, thru,
+thru->through, thru, disabled because one may want to allow informal spelling
```
peternewman: I think we should drop the replacement of itself, in fact I thought I added some code so Travis would catch such anomalies, but perhaps it ignores the disabled ones.

peternewman:
> But if you feel that it's better to remove all mention of names and include misspellings that are also people's names, I could remove the names from this PR.

I'm merely another user, but the pragmatic solution may be to split the uncontentious non-name changes into another PR so they can be merged, while a consensus is potentially reached over what's in a name.

codespell_lib/data/dictionary.txt
qyearsley: (Dropping names including this company name from this CL)

codespell_lib/data/dictionary.txt
qyearsley: Done. There are definitely currently a bunch of current cases where the original spelling is marked as one of the possible good spellings of a potential misspellings.

qyearsley:
Updated The idea in #103 is interesting; it seems like in general we could trend towards one more complicated dictionary, or multiple dictionaries/ignore-lists with more functionality

As-is, for my particular use case, I'm going to have to keep my own [ignore list](https://cs.chromium.org/chromium/infra/go/src/infra/tricium/functions/spellchecker/spellchecker.go?l=54), although ignoring common proper nouns seems like a reasonable think to potentially include in codespell itself.

codespell_lib/data/dictionary.txt
peternewman: Thanks

codespell_lib/data/dictionary.txt
peternewman: Yeah agreed. What's your take on this @larsoner ? I've just added #1437 to continue to try and clean up the dictionary.

The first one it's found is:
`arithmetics->arithmetic, arithmetics,`

Should this be banned throughout, or allowed where it's disabled (for suggestions) or what?

This particular example I'd agree is probably fairly rare in plural form, but there's no note to advise the person doing a manual correction of that, and why its flagged in a non-interactive session. Another multi-dictionary candidate?
https://www.merriam-webster.com/thesaurus/arithmetics

peternewman: @qyearsley . Sorry to keep dragging you into this, if you revert this change, I'm personally happy to merge as is. Otherwise I'd rather get some general consistency. Also please do feel free to re-open a new PR with the names stuff/or raise an issue, so some consensus can be reached.

qyearsley: No problem. I've now reverted the change on this line.

peternewman:
Thanks @qyearsley . Please open a new one with the other more "controversial" changes. And we'll see where #1437 goes too.

peternewman:
@qyearsley we now have a solution. If you can add the names in a new PR to dictionary_names.txt would be great!

qyearsley:
Great news, good to know, thanks :-)

codespell_lib/data/dictionary.txt
peternewman: Done in #1476.

codespell_lib/data/dictionary.txt
peternewman: This is now in https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary_informal.txt

PR #1430, opened 2020-03-03

pabs3:


codespell_lib/data/dictionary.txt
```diff
@@ -6187,6 +6187,8 @@ coomand->command
 coommand->command
 coonstantly->constantly
 coonstructed->constructed
+coopearte->cooperate
+coopeartes->cooperates
```
peternewman: Also suggest the hyphenated version, for non Americans.

peternewman: Personally I was meaning suggest either alternative, i.e.
`coopearte->cooperate, co-operate,`

But I guess others may feel that's too much language translation.

pabs3:
For American vs other English, I tend to prefer to keep an American
typo alongside an American fix and the similar for other English.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #1444, opened 2020-03-14
Add finanace->finance and friends

peternewman:


codespell_lib/data/dictionary.txt
```diff
@@ -9805,6 +9805,13 @@ etablishd->established
 etablished->established
 etablishing->establishing
 etcp->etc
+etensible->extensible
+etend->extend
+etended->extended
+etender->extender
+etends->extends
```
luzpaz: Phonetically could be `attend` and it's plural variants..right?

codespell_lib/data/dictionary.txt
```diff
@@ -9806,10 +9806,10 @@ etablished->established
 etablishing->establishing
 etcp->etc
 etensible->extensible
-etend->extend
-etended->extended
-etender->extender
-etends->extends
+etend->extend, attend,
+etended->extended, attended,
+etender->extender, attender,
```
luzpaz: Oh yea, forgot to mention this: https://ell.stackexchange.com/questions/5482/attender-vs-attendee

peternewman: It seems to be a valid word, Collins has it:
https://www.collinsdictionary.com/dictionary/english/attender

Or are you suggesting we should offer the suggestion of a more popular spelling of a phonetic variation? That seems a bit of a stretch to me.

luzpaz:
Need a `make sort-dictionary`

codespell_lib/data/dictionary.txt
luzpaz: Yes, i do think we could add it. There is no harm in it as it is a 3rd option and phonetically could be a valid mistake.

peternewman: Sorry which one @luzpaz , `attendee`, I fail to see how `dee` sounds like `der`, or am I misunderstanding you?

luzpaz: I think both should be added, that way the reviewer can decide what was intended by the author (depending on the context)

peternewman: That feels more thesaurus like to me, I'm not aware of other bits of the dictionary suggesting entirely different words you might have meant.

luzpaz: It's your call. In my mind, it makes sense to add it. People who learn English as a second language and contribute to programming projects that are based in english, may not know some of these nuances. Codespell could illuminate these nuances. Like I said, it's your call.

peternewman: Sorry, I get your point, but personally I feel that needs to be part of the PR review process when fixing a typo (or just submitting code to the project you're working on). I don't believe it will "illuminate the nuances" unless like a dictionary it also has definitions so you can realise you've used the wrong one. I'd suggest the more sane thing when it flags a spelling mistake would be to check the corrections offered and see if you've even got the right word.

PR #1450, opened 2020-03-23
Fix several spelling corrections

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -7874,7 +7883,7 @@ determind->determined
 determinded->determined
 determinee->determine
 determineing->determining
-determing->determining
+determing->determining, determine,
```
peternewman: Nice find!

codespell_lib/data/dictionary.txt
```diff
@@ -12575,6 +12585,9 @@ imporvement->improvement
 imporving->improving
 imporvment->improvement
 imposible->impossible
+impot->import
+impotr->import, importer,
+impotrt->imported, importer, import,
```
peternewman: I think import should probably be first here, as it's just one extra letter.

codespell_lib/data/dictionary.txt
```diff
@@ -17547,6 +17564,7 @@ poeple->people
 poer->power
 poety->poetry
 pogress->progress
+poindcloud->pointcloud
```
peternewman: Seems a bit specific?

codespell_lib/data/dictionary.txt
```diff
@@ -19236,6 +19254,8 @@ recivership->receivership
 recives->receives
 reciving->receiving
 reclaimation->reclamation
+recnt->recent
```
peternewman: Suggest recant

codespell_lib/data/dictionary.txt
```diff
@@ -20195,6 +20215,10 @@ reprocuce->reproduce, reprocure,
 reprocuced->reproduced, reprocured,
 reprocuces->reproduces, reprocures,
 reprocucing->reproducing, reprocuring,
+reprodice->reproduce
+reprodiced->reproduced
+reprodicing->reproducing
+reprodiction->reproduction
```
peternewman: Can you also add the dic->duc variants of 
```
reproducibly
reproducibility
reproducible
reproducibility
reproducibly
reproducibility
```

codespell_lib/data/dictionary.txt
```diff
@@ -20977,6 +21005,9 @@ scipting->scripting
 scipts->scripts, skips,
 scirpt->script
 scirpts->scripts
+scketch->sketch
+scketched->sketched
+scketching->sketching
```
peternewman: Add scketches->sketches

codespell_lib/data/dictionary.txt
```diff
@@ -23596,6 +23635,9 @@ thurrow->thorough
 thurver->further
 thw->the, thaw,
 thyat->that
+tich->thick, tick,
```
peternewman: Suggest titch too

codespell_lib/data/dictionary.txt
```diff
@@ -25562,7 +25604,7 @@ wiew->view
 wigdet->widget
 wigdets->widgets
 wighed->weighed, wicked,
-wight->weight, white, right, wight,
+wight->weight, white, right, write, wight,
```
peternewman: I wonder if this should be disabled due to https://en.wikipedia.org/wiki/Isle_of_Wight

codespell_lib/data/dictionary.txt
```diff
@@ -1,4 +1,5 @@
 1nd->1st
+2st->2nd
```
peternewman: Is it worth adding some 3rd variants too?

codespell_lib/data/dictionary.txt
luzpaz: I added 
```
reprodicible->reproducible
reprodicibly->reproducibly
```

codespell_lib/data/dictionary.txt
luzpaz: Added 
```
3nd->3rd
3st->3rd
``

**Comment by luzpaz** [comment]
See 99c3f90 for changes

File: codespell_lib/data/dictionary.txt (hunk 8 above)
Comment (luzpaz) [reply codespell_lib/data/dictionary.txt]: i disagree because wight is a name (which can be diabled via the user using the `-L` flag). In my estimation this would most likely will be a typo

File: codespell_lib/data/dictionary.txt (hunk 7 above)
Comment (luzpaz) [reply codespell_lib/data/dictionary.txt]: added + added `teach`

File: codespell_lib/data/dictionary.txt (hunk 3 above)
Comment (arm-in) [reply codespell_lib/data/dictionary.txt]: Why? E.g. 3D lidar scanner for automated driving would deliver a point cloud (but 2 words)

File: codespell_lib/data/dictionary.txt (hunk 4 above)
Comment (arm-in) [reply codespell_lib/data/dictionary.txt]: possibly also "rent"

File: codespell_lib/data/dictionary.txt (hunk 3 above)
Comment (peternewman) [reply codespell_lib/data/dictionary.txt]: Yeah that's what I meant, I wouldn't object to two words, apart from Codespell can't currently do them, and poind->point, which isn't currently in there, so can we add it please @luzpaz .

Comment (luzpaz) [reply codespell_lib/data/dictionary.txt]: https://github.com/search?q=pointcloud&type=Code  
https://github.com/search?q=poindcloud&type=Code  
I think it merits being added. I will also add `poind->point`

Comment (peternewman) [reply codespell_lib/data/dictionary.txt]: Yeah okay you've convinced me, thanks for adding poind too.

File: codespell_lib/data/dictionary.txt (hunk 4 above)
Comment (peternewman) [reply codespell_lib/data/dictionary.txt]: Done.

File: codespell_lib/data/dictionary.txt (hunk 7 above)
Comment (peternewman) [reply codespell_lib/data/dictionary.txt]: Thanks for adding titch, teach seems a huge leap from tich to me, both phonetically and spelling wise.

File: codespell_lib/data/dictionary.txt (hunk 5 above)
Comment (peternewman) [reply codespell_lib/data/dictionary.txt]: Thanks, any reason for not adding the others? If dic->duc is a valid typo, the others are likely to occur sometimes too.

File: codespell_lib/data/dictionary.txt (hunk 8 above)
Comment (peternewman) [reply codespell_lib/data/dictionary.txt]: We've already got `tim->time, Tim, disabled due to being a person's name`. Personally I think we should be consistent, it will still potentially flag up depending on your config. Otherwise we're back in that world of having valid words in our dictionary of typos because they're "rare", which rather depends on what you're doing.

Comment (luzpaz) [reply codespell_lib/data/dictionary.txt]: It's your call. `tim` is a lot more problematic that `wight`. but based on your above logic I'm ok to disable `wight`.

File: codespell_lib/data/dictionary.txt (hunk 7 above)
Comment (luzpaz) [reply codespell_lib/data/dictionary.txt]: swapped `teach` for `stitch`

File: codespell_lib/data/dictionary.txt [hunk 10]
```diff
@@ -25706,7 +25754,7 @@ wiew->view
 wigdet->widget
 wigdets->widgets
 wighed->weighed, wicked,
-wight->weight, white, right, wight,
+wight->weight, white, right, write, wight, disabled due to being a person's name
```
larsoner: I'd rather not disable for names (maybe unless they are very common, but this one doesn't seem all that common)

codespell_lib/data/dictionary.txt
larsoner: I think that we can use some judgment here. At the 100% consistency end we'd have to search each "typo" to see if anyone anywhere has ever that name, and that doesn't make much sense. So it probably should be some sort of judgment call. In this case I think that the benefits of flagging the many errors that `wight` could represent outweight the costs of forcing an addition to `ignore_words` for people who want to use `wight` as a proper noun.

codespell_lib/data/dictionary.txt
luzpaz: removed disabled status 6799b3a

codespell_lib/data/dictionary.txt
luzpaz: Sorry, thought i mentioned it but the list above has repetitive entries. I just added the ones that were unique. Also forgot to mention I added `reprodicibility->reproducibility`

peternewman: Apologies, so it does!

PR #1457, opened 2020-03-28

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -6799,6 +6803,8 @@ custumised->customised
 custumized->customized
 cutom->custom
 cuurently->currently
+cuve->curve
```
arm-in: possibly also "cube", "cave"

codespell_lib/data/dictionary.txt
```diff
@@ -6305,6 +6308,7 @@ coponent->component
 copoying->copying
 coppermines->coppermine
 coppied->copied
+copping->coping, copying,
```
arm-in: also "cropping"

codespell_lib/data/dictionary.txt
```diff
@@ -23423,6 +23438,7 @@ theer->there
 theery->theory
 theese->these
 thefore->therefore
+thei->their
```
arm-in: possibly also "they"

codespell_lib/data/dictionary.txt
```diff
@@ -24759,6 +24775,7 @@ unsychronised->unsynchronised
 untargetted->untargeted
 unter->under
 untill->until
+untils->until
```
arm-in: possibly also "utils" (short for utilities)

luzpaz:
@arm-in applied your feedback to 1ac93c8

codespell_lib/data/dictionary.txt
```diff
@@ -6307,6 +6310,7 @@ coponent->component
 copoying->copying
 coppermines->coppermine
 coppied->copied
+copping->coping, copying, cropping,
```
larsoner: `copping` is a form of the verb `cop`, e.g., "copping an attitude" (not entirely obscure in English)

luzpaz: good point. Though, this is why I for a 2ndary fallback dictionary accessible with a flag that can catch some of these more problematic typos (that are legit words)

larsoner: For now flag as `disabled ...` then? Then once we have the secondary dictionary support we can move it

luzpaz: @larsoner 5959efe  
rebased as well.

PR #1475, opened 2020-04-07
Sort multiple dictionaries

peternewman:
Dictionary sorting, ensure we check they're all sorted and make sorts all of them.


larsoner:
Awesome, this is something I noticed but did not want to fix myself in that PR :)

PR #1476, opened 2020-04-07
Move some misspellings to dictionary_names.txt

qyearsley:
This includes the names originally in PR https://github.com/codespell-project/codespell/pull/1425.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1535,7 +1535,6 @@ androind->Android
 andthe->and the
 ane->and
 anevironment->environment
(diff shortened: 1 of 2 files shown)


peternewman:
I don't know if you want to move synopsys->synopsis into perhaps another dictionary of company names, so people can then choose to exclude them (but probably independently from people's names)?

qyearsley:
Thanks! Perhaps, although I don't know any more company names, so no need I thnk.

peternewman:
We can always start it with one, I'm sure more will come out of the woodwork, names only started with three...

PR #1478, opened 2020-04-09
Duplicated entry

Gelma:


codespell_lib/data/dictionary.txt
```diff
@@ -4056,7 +4056,6 @@ chacacters->characters
 chache->cache
 chached->cached
 chacheline->cacheline
-chack->check, chalk,
```
peternewman, line 4059: Oops, that needs catching in our tests.

codespell_lib/data/dictionary_names.txt
```diff
@@ -6,5 +6,3 @@ liszt->list
 que->queue
 sargent->sergeant, argent,
 tim->time
-wen->when
```
peternewman, line 9: As per @qyearsley this is a name, if it's also in the main dictionary it needs removing from there.

codespell_lib/data/dictionary_names.txt
```diff
@@ -6,5 +6,3 @@ liszt->list
...
-wight->weight, white, right, write,
```
peternewman, line 10: I pointed out Isle of Wight which I think this went in names.

See also these people:
https://en.wikipedia.org/wiki/Wight_(surname)

Perhaps we should split names into first and surname in terms of likelihood of clashes.

peternewman:
Hi @Gelma ,

#1501 now covers testing this and I've also fixed those issues I think, if you'd like to double-check as I see you added a few more bits too.

I've left Wen in names as it's a name, and chack in rare as it is technically a thing too.

Currently I've left Wight in rare as it's only a surname aside from the Isle of Wight, I guess we can see how we go with that.

Anyway if you'd like to update and either there will be no outstanding changes and this can be closed or it can be merged and closed.

bl-ue (commented):
What's the status of this PR @Gelma?

Issue #1302, opened 2019-10-17
create a flag that escapes special characters like apostrophes


luzpaz:
Usecase: A typo for the word `shouldnt` is found within a code block that is syntax sensitive and requires the result have an escaped character: `shouldn\'t` 

Right now codespell doesn't accommodate this. Either kill codespell and manually enter this in the code or make a note to edit after codespell has completed.

Proposal: a flag passed to codespell will scan through the dictionary and find any dictionary results that have special chars and will auto-magically provide an auto-escaped result appended to the regular results.

Currently in dictionary.txt: `shouldnt->shouldn't, should not,`  
I propose a flag like `codespell --escape-chars` which will append a dynamic escaped char as a result which turns the above in to:  
`shouldnt->shouldn't, should not, shouldn\'t`

lurch:
...except for languages like Python, you can either have `"I shouldn't complain"` or `'I shouldn\'t complain'`
(but obviously fixing that would need a proper context-aware lexer approach, rather than a simple regex approach).

I guess when the proposed `--escape-chars` flag is passed, you'd also want codespell to 'unescape' the input text, so that `shouldn\'t` in the input file gets spellchecked as though it was written `shouldn't`? (IYSWIM!)

PR #1482, opened 2020-04-14
Improve the default regular expression for words

### LINKED ISSUE #391: Example of --regex flag in use
Can someone demonstrate how to use the codespell --regex= flag?

dwo:
The `_` character was included via the `\w` character class. The default
regular expression now unrolls the `\w` and drops `_` from it.

This caused words in `snake_case` variables to be missed by codespell.

Resolves #391


dwo:
I'll update the regex test to cover this change. 🙇

README.rst
```diff
@@ -55,6 +55,11 @@ Comma-separated list of files to skip. It accepts globs as well.  Examples:
 
 * to skip directories, invoke ``codespell --skip="./src/3rd-Party,./src/Test"``
 
+The ``-r`` flag can be used to change the regular expression which is used to find words. ::
+
+    codespell -r, --regex=
+
+* to restore previous behaviour of including `_` in words, invoke ``codespell --regex="[\w\-'’\`]+"``
```
dwo, line 62: I spent quite a while getting confused about what has to be escaped: it seems like you can get away with less escaping of this option in `bash` than you can in `sh`? Not sure exactly what is best to put here 😅

codespell_lib/tests/test_basic.py
```diff
@@ -198,7 +198,7 @@ def test_custom_regex(tmpdir, capsys):
     """Test custom word regex."""
     d = str(tmpdir)
     with open(op.join(d, 'bad.txt'), 'w') as f:
-        f.write('abandonned_abondon\n')
+        f.write('abandonned-abondon\n')
```
dwo, line 201: I just kept the existing test working with a dash rather than an underscore. I could add a test specifically for bringing back underscores as per the addition to README.rst

codespell_lib/data/dictionary.txt
```diff
@@ -4475,7 +4475,6 @@ cloacks->cloaks
 cloberring->clobbering
 clocksourc->clocksource
 clockwíse->clockwise
-clock_getttime->clock_gettime
```
peternewman, line 4478: You can add getttime instead on it's own can't you?

README.rst
```diff
@@ -55,6 +55,11 @@ Comma-separated list of files to skip. It accepts globs as well.  Examples:
 
 * to skip directories, invoke ``codespell --skip="./src/3rd-Party,./src/Test"``
 
+The ``-r`` flag can be used to change the regular expression which is used to find words. ::
+
+    codespell -r, --regex=
+
+* to restore previous behavior of including `_` in words, invoke ``codespell --regex="[\w\-'’\`]+"``
```
peternewman, line 62: It would be nice to note which version it changed in.

codespell_lib/data/dictionary.txt
peternewman: Did the tests flag this up or did you check?

codespell_lib/data/dictionary.txt
```diff
@@ -18232,7 +18233,6 @@ physicaly->physically
 physisist->physicist
 phython->python
 phyton->python
-phy_interace->phy_interface
```
peternewman, line 18235: Likewise here, add interace?

codespell_lib/data/dictionary.txt
dwo: It was already there (and I think was automatically deduped by the check) 😄

codespell_lib/data/dictionary.txt
dwo: Same as before, was already present.

README.rst
dwo: Sure, I'll add that.

dwo:
> I use a slightly more complex regex to catch quoted words and stuff too

Perhaps we should discuss such a change in another PR? Seems like the benefit of splitting on underscores is high-value enough on its own, and keeps this change relatively simple (since there were only a handful of underscored words in the dictionary).

I can see the benefit of catching quoted misspellings like you mention, but you would have to make some compromise for entries like this:
https://github.com/codespell-project/codespell/blob/2d12d7a30a5b0dbbd1d9f8b919a6c0658df9228e/codespell_lib/data/dictionary.txt#L905 ...and perhaps others too?

Or maybe we'd actually prefer to lump several "breaking" regex changes into one release? I'm not sure.

dwo:
Seems like there's 26 dictionary entries that have misspellings ending in `'` and I couldn't find any others starting or ending with non-alphanumeric characters. Perhaps losing these 26 corrections would be a small price to pay for all the potential corrections in single quotes?

You could also allow the few 2-letter corrections in dictionary.txt by making the regex
```
[a-zA-Z0-9][a-zA-Z0-9\\-'’`]*[a-zA-Z0-9]
```

dwo:
Also gets me thinking: should we add a test that every error in the dictionaries matches the default regex?

codespell_lib/data/dictionary.txt
peternewman: :facepalm: I swear I checked at least one of these!

codespell_lib/data/dictionary.txt
peternewman: Doh!

README.rst
peternewman: Does the `sh` escaping work in `bash` too?

peternewman:
> > I use a slightly more complex regex to catch quoted words and stuff too
> 
> Perhaps we should discuss such a change in another PR? Seems like the benefit of splitting on underscores is high-value enough on its own, and keeps this change relatively simple (since there were only a handful of underscored words in the dictionary).

Yeah agreed.

Also as per my comment here, my proposal has some issues too:
https://github.com/codespell-project/codespell/issues/391#issuecomment-385183552

> I can see the benefit of catching quoted misspellings like you mention, but you would have to make some compromise for entries like this:
> https://github.com/codespell-project/codespell/blob/2d12d7a30a5b0dbbd1d9f8b919a6c0658df9228e/codespell_lib/data/dictionary.txt#L905
> 
> ...and perhaps others too?
> Or maybe we'd actually prefer to lump several "breaking" regex changes into one release? I'm not sure.

Yes indeed. See also the issue #1302 . While being in the same release might make sense, they could certainly be different PRs as you suggested.

> Seems like there's 26 dictionary entries that have misspellings ending in `'` and I couldn't find any others starting or ending with non-alphanumeric characters. Perhaps losing these 26 corrections would be a small price to pay for all the potential corrections in single quotes?

With our multi-dictionary approach, we could add a dictionary of those type of spellings for people to either manually or automatically include.

> You could also allow the few 2-letter corrections in dictionary.txt by making the regex this:
> 
> ```
> [a-zA-Z0-9][a-zA-Z0-9\\-'’`]*[a-zA-Z0-9]
> ```

Not sure why I didn't think of that, I can't see a failure with it!

> Also gets me thinking: should we add a test that every error in the dictionaries matches the default regex?

That would seem very sensible, well volunteered! :laughing:

cas--:
I was testing Python code with codespell and was surprised to find that it ignored any string which was enclosed with single quotes, compared to double-quoted strings always being checked. After a bit of searching I see that the `word_regex_def` is looking for all single quotes as apostrophes!

I think I saw a passing mention of fixing it here and the example regex expressions does work for me. Hopefully this improvement can be included, (along with some tests ;).

dwo:
Sorry that I have left this unfinished for so long: I was doing it at work and have since gone on furlough due to the pandemic.

bl-ue:
+1 for removing the apostrophe suggestions so as to check for typos in single quotes.  Actually +100 😆

bl-ue:
@dwo would you mind if I took this back up?

dwo:
> @dwo would you mind if I took this back up?

Go for it :smile:

bl-ue:
@peternewman @larsoner @luzpaz why do you think about removing the apostrophe examples in order to check typos among single quotes?

peternewman:
> @peternewman @larsoner @luzpaz why do you think about removing the apostrophe examples in order to check typos among single quotes?

I don't think my opinion has changed:

> With our multi-dictionary approach, we could add a dictionary of those type of spellings for people to either manually or automatically include.

They're all valuable typos. If you're just checking words (not code), you don't care about the single quotes. Or the ideal is you can check against both dictionaries with different regex.

bl-ue:
Sounds good.  What would you call the dictionary? `dictionary_apostrophes.txt`?

peternewman:
Yeah something like that, without having see exactly what all the typos are.

sebweb3r:
any progress regarding this PullRequest?

@bl-ue

bl-ue:
I'm sorry @sebweb3r - I can't continue on this :(

codespell_lib/_codespell.py
```diff
@@ -26,7 +26,7 @@
 import re
 import sys
 
-word_regex_def = u"[\\w\\-'’`]+"
+word_regex_def = u"[a-zA-Z0-9\\-'’`]+"
```
DimitriPapadopoulos, line 29: What about the `é` of `attaché` or `fiancée`?

dwo: I guess that is why we'd need the test that ensures the default word regex matches everything in the dictionaries

PR #1487, opened 2020-04-21
Add extit(ing)->exit(ing)

cfi-gb:
See e.g.

https://github.com/search?q=extiting&type=Code
https://github.com/search?q=extit&type=Commits

There seems to be an "EXTIT" library but i hope this is too uncommon and one using that lib can exclude it in their project.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -11206,6 +11206,8 @@ extesnion->extension
 extesnions->extensions
 extint->extinct, extant,
 extist->exist
+extit->exit
+extiting->exiting
 extnesion->extension
 extrac->extract
 extraced->extracted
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -11206,6 +11206,8 @@ extesnion->extension
 extesnions->extensions
 extint->extinct, extant,
 extist->exist
+extit->exit
+extiting->exiting
```
peternewman, line 11210: A few more bits to add:
```suggestion
extited->excited, exited,
extiting->exciting, exiting,
extits->exits, excites,
```

cfi-gb: Thanks, done in ff6ebe1fbb9eae34ac8d1734e7f5a43b61feb522

PR #1494, opened 2020-04-30
Add several spelling corrections from XaoS

arm-in:


arm-in:
Build failed, but I don't know why. Please advise!

codespell_lib/data/dictionary.txt
```diff
@@ -23308,6 +23316,7 @@ spoonfulls->spoonfuls
 sporatic->sporadic
 sporious->spurious
 sppeches->speeches
+spped => speed, skipped, swapped, stripped,
```
larsoner, line 23319: Need to look at the [error log](https://travis-ci.org/github/codespell-project/codespell/jobs/681582401#L537-L540):
```
>               err, rep = line.decode('utf-8').split('->')
538E               ValueError: not enough values to unpack (expected 2, got 1)
539
540err        = 'sppeches'
```
This suggests that something in the vicinity of `sppeches` does not contain a `->`, looks like it's this one

arm-in: Looks better now. Thank you!

codespell_lib/data/dictionary.txt
```diff
@@ -23308,6 +23318,7 @@ spoonfulls->spoonfuls
 sporatic->sporadic
 sporious->spurious
 sppeches->speeches
+spped->speed, skipped, swapped, stripped,
```
peternewman, line 23321: Suggest sped too. To me the s*pped ones seem rather unlikely. If you were going to keep them, you should probably add stopped etc, but personally I think we should remove them.

codespell_lib/data/dictionary.txt
```diff
@@ -24790,6 +24801,7 @@ thiknesses->thicknesses
 thikning->thinking, thickening,
 thikns->thinks
 thiks->thinks
+thime->time, theme, thine,
```
peternewman, line 24804: Suggest thyme the herb

codespell_lib/data/dictionary.txt
arm-in: Well, "speed" and "sped" will not fit in the case of XaoS.
The other words are imho just suggestions, and not necessarily exhaustive.

arm-in:
Ready to merge?

codespell_lib/data/dictionary.txt
```diff
@@ -16471,6 +16478,7 @@ mystql->mysql
 Mythraic->Mithraic
 myu->my
 nadly->badly
+naerly->nearly
```
peternewman, line 16481: Can you also suggest gnarly

codespell_lib/data/dictionary.txt
peternewman: Personally I'm not really in favour of them all still being there. The fact they are two or more different missing letters seems so unlikely. Or are you saying you've seen them in real life?

sapped should be added too as it seems far more plausible than them to me. Although TBH as soon as there are two alternatives it won't go through automatically in interactive mode.

codespell_lib/data/dictionary.txt
arm-in: Done.

codespell_lib/data/dictionary.txt
arm-in: I don't see the problem.

"spped" is real life from the "Xaos" project.

"spped" -> "skipped" is Levenshtein distance 2, just like your addition above
"naerly" -> "gnarly"

Judge with the same yardstick, please.

peternewman: > "spped" is real life from the "Xaos" project.

Yes, but presumably the correction was to speed?
 
> "spped" -> "skipped" is Levenshtein distance 2, just like your addition above
> "naerly" -> "gnarly"
> 
> Judge with the same yardstick, please.

Mine is phonetic (in the same way as knead->need would be Levenshtein 2, but a more likely real world typo than say knead->plead which has the same Levenshtein), whereas from what I can see all of yours are traditional typos and suggestions.

arm-in: So, what is your expectation now?

1) I have to correct it?
2) I have to discuss further?
3) You correct it?
4) You merge?

peternewman: You haven't clarified what your spped in Xaos corrected to.

Personally phonetic and closer Levenshtein seem more likely to me, such as "sped, speed, sapped, sipped, supped", especially compared to say `stripped` which is Levenshtein 3.

Personally I'd like you to remove those Levenshtein 2/3 for these and then we could merge it, but it's not a dictatorship, others may feel differently or be happy to merge anyway, or you might be able to convince me it should go in anyway.

arm-in: I've no idea what "spped" should correct to. Maybe "skipped", but the text passage is hard to read.

I've removed stripped = Levenshtein 3.

c72578: The `spped` riddle: https://github.com/xaos-project/XaoS/blob/6df9acc5cbe31e5d609f1e4fd6d7a8f98eeb28cf/src/engine/docalc.h#L267

peternewman: Thanks @c72578 .

I guess this is the challenge of a presumably Frenchman writing in English:
https://github.com/xaos-project/XaoS/blob/6df9acc5cbe31e5d609f1e4fd6d7a8f98eeb28cf/src/engine/docalc.h#L250

My best guess of interpreting this paragraph would be (original):
```
/*F. : This is from then lim factor that depends all periodicity check spped :
   the bigger it is, the faster we can detect periodicity, but the bigger it is,
   the more we can introduce errors. I suggest a value of
   (maxx-minx)/(double)getmaxx for a classic Mandelbrot Set, and maybe a lesser
   value for an extra power Mandelbrot. But this should be calculated outer
   from here (ie each frame, for example), to avoid new calculus */
```

My best guess at corrected:
```
/*F. : This is from the lim factor that controls all periodicity check speed :
   the bigger it is, the faster we can detect periodicity, but the bigger it is,
   the more we can introduce errors. I suggest a value of
   (maxx-minx)/(double)getmaxx for a classic Mandelbrot Set, and maybe a lesser
   value for an extra power Mandelbrot. But this should be calculated outside
   here (ie each frame, for example), to avoid new calculations */
```

I'm not sure if lim should be limit too.

Looking at context, I'm pretty confident it's speed. Indeed looking at the first five pages of GitHub search results for `spped` I'd say 75%+ look like they are speed and maybe 25% are definitely speed, as between the repo name and the title/description, one has the typo and one doesn't. Specifically I can't find any places where it would be anything other than sped or speed.

arm-in: I removed all those rather comprehensive alternatives.
As you've already said, corrections won't go through automatically, anyway.

But hey, those are just ideas or suggestions. They are neither exact nor rocket science nor programming. No memory leaks, null pointers, buffer overflows. Reviewing the correction suggestions of a dictionary (not even the dictionary itself) could be more generous.

arm-in:
Ready to merge?

PR #1496, opened 2020-05-05
dictionary_rare: add unsecure->insecure

waldyrious:


larsoner:
I'm on the fence on this one. Especially since `codespell` is ideally for code, and `unsecure` is a fairly standard computing term, even `rare` seems like it might not be right (it probably won't actually be that rare for downstream users).

waldyrious:
Happy to close this PR if you prefer not to add it. I don't have a strong opinion on the matter. I was myself not familiar with "unsecure" when I bumped into it, and at first thought it was a typo, then I figured that [Wiktionary declares it](https://en.wiktionary.org/wiki/unsecure) a rare spelling of "insecure", hence this PR.

bl-ue (approved):
+1 for putting in it rare — I've never even seen it before. (only `insecure`).  I'll merge this old PR in a week if there's no objections by then.

larsoner:
"unsecured debt," is a common phrase in lending, but I agree even that form of unsecure would be comparatively rare enough to go in rare

bl-ue:
Thank you @waldyrious :D

PR #1502, opened 2020-05-11
Add flake8 annotions

peternewman:
Via https://github.com/TrueBrain/actions-flake8

Blocked behind #1501 as I wanted some test failures!


larsoner:
Seems to work nicely, now just need to fix the error

peternewman:
This is ready for review, with one obvious fault so you can see it works! :rofl:

peternewman:
Should be green now @larsoner so merge if you're happy. Sorry I missed you'd commented at exactly the same time as me before.

I guess in future we might want to tune it to not annotate twice when it's within the main codespell repo somehow.

larsoner:
I usually make all changes in my fork and open PRs. It's a better and more standard practice in my experience than making branches in the main repo

peternewman:
> I usually make all changes in my fork and open PRs. It's a better and more standard practice in my experience than making branches in the main repo

Yeah I would do too normally if I was doing it properly via the CLI. Irritatingly GitHub takes that option away if you've got edit rights and you're just doing it lazily via the website. I just get:

- You can’t commit to master because it is a protected branch.
- Create a new branch for this commit and start a pull request.

Let me know if it's really bugging you though and I'll try and do the overhead to keep one of my branches in sync enough to make edits.

larsoner:
No the branches are okay, just make sure to periodically remove them to ensure they don't accumulate. It's not great when people `clone` the upstream repo and end up with a ton of branches

peternewman:
> No the branches are okay, just make sure to periodically remove them to ensure they don't accumulate.

I'll just have to make sure I keep on top of the features then!

> It's not great when people `clone` the upstream repo and end up with a ton of branches

Yeah agreed.

PR #1513, opened 2020-05-20
Add fomrat->format and friends

peternewman:


larsoner:
@peternewman I'm finding the requirement to update each branch and also require an approved review to be a bit burdensome. It adds some minutes to each review+merge cycle, when before the approval could be implicit just by merging. And we haven't had many cases where not being up to date when merging was a problem -- conflicts will be caught anyway, which should flag ordering problems, and collisions with corrections from one PR being errors in another should be low. Okay to remove these restrictions?

peternewman:
> @peternewman I'm finding the requirement to update each branch and also require an approved review to be a bit burdensome. It adds some minutes to each review+merge cycle, when before the approval could be implicit just by merging.

The only thing is this wouldn't stop us just merging our own PRs, either accidentally or on purpose.

> And we haven't had many cases where not being up to date when merging was a problem -- conflicts will be caught anyway, which should flag ordering problems, and collisions with corrections from one PR being errors in another should be low.

It does mean new CI improvements will only get run when a PR is merged (therefore giving us the burden to fix them).

> Okay to remove these restrictions?

I've removed them for now. See if that seems any better.

peternewman:
I've added "Include administrators" back on @larsoner as with the last set of changes there was nothing to stop me accidentally editing master directly.

PR #1517, opened 2020-05-22
FIX: Packaging and 1.17.1 bump

### LINKED ISSUE #1516: dictionary_rare missing
`codespell_lib/data/dictionary_rare.txt` is missing in the 1.17.0 PyPI source package causing a crash when all builtin dictionaries are used.

larsoner:
Closes #1516.

I already pushed 1.17.1 to PyPI after checking that the sdist actually had the dictionaries in it from the `MANIFEST.in` changes. But `check-manifest` should prevent us from having this problem in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,8 +29,8 @@ before_install:
     - virtualenv --python=python venv
     - source venv/bin/activate
     - python --version  # just to check
-    - pip install -U pip wheel  # upgrade to latest pip find 3.5 wheels; wheel to avoid errors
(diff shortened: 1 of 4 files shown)


c72578:
Will there be a release of 1.17.0 and 1.17.1 at github too?
https://github.com/codespell-project/codespell/releases

larsoner:
@c72578 done

larsoner:
(at least for 1.17.1 -- 1.17.0 was quite broken so I don't see a point)

PR #1535, opened 2020-05-31
Add note about not being a general spell checking tool.

tir38:
I spent *way* too much time figuring this out the hard way.


peternewman:
Out of interest @tir38 what was missing to make it a general spell checking tool? What makes you think it isn't one?

tir38:
I tried to test this by just barfing rando text into a file (e.g. `asdadaflkjasd`) and codespell didn't report this as a misspelling

larsoner:
It looks for common spelling mistakes rather than membership in a complete dictionary. Maybe it can be rephrased to take this into account?

tir38:
@larsoner yep, that was what I hoped to accomplish by updating the README

README.rst
```diff
@@ -3,6 +3,7 @@ codespell
 
 Fix common misspellings in text files. It's designed primarily for checking
 misspelled words in source code, but it can be used with other files as well.
+It is not a general spell checker.
```
larsoner, line 6: How about something like this?
```suggestion
It does not check for word membership in a complete dictionary, but instead
looks for a set of common misspellings. Therefore it might catch errors like
"adn", but it will not catch "adnasdfasdf".
```

peternewman: Perhaps it should be "a theoretically complete" dictionary?

peternewman: Could we also make it a bit more positive, by saying "so also shouldn't generate false-positives when you use a niche term it doesn't know about".

README.rst
```diff
@@ -3,6 +3,9 @@ codespell
 
 Fix common misspellings in text files. It's designed primarily for checking
 misspelled words in source code, but it can be used with other files as well.
+It does not check for word membership in a complete dictionary, but instead
+looks for a set of common misspellings. Therefore it might catch errors like
+"adn", but it will not catch "adnasdfasdf".
```
tir38, line 9: Derp. Talk about circular dependency. Using Codespell to check readme of repo finds spelling error, which only exists to show example of spelling error. Not sure if I should fix this.

README.rst
```diff
@@ -3,6 +3,9 @@ codespell
...
```
peternewman, line 8: I still think it would be nice to make it a bit more positive, for example by saying "it also shouldn't generate false-positives when you use a niche term it doesn't know about".

README.rst
peternewman: Well at least we know it's working @tir38 ! Joking apart we could either ignore the line or add the specific typo to an allowed list (although if we do the latter it would be better if it was a more obscure word/less likely typo, but perhaps that defeats the point of the example; `abandonned` is our goto typo, but it's not as obvious as the example one).

@larsoner Travis didn't catch this as it's only checking the module source stuff, not the surrounding files:
`codespell --skip="codespell_lib/tests/test_basic.py,codespell_lib/data/*" codespell_lib/`

larsoner: Feel free to add it to the checks

peternewman: Done in #1596

peternewman:
Sorry this slipped by the wayside @tir38 . I think it would still make sense to get it in.

To avoid the error it's now trapping (due to #1596), we'll need to skip it, the safest way is probably adding an exclude file into 
https://github.com/codespell-project/codespell/blob/master/.github/workflows/codespell.yml
For the syntax, see:
https://github.com/codespell-project/actions-codespell#parameter-exclude_file

README.rst
peternewman: Improved suggestion.
```suggestion
"adn", but it will not catch "adnasdfasdf". This also means it shouldn't
generate false-positives when you use a niche term it doesn't know about.
```

README.rst
```diff
@@ -3,6 +3,9 @@ codespell
 
 Fix common misspellings in text files. It's designed primarily for checking
 misspelled words in source code, but it can be used with other files as well.
+It does not check for word membership in a complete dictionary, but instead
+looks for a set of common misspellings. Therefore it might catch errors like
```
peternewman, line 7: More positivity
```suggestion
looks for a set of common misspellings. Therefore it should catch errors like
```

README.rst
peternewman: See https://github.com/codespell-project/codespell/pull/1535#issuecomment-782903268 for more detail of fixes.

peternewman (changes requested):
Sorry this slipped by the wayside @tir38 . I think it would still make sense to get it in, do you want to have a go. I think the following changes should do the trick.

bl-ue (approved):
🎉

peternewman (commented):
Sorry @bl-ue , we can't just skip the whole of README.rst because of one (deliberate) typo! How daft is codespell going to look when it ships a readme with typos in it?!?

The ideal solution is to check readme separately and ignore just the one or two deliberate typos (via ignore words). The more general solution for other people where it's less critical would be to ignoreit across the entire codebase. We should really use the ideal solution given what we're representing!

peternewman:
Are you happy to re-engineer the CI tests then after this @bl-ue , or do you want me to?

bl-ue:
>Sorry @bl-ue , we can't just skip the whole of README.rst because of one (deliberate) typo! How daft is codespell going to look when it ships a readme with typos in it?!?

I'm sorry! That's what we get with my eager merging 😝😆

>Are you happy to re-engineer the CI tests then after this @bl-ue , or do you want me to?

You mean update the CI to only ignore the specific typos in the readme?  Sure, I'll do it!

peternewman:
> I'm sorry! That's what we get with my eager merging stuck_out_tongue_closed_eyeslaughing

:smile: 

> > Are you happy to re-engineer the CI tests then after this @bl-ue , or do you want me to?
> 
> You mean update the CI to only ignore the specific typos in the readme? Sure, I'll do it!

The gold-plated solution would probably be best, skip the readme as you've done in the main run, but add another run which checks just the readme but ignores the additional deliberate typos, then we won't get "adn" all over the rest of the codebase.

bl-ue:
Good idea. I'll do it 👍🏻

tir38:
thanks y'all!

bl-ue:
No problem @tir38 — thank _you_ for evaluating codespell!  I'd recommend some other tools that are probably more what you're thinking of when you hear spellchecker:

- https://github.com/streetsidesoftware/cspell
- https://github.com/GNUAspell/aspell
- https://github.com/hunspell/hunspell

peternewman:
> No problem @tir38 — thank _you_ for evaluating codespell! I'd recommend some other tools that are probably more what you're thinking of when you hear spellchecker:

:laughing: You're not supposed to plug the competition @bl-ue ! :rofl:

bl-ue:
Hahaha — sorry @peternewman! 😂

PR #1537, opened 2020-06-01

pabs3:


codespell_lib/data/dictionary.txt
```diff
@@ -21220,14 +21272,34 @@ reguarlize->regularize
...
+regularlises->regularises
+regularlising->regularising
+regularlization->regularization
+regularlize->regularize
+regularlized->regularized
+regularlizer->regularizer
+regularlizes->regularizes
+regularlizing->regularizing
+regularlly->regularly
+regulary->regular, regularly,
```
larsoner, line 21296:
```
AssertionError: error regulariry: correction regulary is an error itself in the same dictionary file dictionary.txt
```

pabs3:
Fixed the exception.

-- 
bye,
pabs

https://wiki.debian.org/PaulWise

codespell_lib/data/dictionary.txt
```diff
@@ -21220,14 +21272,34 @@ reguarlize->regularize
...
+regulax->regular
+reguler->regular
+regulr->regular
 regultor->regulator
 regultors->regulators
+regultory->regulatory
 regurlarly->regulary
```
larsoner, line 21303: AssertionError: error regurlarly: correction regulary is an error itself in the same dictionary file dictionary.txt

pabs3:
Also fixed.

-- 
bye,
pabs

https://wiki.debian.org/PaulWise

codespell_lib/data/dictionary.txt
peternewman: Another one for the problem matcher...

pabs3:
Thats fixed.

-- 
bye,
pabs

https://wiki.debian.org/PaulWise

larsoner:
Grr, Travis will not report:

1. Closing-reopening did not get it to report
2. I can't override the requirement to merge
3. I can't push directly to `master` (after merging on the command line) because it's not allowed (I tried) 

@peternewman what do we do here?

peternewman:
> Grr, Travis will not report:
> 
>     1. Closing-reopening did not get it to report

If you sign into Travis you can restart tests against this PR, which looks promising as GitHub knows it's pending.

If that doesn't work, the next best bet would be Settings->Branches->Branch protection rules and untick the Travis status check.

>     2. I can't override the requirement to merge
> 
>     3. I can't push directly to `master` (after merging on the command line) because it's not allowed (I tried)

Alternatively unticking "Include administrators" in the same place would let you override either of the above options, but I'd recommend we put it back on afterwards as otherwise I end up editing master by mistake! :facepalm:

peternewman:
Indeed Travis has reported back green so I'll merge it.

codespell_lib/data/dictionary.txt
```diff
@@ -7823,13 +7824,41 @@ depandent->dependent
 deparment->department
 deparmental->departmental
 deparments->departments
+depden->depend
+depdence->dependence
```
peternewman, line 7828: It would be worth adding the in- versions of these typos too where they aren't already in the dictionary, e.g. indepdence->independence

codespell_lib/data/dictionary.txt
```diff
@@ -20116,6 +20167,7 @@ provsioning->provisioning
 proximty->proximity
 prozess->process
 prpeparations->preparations
+prposiing->proposing
```
peternewman, line 20170: Would be good to add prpose->propose and the -d and -r versions too.

pabs3:
I've added the newly suggested corrections.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

peternewman:
Thanks @pabs3 .

PR #1548, opened 2020-06-06

pabs3:


pabs3:
Dropped the typo correction that contained Unicode, since that was breaking the travis-ci aspell tests, which seem to require latin-1 encoding.

Issue #1549, opened 2020-06-09
Hidden directories


peternewman:
Currently this generates some warnings

codespell --check-filenames --skip ./codespell_lib/data,./example/code.c,test_basic.py,*.pyc
##[warning]WARNING: Binary file: ./.git/index 

Which implies we're not ignoring hidden directories.

I don't know if we should or not? Should this follow --check-hidden so hidden files and directories are treated the same?

peternewman:
See also some discussion in #1550:
> Yes probably, though maybe we should make it that you can do `--check-hidden files` or `--check-hidden directories` to do one or the other, and `--check-hidden` does both, and omitting `--check-hidden` ignores both. That way if someone wants to get the old behavior back they would do `--check-hidden directories`.

_Originally posted by @larsoner in https://github.com/codespell-project/codespell/pull/1550#issuecomment-641645556_

> Related discussion in #783.

_Originally posted by @waldyrious in https://github.com/codespell-project/codespell/pull/1550#issuecomment-641924220_

Issue #783, opened 2018-12-18
Ignore .svn and .git directories per default


arm-in:
Hello,

I would like to ignore hidden directories like .svn or .git per default

Rationale:
* Yes, ignoring directories is possible via option -S, --skip or using xargs (see #60 and #99)
* Hidden files have to be enabled via option -H, --check-hidden
* However, hidden directories like .svn and .git are checked per default. This sounds inconsequent / illogical

luzpaz:
This makes sense to me. Though for the purists, we could add a flag to enable search `.svn` and `.git` files.

larsoner:
I agree that if hidden files are not checked, hidden dirs should not be, either. Then `-H` should enable both.

luzpaz:
:+1:

akien-mga:
I came here to ask for this feature too, I was surprised when seeing this in my log right now:
```
$ codespell -w -q 3 --skip="./thirdparty,./editor/translations" -I ../godot-word-whitelist.txt
[...]
./.git/logs/HEAD:50: Readd  ==> Readd, re-add, read
./.git/logs/HEAD:600: Readd  ==> Readd, re-add, read
./.git/logs/HEAD:604: Readd  ==> Readd, re-add, read
./.git/logs/HEAD:685: Readd  ==> Readd, re-add, read
./.git/logs/HEAD:687: Readd  ==> Readd, re-add, read
./.git/logs/HEAD:688: Readd  ==> Readd, re-add, read
FIXED: ./.git/logs/HEAD
./.git/logs/refs/heads/master:279: Readd  ==> Readd, re-add, read
```

mandarvaze:
`.venv` folder is also being checked, which does not make sense. IMO, it should also be skipped, and needs to be explicitly checked via `-H` option

dseight:
As a workaround it's possible to add shell alias like this one:
```bash
alias codespell="codespell --skip='.git,.svn,.venv'"
```
#1351 should help with extending per-project ignore list

arm-in:
See also 166147f5765ce35d77cbf1e20b495ce640314c16

PR #1550, opened 2020-06-09
Ignore the .git folder

peternewman:
See also #1549 I don't know if this is a Codespell bug or a config error on our part. Thoughts?

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -13,4 +13,4 @@ jobs:
         with:
           check_filenames: true
           # When using this Action in other repos, the --skip option below can be removed
-          skip: ./codespell_lib/data,./example/code.c,test_basic.py,*.pyc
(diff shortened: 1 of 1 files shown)


larsoner:
Can't hurt to add it here, even if we change things later to automatically ignore `.` directories, so let's merge

peternewman:
> Can't hurt to add it here, even if we change things later to automatically ignore `.` directories

Do you think we should ignore `.` directories too then? It would seem to match e.g. `ls`'s behaviour.

larsoner:
Yes probably, though maybe we should make it that you can do `--check-hidden files` or `--check-hidden directories` to do one or the other, and `--check-hidden` does both, and omitting `--check-hidden` ignores both. That way if someone wants to get the old behavior back they would do `--check-hidden directories`.

waldyrious:
Related discussion in #783.

PR #1551, opened 2020-06-09

pabs3:


larsoner:
I think it's time to retire Python 2.7 from the CIs

PR #1554, opened 2020-06-10
Add renaming advice based on Linux kernel spelling

hadess:
Recommend the renaming of whitelist/blacklist to allowlist/blocklist as
per kernel patch:
https://lore.kernel.org/lkml/20200609122549.26304-1-sjpark@amazon.com/


hadess:
The failing test is because codespell expects to not find any source words in the dictionary. That might need to be relaxed somewhat to implement this recommendation feature.

PR #1556, opened 2020-06-10
Add renaming advice based on IETF draft-knodel-terminology-00

hadess:
Based on:
https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.1.1

Offer to replace master/slave with:
- Primary-secondary
- Leader-follower
- Active-standby
- Primary-replica
- Writer-reader
- Coordinator-worker
- Parent-helper

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,11 +41,11 @@ The ``-w`` flag will actually implement the changes recommended by codespell. No
 
     codespell -I FILE, --ignore-words=FILE
 
-The ``-I`` flag can be used to whitelist certain words that are in the ``codespell_lib/data/dictionary.txt``. The format of the whitelist file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said whitelist. **Important note:** The whitelist passed to ``-I`` is case-sensitive based on how it is listed in ``dictionary.txt``. ::
+The ``-I`` flag can be used for a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said list of allowed words. **Important note:** The list passed to ``-I`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
 
     codespell -L word1,word2,word3,word4
 
-The ``-L`` flag can be used to whitelist certain words that are comma-separated placed immediately after it. ::
+The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it. ::
 
     codespell -S, --skip=
 
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -43,7 +43,7 @@ _builtin_dictionaries = (
     ('clear', 'for unambiguous errors', '', False, None),
     ('rare', 'for rare but valid words', '_rare', None, None),
     ('informal', 'for informal words', '_informal', True, True),
-    ('usage', 'for recommended terms', '_usage', True, None),
+    ('usage', 'for recommended terms', '_usage', None, None),
     ('code', 'for words common to code and/or mathematics', '_code', None, None),  # noqa: E501
     ('names', 'for valid proper names that might be typos', '_names', None, None),  # noqa: E501
     ('en-GB_to_en-US', 'for corrections from en-GB to en-US', '_en-GB_to_en-US', True, True),  # noqa: E501
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -43,6 +43,7 @@ _builtin_dictionaries = (
     ('clear', 'for unambiguous errors', '', False, None),
     ('rare', 'for rare but valid words', '_rare', None, None),
     ('informal', 'for informal words', '_informal', True, True),
+    ('usage', 'for recommended terms', '_usage', True, None),
     ('code', 'for words common to code and/or mathematics', '_code', None, None),  # noqa: E501
     ('names', 'for valid proper names that might be typos', '_names', None, None),  # noqa: E501
     ('en-GB_to_en-US', 'for corrections from en-GB to en-US', '_en-GB_to_en-US', True, True),  # noqa: E501
--- a/codespell_lib/tests/test_dictionary.py
+++ b/codespell_lib/tests/test_dictionary.py
@@ -175,6 +175,8 @@ def test_error_checking_in_aspell(err, rep, err_aspell, rep_aspell, match):
 allowed_dups = {
     ('dictionary.txt', 'dictionary_en-GB_to_en-US.txt'),
     ('dictionary.txt', 'dictionary_rare.txt'),
+    ('dictionary.txt', 'dictionary_usage.txt'),
+    ('dictionary_rare.txt', 'dictionary_usage.txt'),
 }
 
 
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3556,7 +3556,7 @@ bizzare->bizarre
 bject->object
 bjects->objects
 blackslashes->backslashes
-blaclist->blocklist
+blaclist->blacklist
 blaim->blame
 blaimed->blamed
 blak->black, blank,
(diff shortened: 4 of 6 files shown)


hadess:
Apologies for the spam, my distro doesn't have aspell packaged up, and I thought I could get away without using a virtualenv...

codespell_lib/tests/test_dictionary.py
```diff
@@ -175,6 +175,8 @@ def test_error_checking_in_aspell(err, rep, err_aspell, rep_aspell, match):
 allowed_dups = {
     ('dictionary.txt', 'dictionary_en-GB_to_en-US.txt'),
     ('dictionary.txt', 'dictionary_rare.txt'),
+    ('dictionary.txt', 'dictionary_usage.txt'),
+    ('dictionary_rare.txt', 'dictionary_usage.txt'),
```
peternewman, line 179: I don't think this one is required now is it since you removed the terms from there?

codespell_lib/_codespell.py
```diff
@@ -43,6 +43,7 @@
     ('clear', 'for unambiguous errors', '', False, None),
     ('rare', 'for rare but valid words', '_rare', None, None),
     ('informal', 'for informal words', '_informal', True, True),
+    ('usage', 'for recommended terms', '_usage', None, None),
```
peternewman, line 46: Shouldn't this be True, True (as master will be in Aspell), or does it need to be True, None if allowlist isn't in aspell?

README.rst
```diff
@@ -41,11 +41,11 @@ The ``-w`` flag will actually implement the changes recommended by codespell. No
 
     codespell -I FILE, --ignore-words=FILE
 
-The ``-I`` flag can be used to whitelist certain words that are in the ``codespell_lib/data/dictionary.txt``. The format of the whitelist file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said whitelist. **Important note:** The whitelist passed to ``-I`` is case-sensitive based on how it is listed in ``dictionary.txt``. ::
+The ``-I`` flag can be used to allowlist certain words that are in the ``codespell_lib/data/dictionary.txt``. The format of the allowlist file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said allowlist. **Important note:** The allowlist passed to ``-I`` is case-sensitive based on how it is listed in ``dictionary.txt``. ::
```
peternewman, line 44: Personally I'd be tempted to say if we're rewording these why don't we drop the made up word entirely, it's not a variable so we can use a phrase that's more easily understood by translation and non-native speakers etc.

Perhaps something like:
```suggestion
The ``-I`` flag can be used for a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said list of allowed words. **Important note:** The list passed to ``-I`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
```

README.rst
```diff
@@ -41,11 +41,11 @@ The ``-w`` flag will actually implement the changes recommended by codespell. No
...
 
     codespell -L word1,word2,word3,word4
 
-The ``-L`` flag can be used to whitelist certain words that are comma-separated placed immediately after it. ::
+The ``-L`` flag can be used to allowlist certain words that are comma-separated placed immediately after it. ::
```
peternewman, line 48: Similarly:
```suggestion
The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it. ::
```

codespell_lib/tests/test_dictionary.py
hadess: Removing the first one would trigger:
```
E                       AssertionError: error blaclist: correction blacklist from dictionary dictionary.txt is an error itself in dictionary dictionary_usage.txt
E                       assert 'blacklist' not in {'blacklist': ['blocklist'], 'blacklists': ['blocklists'], 'master': ['primary', 'leader', 'active', 'writer', 'coordinator', 'parent'], 'masters': ['primaries', 'leaders', 'actives', 'writers', 'coordinators', 'parents'], ...}
```

Removing the latter would trigger:
```
E                       AssertionError: error mater: correction master from dictionary dictionary_rare.txt is an error itself in dictionary dictionary_usage.txt
E                       assert 'master' not in {'blacklist': ['blocklist'], 'blacklists': ['blocklists'], 'master': ['primary', 'leader', 'active', 'writer', 'coordinator', 'parent'], 'masters': ['primaries', 'leaders', 'actives', 'writers', 'coordinators', 'parents'], ...}
```

codespell_lib/_codespell.py
hadess: `True, None` works. It didn't like `allowlist` and `blocklist` without it.

peternewman: Ah great thanks.

codespell_lib/tests/test_dictionary.py
peternewman: Makes sense, I'd forgotten about mater.

codespell_lib/_codespell.py
```diff
@@ -43,6 +43,7 @@
     ('clear', 'for unambiguous errors', '', False, None),
     ('rare', 'for rare but valid words', '_rare', None, None),
     ('informal', 'for informal words', '_informal', True, True),
+    ('usage', 'for recommended terms', '_usage', True, None),
```
larsoner, line 46:
```suggestion
    ('usage', 'for recommended terms', '_usage', None, None),
```
Should fix the CI error

larsoner:
Thanks @hadess !

larsoner:
Next step I think would be to make sure we enable this dictionary in testing our own repo at least. Probably will require adding a `-L master` somewhere, other than that I think we should be able to make the other changes easily

codespell_lib/_codespell.py
peternewman: Oops sorry, my duff suggestion.

peternewman:
> Next step I think would be to make sure we enable this dictionary in testing our own repo at least. Probably will require adding a `-L master` somewhere, other than that I think we should be able to make the other changes easily

I've made a start on this here https://github.com/codespell-project/actions-codespell/pull/14 just need to fix some of the tests.

codespell_lib/_codespell.py
hadess: The check passed locally with your suggestion @peternewman, which is why I applied it. Not sure what the difference is between the setups.

hadess:
> Thanks @hadess !

Thanks for being a receptive project :)

codespell_lib/_codespell.py
peternewman: > Not sure what the difference is between the setups.

Can we see your test output. it's probably down to the Aspell stuff. Try running it with `export REQUIRE_ASPELL=true` first.

peternewman:
> Probably will require adding a `-L master` somewhere

Not for much longer: https://www.bbc.co.uk/news/technology-53050955

codespell_lib/_codespell.py
hadess: If I go with `    ('usage', 'for recommended terms', '_usage', True, None),` for the dictionary definition, and remove `man-in-the-middle` from the dictionary (it's not in aspell), the test passes cleanly.

My guess is that my aspell dictionaries (`aspell-en-2019.10.06-2.fc32.x86_64` on Fedora 32) are newer and contain a word that's not in the aspell dictionary that's in the CI.

peternewman: > If I go with ` ('usage', 'for recommended terms', '_usage', True, None),` for the dictionary definition, and remove `man-in-the-middle` from the dictionary (it's not in aspell), the test passes cleanly.
> 
> My guess is that my aspell dictionaries (`aspell-en-2019.10.06-2.fc32.x86_64` on Fedora 32) are newer and contain a word that's not in the aspell dictionary that's in the CI.

Yeah it was complaining about these:
```
E           error 'whitelist' should be in aspell for dictionary /home/travis/build/codespell-project/codespell/codespell_lib/tests/../data/dictionary_usage.txt
E           error 'whitelists' should be in aspell for dictionary /home/travis/build/codespell-project/codespell/codespell_lib/tests/../data/dictionary_usage.txt
```

Which is in the aspell dictionary:
http://app.aspell.net/lookup?dict=en_US&words=whitelists

Our CI is using:
https://packages.ubuntu.com/bionic/aspell-en
Which is 2017.08.24-0-0.1

So yeah you're just a bit newer!

Issue #1469, opened 2020-04-05
Consider splitting the rare dictionary into two


peternewman:
Aren't these words that are actually valid words, but the typo is more likely than the real use? For example `bellow->below`, where bellow is a valid word, but below is much more common so it's more likely to be the typo.

I wonder if it's worth having effectively two rare dictionaries, a rare one and a common one, possibly based on hits on GitHub or different size aspell word lists.

_Originally posted by @peternewman in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3Mzg3NjQ4Mjg2/pull_request_reviews/more_threads_

larsoner:
You have described what to me is already the purpose of the rare dictionary, namely to have corrections whose "error" is actually a valid (but rare) word but in most cases is probably meant to be something else.

So I'm still not sure what the split would be...

peternewman:
Take for example our `nickle->nickel` case, unless you're talking about a specific woodpecker (ignoring the programming language as I'd never heard of it), you've most likely made a typo, so we should probably enable that as a typo by default (but have it in a separate dictionary so it can be filtered out and so we can do the aspell thing).

Whereas `sting->string` or `wit->with` has a much higher chance of yielding a false positive, so we probably don't want them enabled by default.

I guess things that are "technically" valid words, versus things that you are likely to use. I suspect realistically if we think like xkcd ( https://xkcd.com/1133/ ) and filter based on some deemed level of popularity. Either which aspell dictionary it's in, or the stats they have here:
http://app.aspell.net/lookup?dict=en_US&words=pneumonic

larsoner:
Sounds like you're talking splitting based on how rare the "error" is and how common the corrections are (and likelihood of such typos), i.e., a balance of false alarms and misses. This sounds reasonable but we'd want to think of an API and a way to standardize it.

For example we could just adopt the "star" system of `aspell`. I don't think it would be too difficult to allow per-dict star levels, then also modifications of individual entries within those (where necessary). For example we could allow the first line of the dict to be a number of stars, and allow each entry to end (optionally) with a number of stars, then add some conditionals within our code to deal with it.

Not sure if it's worth going to this level of detail or not, though. Maybe just having the `_rare` dictionary (something like `***` equivalent) is already enough for people.

PR #1557, opened 2020-06-10
Move some valid words I stripped out of the dictionary ages ago into rare

peternewman:
I don't know if these are rare enough (hence my discussion in #1469).


peternewman:
Bump @larsoner

PR #1558, opened 2020-06-11
BUG: Use return codes properly, add --count

### LINKED ISSUE #1495: buggy exit status
The `codespell` command currently returns the number of spelling errors, but as an exit status, this has unspecified effects for large values (larger than 125). For instance, 256 will give a 0 exit status. While the number of spelling errors may be fine for the library, it has little interest for the command. The exit status should be either a fixed value (different from those used for other kinds of errors, such as "unrecognized arguments") or limited to some bound (125 or lower).

larsoner:
Use standard return codes:

1. Return ~~2~~ `64` instead of `1` for bad usage
2. Return ~~1~~ `65` when there are spelling errors
3. Add `--count` to print to the last line of `stderr` the number of errors (like `flake8`); should allow easy-ish fixes for people who actually want the count

Adds a wrapper to our tests that goes and grabs the `capsys` output from one stack frame higher so that a ton of lines didn't have to change.

Closes #1495.

@vinc17fr @craigbarnes @kkirsche WDYT?

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -51,9 +51,11 @@ _builtin_dictionaries = (
 )
 _builtin_default = 'clear,rare'
 
-# docs say os.EX_USAGE is only available on Unix systems, so to be safe we
-# protect and just use the value it is on macOS and Linux (64)
+# docs say os.EX_USAGE et al. are only available on Unix systems, so to be safe
+# we protect and just use the values they are on macOS and Linux
+EX_OK = 0
 EX_USAGE = 64
+EX_DATAERR = 65
 
 # OPTIONS:
 #
@@ -245,11 +247,11 @@ def parse_options(args):
 
     parser.add_argument('-d', '--disable-colors',
                         action='store_false', dest='colors',
-                        help='disable colors, even when printing to terminal '
+                        help='Disable colors, even when printing to terminal '
                              '(always set for Windows)')
     parser.add_argument('-c', '--enable-colors',
                         action='store_true', dest='colors',
-                        help='enable colors, even when not printing to '
+                        help='Enable colors, even when not printing to '
                              'terminal')
 
     parser.add_argument('-w', '--write-changes',
@@ -291,11 +293,11 @@ def parse_options(args):
                              'specified together with --write-changes.')
     parser.add_argument('-s', '--summary',
                         action='store_true', default=False,
-                        help='print summary of fixes')
+                        help='Print summary of fixes')
 
     parser.add_argument('--count',
                         action='store_true', default=False,
-                        help='print the number of errors as the last line of '
+                        help='Print the number of errors as the last line of '
                              'stderr')
 
     parser.add_argument('-S', '--skip',
@@ -340,21 +342,21 @@ def parse_options(args):
 
     parser.add_argument('-f', '--check-filenames',
                         action='store_true', default=False,
-                        help='check file names as well')
+                        help='Check file names as well')
 
     parser.add_argument('-H', '--check-hidden',
                         action='store_true', default=False,
                         help='Check hidden files and directories (those '
                              'starting with ".") as well.')
     parser.add_argument('-A', '--after-context', type=int, metavar='LINES',
-                        help='print LINES of trailing context')
+                        help='Print LINES of trailing context')
     parser.add_argument('-B', '--before-context', type=int, metavar='LINES',
-                        help='print LINES of leading context')
+                        help='Print LINES of leading context')
     parser.add_argument('-C', '--context', type=int, metavar='LINES',
-                        help='print LINES of surrounding context')
+                        help='Print LINES of surrounding context')
 
     parser.add_argument('files', nargs='*',
-                        help='files or directories to check')
+                        help='Files or directories to check')
 
     options = parser.parse_args(list(args))
 
@@ -783,4 +785,4 @@ def main(*args):
         print(summary)
     if options.count:
         print(bad_count, file=sys.stderr)
-    return int(bool(bad_count))
+    return EX_DATAERR if bad_count else EX_OK
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -51,6 +51,10 @@ _builtin_dictionaries = (
 )
 _builtin_default = 'clear,rare'
 
+# docs say os.EX_USAGE is only available on Unix systems, and it is 64 on
+# macOS, but things like bash and flake8 give 2, so let's just KISS and use:
+EX_USAGE = 2
+
 # OPTIONS:
 #
 # ARGUMENTS:
@@ -647,7 +651,7 @@ def main(*args):
         print("ERROR: --write-changes cannot be used together with "
               "--regex")
         parser.print_help()
-        return 2
+        return EX_USAGE
     word_regex = options.regex or word_regex_def
     try:
         word_regex = re.compile(word_regex)
@@ -655,7 +659,7 @@ def main(*args):
         print("ERROR: invalid regular expression \"%s\" (%s)" %
               (word_regex, err), file=sys.stderr)
         parser.print_help()
-        return 2
+        return EX_USAGE
 
     ignore_words_files = options.ignore_words or []
     ignore_words = set()
@@ -664,7 +668,7 @@ def main(*args):
             print("ERROR: cannot find ignore-words file: %s" %
                   ignore_words_file, file=sys.stderr)
             parser.print_help()
-            return 2
+            return EX_USAGE
         build_ignore_words(ignore_words_file, ignore_words)
 
     ignore_words_list = options.ignore_words_list or []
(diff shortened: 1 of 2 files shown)


kkirsche:
LGTM :shipit:

codespell_lib/_codespell.py
```diff
@@ -612,15 +617,15 @@ def main(*args):
         print("ERROR: --write-changes cannot be used together with "
               "--regex")
         parser.print_help()
-        return 1
+        return 2
```
peternewman, line 620: https://docs.python.org/2/library/os.html#os.EX_USAGE

codespell_lib/_codespell.py
```diff
@@ -612,15 +617,15 @@ def main(*args):
...
     word_regex = options.regex or word_regex_def
     try:
         word_regex = re.compile(word_regex)
     except re.error as err:
         print("ERROR: invalid regular expression \"%s\" (%s)" %
               (word_regex, err), file=sys.stderr)
         parser.print_help()
-        return 1
+        return 2
```
peternewman, line 628: Again

codespell_lib/_codespell.py
```diff
@@ -629,7 +634,7 @@ def main(*args):
             print("ERROR: cannot find ignore-words file: %s" %
                   ignore_words_file, file=sys.stderr)
             parser.print_help()
-            return 1
+            return 2
```
peternewman, line 637: Again

codespell_lib/_codespell.py
```diff
@@ -742,4 +747,6 @@ def main(*args):
     if summary:
         print("\n-------8<-------\nSUMMARY:")
         print(summary)
-    return bad_count
+    if options.count:
+        print(bad_count, file=sys.stderr)
```
peternewman, line 751: Why don't we have some context to this anyway, people can still grep it out.

peternewman: Edit, I see this is exactly what flake8 does, seems kind of strange to me but it's good to be consistent with other stuff...

codespell_lib/_codespell.py
craigbarnes: The following seems like it would be the simplest way to use the `sysexits.h` codes:

```python
return 65 # os.EX_USAGE
```

For some reason Python seems to only make these constants available when the actual `sysexits.h` header exists on the system ("Availability: Unix"), but they're exactly the same on every platform I'm aware of (NetBSD, FreeBSD, OpenBSD, glibc, musl). The original BSD man page explicitly specifies the values.

larsoner: I went with `2` because it's what `codespell --foo` gave me (same as `bash --foo`). `EX_USAGE` says it's [available on UNIX](https://docs.python.org/3/library/os.html#os.EX_USAGE), which I assume means it's not on Windows. So I think `2` might be the best option here

craigbarnes: Meh, I like `sysexits.h` codes, but I guess as long as the wrapping bug (`256`) is fixed, it doesn't matter all that much.

larsoner: Looks like `EX_USAGE==64` on my macOS machine actually (!). Even `brew --foo` gives `1` so who knows... I'm not sure we should allow it to be system-dependent, given that `bash` and `flake8` don't do it. I'd err on the side of Python-ecosystem consistency here.

craigbarnes: The point of semantic exit codes is that you define ahead of time what they mean (preferably documented) and then use them accordingly. Most other programs aren't really "consistent" -- they generally just return whatever they feel like, either chosen at random or some basic scheme of their own invention. Looking to programs like that for guidance is just propagating the same lack of direction, but alas, it doesn't really matter if the wrapping bug is at least fixed.

codespell_lib/_codespell.py
```diff
@@ -50,6 +50,10 @@
 )
 _builtin_default = 'clear,rare'
 
+# docs say os.EX_USAGE is only available on Unix systems, and it is 64 on
+# macOS, but things like bash and flake8 give 2, so let's just KISS and use:
+EX_USAGE = 2
+
```
craigbarnes, line 56: If you're going to call it `EX_USAGE`, why not use the same value (`64`) as defined by BSD 30 years ago?

larsoner: Agreed, I errantly thought it was `2` on Unix and `64` on macOS -- I've now changed it to use `EX_USAGE`/`64`

craigbarnes: I'd just do `EX_USAGE = 64` and forget about `os.EX_USAGE`. Python made a big mistake in reading this from the header instead of just hard-coding it. The way they define it in the code and the documentation naturally leads people to believe the value varies by platform, but it doesn't -- it's either `64` or the header doesn't exist. Having the the value be `64` everywhere is the whole point of semantic exit codes. Then you can just say "our exit codes are documented in `man sysexits`".

codespell_lib/_codespell.py
```diff
@@ -747,4 +756,6 @@ def main(*args):
     if summary:
         print("\n-------8<-------\nSUMMARY:")
         print(summary)
-    return bad_count
+    if options.count:
+        print(bad_count, file=sys.stderr)
+    return int(bool(bad_count))
```
craigbarnes, line 761: This can be:

```python
if bad_count >= 1:
    return EX_DATAERR
else:
    return EX_OK
```

codespell_lib/_codespell.py
```diff
@@ -50,6 +50,10 @@
 )
 _builtin_default = 'clear,rare'
 
+# docs say os.EX_USAGE is only available on Unix systems, so to be safe we
+# protect and just use the value it is on macOS and Linux (64)
+EX_USAGE = 64
```
craigbarnes, line 55: Also:

```python
EX_OK = 0
EX_DATAERR = 65
```

codespell_lib/_codespell.py
peternewman: I quite like the import option, we do similar in something else I work on:
https://github.com/OpenLightingProject/ola/blob/master/common/base/SysExits.cpp

codespell_lib/_codespell.py
peternewman: Would it be safer and clearer to use something that doesn't clash with the os definition, although it's a different namespace, e.g. EXIT_USAGE?

codespell_lib/tests/test_basic.py
```diff
@@ -53,22 +83,21 @@ def test_basic(tmpdir, capsys):
...
     os.remove(fname)
 
     with open(op.join(d, 'bad.txt'), 'w') as f:
         f.write('abandonned\nAbandonned\nABANDONNED\nAbAnDoNnEd')
     assert cs.main(d) == 4
-    capsys.readouterr()
-    assert cs.main('-w', d) == 0
-    assert 'FIXED:' in capsys.readouterr()[1]
+    code, _, stderr = cs.main('-w', d, std=True)
+    assert code == 0
```
peternewman, line 99: EX_OK

codespell_lib/_codespell.py
larsoner: I think the comment is clear enough, safety wise we can count on this not to change in `os`, and it's arguably clearer to use the name `EX_USAGE` than anything else since that's what we're trying to use

codespell_lib/tests/test_basic.py
```diff
@@ -32,13 +61,14 @@ def test_command(tmpdir):
...
-    """Test some basic functionality"""
+    """Test some basic functionality."""
     assert cs.main('_does_not_exist_') == 0
     fname = op.join(str(tmpdir), 'tmp')
     with open(fname, 'w') as f:
         pass
-    assert cs.main('-D', 'foo', f.name) == 1, 'missing dictionary'
-    assert 'cannot find dictionary' in capsys.readouterr()[1]
+    code, _, stderr = cs.main('-D', 'foo', f.name, std=True)
+    assert code == EX_USAGE, 'missing dictionary'
```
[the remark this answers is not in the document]
larsoner: it's a description
```
>>> assert 1 == 0, 'foo'
AssertionError: foo
```
in a pytest context it gives you more information like what the two values were

codespell_lib/tests/test_basic.py
larsoner: I don't see value in updating all of these tests, I think using `0` interchangeably with `EX_OK` is fine

kkirsche:
If using semver it should be a major as it’s a breaking API change

codespell_lib/tests/test_basic.py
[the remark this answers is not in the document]
peternewman: Ah gotcha.

peternewman:
Sorry @larsoner , you probably want to revert a lot of https://github.com/codespell-project/codespell/pull/1558/commits/7157f15c27846927809418df05a43ef9b0ecd1e4 . I was looking for consistency and assumed sentence case, but looking at `ls` and `flake8` their help strings are lower case.

codespell_lib/tests/test_basic.py
peternewman: Okay.

codespell_lib/_codespell.py
peternewman: Okay.

larsoner:
Okay to merge now that all `help` entries start with lower case @peternewman ?

peternewman:
Do you want to see if @craigbarnes has further feedback or @vinc17fr has any.

Also as @kkirsche and I both commented, please can we bump the major version in this PR.

larsoner:
> please can we bump the major version in this PR.

Done

craigbarnes:
LGTM.

PR #1559, opened 2020-06-11
ENH: Multiline help

larsoner:
@waldyrious @peternewman can you try and see if it's what you had in mind? This is what I see now:
<details>
<summary>codespell --help</summary>

```
usage: codespell [-h] [--version] [-d] [-c] [-w] [-D DICTIONARY]
                 [--builtin BUILTIN-LIST] [-I FILE] [-L WORDS] [-r REGEX] [-s]
                 [-S SKIP] [-x FILE] [-i INTERACTIVE] [-q QUIET_LEVEL] [-e]
                 [-f] [-H] [-A LINES] [-B LINES] [-C LINES]
                 [files [files ...]]

positional arguments:
  files                 files or directories to check

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  -d, --disable-colors  disable colors, even when printing to terminal (always
                        set for Windows)
  -c, --enable-colors   enable colors, even when not printing to terminal
  -w, --write-changes   write changes in place if possible
  -D DICTIONARY, --dictionary DICTIONARY
                        Custom dictionary file that contains spelling
                        corrections. If this flag is not specified or equals
                        "-" then the default dictionary is used. This option
                        can be specified multiple times.
  --builtin BUILTIN-LIST
                        Comma-separated list of builtin dictionaries to
                        include (when "-D -" or no "-D" is passed). Current
                        options are:'clear' for unambiguous errors
                        - 'rare' for rare but valid words
                        - 'informal' for informal words
                        - 'usage' for recommended terms
                        - 'code' for words common to code and/or mathematics
                        - 'names' for valid proper names that might be typos
                        - 'en-GB_to_en-US' for corrections from en-GB to en-US
                        The default is "--builtin clear,rare".
  -I FILE, --ignore-words FILE
                        File that contains words which will be ignored by
                        codespell. File must contain 1 word per line. Words
                        are case sensitive based on how they are written in
                        the dictionary file
  -L WORDS, --ignore-words-list WORDS
                        Comma separated list of words to be ignored by
                        codespell. Words are case sensitive based on how they
                        are written in the dictionary file
  -r REGEX, --regex REGEX
                        Regular expression which is used to find words. By
                        default any alphanumeric character, the underscore,
                        the hyphen, and the apostrophe is used to build words.
                        This option cannot be specified together with --write-
                        changes.
  -s, --summary         print summary of fixes
  -S SKIP, --skip SKIP  Comma-separated list of files to skip. It accepts
                        globs as well. E.g.: if you want codespell to skip
                        .eps and .txt files, you'd give "*.eps,*.txt" to this
                        option.
  -x FILE, --exclude-file FILE
                        FILE with lines that should not be checked for errors
                        or changed
  -i INTERACTIVE, --interactive INTERACTIVE
                        Set interactive mode when writing changes:
                        - 0: no interactivity.
                        - 1: ask for confirmation.
                        - 2: ask user to choose one fix when more than one is
                          available.
                        - 3: both 1 and 2
  -q QUIET_LEVEL, --quiet-level QUIET_LEVEL
                        Bitmask that allows suppressing messages:
                        - 0: print all messages.
                        - 1: disable warnings about wrong encoding.
                        - 2: disable warnings about binary files.
                        - 4: omit warnings about automatic fixes that were
                          disabled in the dictionary.
                        - 8: don't print anything for non-automatic fixes.
                        - 16: don't print the list of fixed files.
                        As usual with bitmasks, these levels can be combined;
                        e.g. use 3 for levels 1+2, 7 for 1+2+4, 23 for
                        1+2+4+16, etc. The default mask is 2.
  -e, --hard-encoding-detection
                        Use chardet to detect the encoding of each file. This
                        can slow down codespell, but is more reliable in
                        detecting encodings other than utf-8, iso8859-1, and
                        ascii.
  -f, --check-filenames
                        check file names as well
  -H, --check-hidden    Check hidden files and directories (those starting
                        with ".") as well.
  -A LINES, --after-context LINES
                        print LINES of trailing context
  -B LINES, --before-context LINES
                        print LINES of leading context
  -C LINES, --context LINES
                        print LINES of surrounding context

```

</details>

Also now allows doing `python -m codespell_lib ...`, which is nice for debugging things like this because you can do `python -im codespell_lib`, then `import pdb; pdb.pm()` on errors.

### DIFF
--- a/codespell_lib/__main__.py
+++ b/codespell_lib/__main__.py
@@ -1,0 +1,4 @@
+from ._codespell import _script_main
+
+if __name__ == '__main__':
+    _script_main()
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -25,6 +25,7 @@ import fnmatch
 import os
 import re
 import sys
+import textwrap
 
 word_regex_def = u"[\\w\\-'’`]+"
 encodings = ('utf-8', 'iso-8859-1')
@@ -206,8 +207,31 @@ class FileOpener(object):
 # -.-:-.-:-.-:-.:-.-:-.-:-.-:-.-:-.:-.-:-.-:-.-:-.-:-.:-.-:-
 
 
+class NewlineHelpFormatter(argparse.HelpFormatter):
+    """Help formatter that preserves newlines and deals with lists."""
+
+    def _split_lines(self, text, width):
+        parts = text.split('\n')
+        out = list()
+        for pi, part in enumerate(parts):
+            # Eventually we could allow others...
+            indent_start = '- '
+            if part.startswith(indent_start):
+                offset = len(indent_start)
+            else:
+                offset = 0
+            part = part[offset:]
+            part = self._whitespace_matcher.sub(' ', part).strip()
+            parts = textwrap.wrap(part, width - offset)
+            parts = [' ' * offset + p for p in parts]
+            if offset:
+                parts[0] = indent_start + parts[0][offset:]
+            out.extend(parts)
+        return out
+
+
 def parse_options(args):
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(formatter_class=NewlineHelpFormatter)
 
     parser.set_defaults(colors=sys.stdout.isatty())
     parser.add_argument('--version', action='version', version=VERSION)
@@ -231,14 +255,14 @@ def parse_options(args):
                              'corrections. If this flag is not specified or '
                              'equals "-" then the default dictionary is used. '
                              'This option can be specified multiple times.')
-    builtin_opts = ', '.join(
-        '%r %s' % (d[0], d[1]) for d in _builtin_dictionaries)
+    builtin_opts = '\n- '.join([] + [
+        '%r %s' % (d[0], d[1]) for d in _builtin_dictionaries])
     parser.add_argument('--builtin',
                         dest='builtin', default=_builtin_default,
                         metavar='BUILTIN-LIST',
                         help='Comma-separated list of builtin dictionaries '
                         'to include (when "-D -" or no "-D" is passed). '
-                        'Current options are:\n%s. The default is '
+                        'Current options are:%s\nThe default is '
                         '"--builtin %s".'
                         % (builtin_opts, _builtin_default))
     parser.add_argument('-I', '--ignore-words',
@@ -276,22 +300,21 @@ def parse_options(args):
 
     parser.add_argument('-i', '--interactive',
                         action='store', type=int, default=0,
-                        help='Set interactive mode when writing changes. '
-                             '0: no interactivity. 1: ask for confirmation. '
-                             '2 ask user to choose one fix when more than one '
-                             'is available. 3: both 1 and 2')
+                        help='Set interactive mode when writing changes:\n'
+                             '- 0: no interactivity.\n'
+                             '- 1: ask for confirmation.\n'
+                             '- 2: ask user to choose one fix when more than one is available.\n'  # noqa: E501
+                             '- 3: both 1 and 2')
 
     parser.add_argument('-q', '--quiet-level',
                         action='store', type=int, default=2,
-                        help='Bitmask that allows suppressing messages. '
-                             '0: print all messages. '
-                             '1: disable warnings about wrong encoding. '
-                             '2: disable warnings about binary files. '
-                             '4: omit warnings about automatic fixes that '
-                             'were disabled in the dictionary. '
-                             '8: don\'t print anything for non-automatic '
-                             'fixes. '
-                             '16: don\'t print the list of fixed files. '
+                        help='Bitmask that allows suppressing messages:\n'
+                             '- 0: print all messages.\n'
+                             '- 1: disable warnings about wrong encoding.\n'
+                             '- 2: disable warnings about binary files.\n'
+                             '- 4: omit warnings about automatic fixes that were disabled in the dictionary.\n'  # noqa: E501
+                             '- 8: don\'t print anything for non-automatic fixes.\n'  # noqa: E501
+                             '- 16: don\'t print the list of fixed files.\n'
                              'As usual with bitmasks, these levels can be '
                              'combined; e.g. use 3 for levels 1+2, 7 for '
                              '1+2+4, 23 for 1+2+4+16, etc. '
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -255,16 +255,15 @@ def parse_options(args):
                              'corrections. If this flag is not specified or '
                              'equals "-" then the default dictionary is used. '
                              'This option can be specified multiple times.')
-    builtin_opts = '\n- '.join([] + [
+    builtin_opts = '\n- '.join([''] + [
         '%r %s' % (d[0], d[1]) for d in _builtin_dictionaries])
     parser.add_argument('--builtin',
                         dest='builtin', default=_builtin_default,
                         metavar='BUILTIN-LIST',
                         help='Comma-separated list of builtin dictionaries '
                         'to include (when "-D -" or no "-D" is passed). '
-                        'Current options are:%s\nThe default is '
-                        '"--builtin %s".'
-                        % (builtin_opts, _builtin_default))
+                        'Current options are:' + builtin_opts + '\n'
+                        'The default is %(default)r.')
     parser.add_argument('-I', '--ignore-words',
                         action='append', metavar='FILE',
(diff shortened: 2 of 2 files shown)


codespell_lib/__main__.py
```diff
@@ -0,0 +1,4 @@
+from ._codespell import _script_main
+
+if __name__ == '__main__':
```
peternewman, line 3: You can also add the __name__ stuff in _codespell.py (which is what I'd been doing just now), but perhaps that has path issues which this doesn't.

codespell_lib/_codespell.py
```diff
@@ -231,14 +255,14 @@ def parse_options(args):
                              'corrections. If this flag is not specified or '
                              'equals "-" then the default dictionary is used. '
                              'This option can be specified multiple times.')
-    builtin_opts = ', '.join(
-        '%r %s' % (d[0], d[1]) for d in _builtin_dictionaries)
+    builtin_opts = '\n- '.join([] + [
```
peternewman, line 258: Why do we need the empty array too?

codespell_lib/_codespell.py
```diff
@@ -231,14 +255,14 @@ def parse_options(args):
...
+        '%r %s' % (d[0], d[1]) for d in _builtin_dictionaries])
     parser.add_argument('--builtin',
                         dest='builtin', default=_builtin_default,
                         metavar='BUILTIN-LIST',
                         help='Comma-separated list of builtin dictionaries '
                         'to include (when "-D -" or no "-D" is passed). '
-                        'Current options are:\n%s. The default is '
+                        'Current options are:%s\nThe default is '
                         '"--builtin %s".'
```
peternewman, line 266: The default should use the format variable thing instead, rather than our own string replacement `%(default)s`.

codespell_lib/_codespell.py
```diff
@@ -231,14 +255,14 @@ def parse_options(args):
...
-        '%r %s' % (d[0], d[1]) for d in _builtin_dictionaries)
+    builtin_opts = '\n- '.join([] + [
+        '%r %s' % (d[0], d[1]) for d in _builtin_dictionaries])
     parser.add_argument('--builtin',
                         dest='builtin', default=_builtin_default,
                         metavar='BUILTIN-LIST',
                         help='Comma-separated list of builtin dictionaries '
                         'to include (when "-D -" or no "-D" is passed). '
-                        'Current options are:\n%s. The default is '
+                        'Current options are:%s\nThe default is '
```
peternewman, line 265: Needs another `\n` before too, from your example output:
```
                        options are:'clear' for unambiguous errors
                        - 'rare' for rare but valid words
```

codespell_lib/__main__.py
larsoner: I don't see a need to do it in this PR (seems like a separate issue) or really have it in the codebase now that we can just do `python -m codespell_lib ...`

codespell_lib/_codespell.py
larsoner: To add the prefix for the first line (not having it elsewhere is causing the bug you point out below), otherwise you end up with N-1 `\n- `'s and N entries

larsoner:
> Would feeding into one of the raw ones not work and be a bit more compliant?

There is indeed some risk that the current implementation could break because we are using a private method (`_split_lines`) to do our work, but hopefully it'll be easy to spot and fix if it happens in the future. We could use `RawTextHelpFormatter`, but it does not line wrap so it does not look nice. We could use `RawTextHelpFormatter` and wrap our own lines at some specified width, but this does not look as nice on small terminals (wraps too late) or use space as well as automatic wrapping on larger terminals (wraps too early).

codespell_lib/__main__.py
peternewman: Yeah I suspect the module is nicer than loading some random file. It was more just a comment it's possible instead of needing the __main__, your options is probably better though.

codespell_lib/_codespell.py
peternewman: Ah yes, makes sense, and probably neater than hardcoding it into the string elsewhere.

peternewman:
> > Would feeding into one of the raw ones not work and be a bit more compliant?
> 
> There is indeed some risk that the current implementation could break because we are using a private method (`_split_lines`) to do our work, but hopefully it'll be easy to spot and fix if it happens in the future. We could use `RawTextHelpFormatter`, but it does not line wrap so it does not look nice. We could use `RawTextHelpFormatter` and wrap our own lines at some specified width, but this does not look as nice on small terminals (wraps too late) or use space as well as automatic wrapping on larger terminals (wraps too early).

Yeah it did seem odd there isn't a better middle ground, I guess it's rather niche they've not made it feature complete enough yet.

larsoner:
Output:

<details>

```
$ codespell --help
usage: codespell [-h] [--version] [-d] [-c] [-w] [-D DICTIONARY] [--builtin BUILTIN-LIST] [-I FILE] [-L WORDS] [-r REGEX] [-s] [-S SKIP] [-x FILE]
                 [-i INTERACTIVE] [-q QUIET_LEVEL] [-e] [-f] [-H] [-A LINES] [-B LINES] [-C LINES]
                 [files [files ...]]

positional arguments:
  files                 files or directories to check

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  -d, --disable-colors  disable colors, even when printing to terminal (always set for Windows)
  -c, --enable-colors   enable colors, even when not printing to terminal
  -w, --write-changes   write changes in place if possible
  -D DICTIONARY, --dictionary DICTIONARY
                        Custom dictionary file that contains spelling corrections. If this flag is not specified or equals "-" then the default dictionary is
                        used. This option can be specified multiple times.
  --builtin BUILTIN-LIST
                        Comma-separated list of builtin dictionaries to include (when "-D -" or no "-D" is passed). Current options are:
                        - 'rare' for rare but valid words
                        - 'informal' for informal words
                        - 'usage' for recommended terms
                        - 'code' for words common to code and/or mathematics
                        - 'names' for valid proper names that might be typos
                        - 'en-GB_to_en-US' for corrections from en-GB to en-US
                        The default is 'clear,rare'.
  -I FILE, --ignore-words FILE
                        File that contains words which will be ignored by codespell. File must contain 1 word per line. Words are case sensitive based on how
                        they are written in the dictionary file
  -L WORDS, --ignore-words-list WORDS
                        Comma separated list of words to be ignored by codespell. Words are case sensitive based on how they are written in the dictionary
                        file
  -r REGEX, --regex REGEX
                        Regular expression which is used to find words. By default any alphanumeric character, the underscore, the hyphen, and the apostrophe
                        is used to build words. This option cannot be specified together with --write-changes.
  -s, --summary         print summary of fixes
  -S SKIP, --skip SKIP  Comma-separated list of files to skip. It accepts globs as well. E.g.: if you want codespell to skip .eps and .txt files, you'd give
                        "*.eps,*.txt" to this option.
  -x FILE, --exclude-file FILE
                        FILE with lines that should not be checked for errors or changed
  -i INTERACTIVE, --interactive INTERACTIVE
                        Set interactive mode when writing changes:
                        - 0: no interactivity.
                        - 1: ask for confirmation.
                        - 2: ask user to choose one fix when more than one is available.
                        - 3: both 1 and 2
  -q QUIET_LEVEL, --quiet-level QUIET_LEVEL
                        Bitmask that allows suppressing messages:
                        - 0: print all messages.
                        - 1: disable warnings about wrong encoding.
                        - 2: disable warnings about binary files.
                        - 4: omit warnings about automatic fixes that were disabled in the dictionary.
                        - 8: don't print anything for non-automatic fixes.
                        - 16: don't print the list of fixed files.
                        As usual with bitmasks, these levels can be combined; e.g. use 3 for levels 1+2, 7 for 1+2+4, 23 for 1+2+4+16, etc. The default mask
                        is 2.
  -e, --hard-encoding-detection
                        Use chardet to detect the encoding of each file. This can slow down codespell, but is more reliable in detecting encodings other than
                        utf-8, iso8859-1, and ascii.
  -f, --check-filenames
                        check file names as well
  -H, --check-hidden    Check hidden files and directories (those starting with ".") as well.
  -A LINES, --after-context LINES
                        print LINES of trailing context
  -B LINES, --before-context LINES
                        print LINES of leading context
  -C LINES, --context LINES
                        print LINES of surrounding context
```

</details>

peternewman:
> Output:

I was a bit worried as your sample didn't seem to line wrap much/at all, but I see it's fine in a checkout. Nice work @larsoner .

larsoner:
The wrap length depends on the terminal size used when doing `--help`

PR #1480, opened 2020-04-10
Expand en_GB->en_US dictionary

waldyrious:
Based on the resources collected at #103.

I made my best effort to make this exhaustive regarding all the pairs listed in that thread and in the pages linked from it, so I suppose this PR closes #103.

PR #1485, opened 2020-04-18
Download a custom Aspell dictionary and build it

peternewman:

PR #1562, opened 2020-06-12
Add more entries to en_GB->en_US dictionary

waldyrious:
Follow-up of #1480; also addresses #103.

The tests are expected to fail for this PR, for the time being; it's being opened mostly to keep track of the words to be added, as collected in #103. See discussion in #1480 for context.


peternewman:
Blocked behind either #1485 or some other option, such as splitting the dictionary.

bl-ue:
Resolved conflicts.  @waldyrious can we merge, or do you have anything against what I've done?

bl-ue (approved):

PR #1577, opened 2020-06-26
Add satement and others

peternewman:


codespell_lib/data/dictionary.txt
```diff
@@ -5995,6 +5998,7 @@ conetxts->contexts, connects,
 conexant->connexant
 conext->context, connect, connects,
 conexts->contexts, connects,
+confectionary->confectionery
```
larsoner, line 6001: This probably needs to go in rare

https://wikidiff.com/confectionery/confectionary

peternewman: Done. You learn something new every day. Does this mean I'm not being anywhere near as thorough as you on my PR reviews, or was it something you already knew?

larsoner: I don't look up every word, my spider-sense was tingling

Issue #676, opened 2018-09-25
Don't report spelling mistakes in HTTP URLs (e.g. in comments)


cfi-gb:
Running codespell 1.13.0 against a file containing e.g. a comment like:

```
# Please see http://example.com/cas/ for more info
```

is showing a spelling mistake for "cas". Codespell could ignore such URLs as (from my experience) most of these are false positives.

rgetz:
Yeah, I see the same with version 1.16.0

```
/* Many of these debug printf include a Flawfinder: ignore, this is because,
 * according to https://cwe.mitre.org/data/definitions/134.html which describes
 * functions that accepts a format string as an argument, but the format
 * ... */
```

produces 

`./debug.h:49: mitre  ==> miter`

which would obviously break the url.

zygoloid:
Similarly this fires for URLs in Markdown sources. For example:

```
- [Type inference and type error diagnosis for Hindley/Milner with extensions (Wazny, J. R., 2006)](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.110.5050&rep=rep1&type=pdf)
```

... reports that "ist" is a spelling error. It seems like it should be straightforward to ignore URLs with common protocols (eg, skip from `http[s]://` to the next likely word break); would a patch to do that be accepted?

peternewman:
Thanks to @jonmeow for his work on this in #1592 .

Initially I personally thought this was really neat and a nice way to fix it, but then I pondered a bit more, it only took a few tries with our list of Microsoft typos to find some on grep.app that were genuine typos (as opposed to all the fishing attempts), for example this one:
https://grep.app/search?q=Micrsoft.com

Wouldn't it be better if they were caught in your documentation?

I wonder if the better solution is a URI/email specific skip list, so you can tell it to ignore ist or cas in those places, but still find the other places where you meant its or case.

>  * according to https://cwe.mitre.org/data/definitions/134.html which describes
> 
> `./debug.h:49: mitre ==> miter`

This specific example the typo has been correctly moved to the GB to US dictionary.

rgetz:
> > * according to https://cwe.mitre.org/data/definitions/134.html which describes
> > 
> > `./debug.h:49: mitre ==> miter`
> 
> This specific example the typo has been correctly moved to the GB to US dictionary.

but that's not the fix - "mitre" is the name of a company. it's not an "re" to "er" issue.

https://www.mitre.org/about/corporate-overview

[...]

PR #1592, opened 2020-07-08
Add --ignore-regex

jonmeow:
This was originally for issue #676, but the preference there is for a different approach. We're keeping it because it may still prove useful, e.g., a regex like (?:[a-fA-F0-9]{0,4}:){2,7}[a-fA-F0-9]{0,4})\\]? could be used to ignore spelling mistakes in ipv6 addresses.

Mechanically, this erases the matching text before the word regex is applied.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -28,8 +28,6 @@ import sys
 import textwrap
 
 word_regex_def = u"[\\w\\-'’`]+"
-# Matches common URIs and email addresses, in that order.
-ignore_word_regex_def = r"(?:(?:https?|ftp|smtp):\/\/([\w-]+\.)+\w{2,}(?:/(?:[\w:/?#\[\]@!$&'()*+,;=.~-]*/?)*)?|[\w.%+-]+@[\w.-]+\.[a-z]{2,})"  # noqa: E501
 encodings = ('utf-8', 'iso-8859-1')
 USAGE = """
 \t%prog [OPTIONS] [file1 file2 ... fileN]
@@ -277,9 +275,11 @@ def parse_options(args):
                         'The default is %(default)r.')
     parser.add_argument('--ignore-regex',
                         action='store', type=str,
-                        help='regular expression which is used to find words '
-                             'to ignore. Matches URIs and emails by default. '
-                             'Can be disabled by setting to "^$".')
+                        help='regular expression which is used to find '
+                             'patterns to ignore by treating as whitespace. '
+                             'When writing regexes, consider ensuring there '
+                             'is boundary non-word chars, e.g., '
+                             '"\\Wmatch\\W". Defaults to empty/disabled.')
     parser.add_argument('-I', '--ignore-words',
                         action='append', metavar='FILE',
                         help='file that contains words which will be ignored '
@@ -497,8 +497,9 @@ def print_context(lines, index, context):
 
 
 def extract_words(text, word_regex, ignore_word_regex):
-    interesting_text = ignore_word_regex.sub(' ', text)
-    return word_regex.findall(interesting_text)
+    if ignore_word_regex:
+        text = ignore_word_regex.sub(' ', text)
+    return word_regex.findall(text)
 
 
 def parse_file(filename, colors, summary, misspellings, exclude_lines,
@@ -674,14 +675,17 @@ def main(*args):
               (word_regex, err), file=sys.stderr)
         parser.print_help()
         return EX_USAGE
-    ignore_word_regex = options.ignore_regex or ignore_word_regex_def
-    try:
-        ignore_word_regex = re.compile(ignore_word_regex)
-    except re.error as err:
-        print("ERROR: invalid regular expression \"%s\" (%s)" %
-              (ignore_word_regex, err), file=sys.stderr)
-        parser.print_help()
-        return EX_USAGE
+
+    if options.ignore_regex:
+        try:
+            ignore_word_regex = re.compile(options.ignore_regex)
+        except re.error as err:
+            print("ERROR: invalid regular expression \"%s\" (%s)" %
+                  (options.ignore_regex, err), file=sys.stderr)
+            parser.print_help()
+            return EX_USAGE
+    else:
+        ignore_word_regex = None
 
     ignore_words_files = options.ignore_words or []
     ignore_words = set()
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -278,7 +278,7 @@ def parse_options(args):
                         help='regular expression which is used to find '
                              'patterns to ignore by treating as whitespace. '
                              'When writing regexes, consider ensuring there '
-                             'is boundary non-word chars, e.g., '
+                             'are boundary non-word chars, e.g., '
                              '"\\Wmatch\\W". Defaults to empty/disabled.')
     parser.add_argument('-I', '--ignore-words',
                         action='append', metavar='FILE',
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -671,7 +671,7 @@ def main(*args):
     try:
         word_regex = re.compile(word_regex)
     except re.error as err:
-        print("ERROR: invalid regular expression \"%s\" (%s)" %
+        print("ERROR: invalid --regex \"%s\" (%s)" %
               (word_regex, err), file=sys.stderr)
         parser.print_help()
         return EX_USAGE
@@ -680,7 +680,7 @@ def main(*args):
         try:
             ignore_word_regex = re.compile(options.ignore_regex)
         except re.error as err:
-            print("ERROR: invalid regular expression \"%s\" (%s)" %
+            print("ERROR: invalid --ignore-regex \"%s\" (%s)" %
                   (options.ignore_regex, err), file=sys.stderr)
             parser.print_help()
             return EX_USAGE
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -28,6 +28,8 @@ import sys
 import textwrap
 
 word_regex_def = u"[\\w\\-'’`]+"
+# Matches common URIs and email addresses, in that order.
+ignore_word_regex_def = r"(?:(?:https?|ftp|smtp):\/\/([\w-]+\.)+\w{2,}(?:/(?:[\w:/?#\[\]@!$&'()*+,;=.~-]*/?)*)?|[\w.%+-]+@[\w.-]+\.[a-z]{2,})"  # noqa: E501
 encodings = ('utf-8', 'iso-8859-1')
 USAGE = """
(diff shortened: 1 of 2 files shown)


codespell_lib/tests/test_basic.py
```diff
@@ -455,6 +455,86 @@ def test_context(tmpdir, capsys):
...
+    assert cs.main(f.name) == 0
+
+    # Typo because there's no protocol.
+    with open(op.join(d, 'uri.txt'), 'w') as f:
+        f.write('# Please see example.com/abandonned for info\n')
+    assert cs.main(f.name) == 1
+
+    # Typo because there aren't enough domain parts.
+    with open(op.join(d, 'uri.txt'), 'w') as f:
+        f.write('# Please see http://abandonned for info\n')
```
peternewman, line 514: This would be fine in some scenarios, for example within a company.

codespell_lib/tests/test_basic.py
```diff
@@ -455,6 +455,86 @@ def test_context(tmpdir, capsys):
...
+    assert cs.main(f.name) == 0
+
+    # Ignoring text in domain.
+    with open(op.join(d, 'email.txt'), 'w') as f:
+        f.write('# Please contact example@abandonned.com for info\n')
+    assert cs.main(f.name) == 0
+
+    # Typo because there's no TLD for an email.
+    with open(op.join(d, 'email.txt'), 'w') as f:
+        f.write('# Please contact abandonned@example for info\n')
```
peternewman, line 534: This in theory would too, although it's less likely.

codespell_lib/_codespell.py
```diff
@@ -28,6 +28,8 @@
 import textwrap
 
 word_regex_def = u"[\\w\\-'’`]+"
+# Matches common URIs and email addresses, in that order.
+ignore_word_regex_def = r"(?:(?:https?|ftp|smtp):\/\/([\w-]+\.)+\w{2,}(?:/(?:[\w:/?#\[\]@!$&'()*+,;=.~-]*/?)*)?|[\w.%+-]+@[\w.-]+\.[a-z]{2,})"  # noqa: E501
```
peternewman, line 32: Perhaps add file/git/smb (and tftp as it's an easy addition on the existing ftp).

SMTP isn't listed here and I don't think I've particularly seen it either.
https://en.wikipedia.org/wiki/List_of_URI_schemes

peternewman: Also you should optionally accept `mailto:` at the start of the email regex.

peternewman: This doesn't look like it will accept an IPv4 address (or probably an IPv6 one), which it should ideally match too. Was it sourced from somewhere or did you write it yourself?

jonmeow: I'll keep this in mind for a possible uri ignore, but it's removed from this PR.

This is something I wrote myself, just trying to write a short/readable matcher, vs something that'd only match canonical HTTP; I don't think even the examples I found were doing IPv6. I looked to RFCs for the path/cgi/anchor char list. If you have a regex you'd prefer, please let me know.

codespell_lib/tests/test_basic.py
jonmeow: True, although it didn't feel like it'd be a common enough issue to handle. Let me know if you'd rather support this style (it was deliberate as an example). But removed here.

codespell_lib/tests/test_basic.py
jonmeow: This I felt was less worth supporting -- "@" could be used as the literal abbreviation for "at", and Java uses it for doc syntax. I haven't seen emails drop the .com like http does.

peternewman: Javadoc only uses at as a prefix, e.g. `@brief`. You wouldn't see external emails without it, as they wouldn't be routed.

codespell_lib/tests/test_basic.py
peternewman: I think a lot of large businesses will have stuff that's not FQDN, also things like open source projects (although this not great example uses pi.hole):
https://docs.pi-hole.net/guides/vpn/only-dns-via-vpn/#alternative-2-set-dns-server-address-of-your-pi-hole-manually-on-the-main-interface

codespell_lib/_codespell.py
peternewman: I've not looked, but I assume there should be some available somewhere which have been well tested and cover a lot of edge cases.

codespell_lib/_codespell.py
```diff
@@ -663,6 +676,17 @@ def main(*args):
         parser.print_help()
         return EX_USAGE
 
+    if options.ignore_regex:
+        try:
+            ignore_word_regex = re.compile(options.ignore_regex)
+        except re.error as err:
+            print("ERROR: invalid regular expression \"%s\" (%s)" %
```
peternewman, line 683: Perhaps add which regex it is given we've already got the word regex and may have more if you add the URI stuff in.

codespell_lib/tests/test_basic.py
```diff
@@ -455,6 +455,25 @@ def test_context(tmpdir, capsys):
...
+    assert 'usage:' in stdout
+
+    with open(op.join(d, 'flag.txt'), 'w') as f:
+        f.write('# Please see http://example.com/abandonned for info\n')
+    # Non-matching regex results in nothing being ignored.
+    assert cs.main(f.name, '--ignore-regex=^$') == 1
+    # A word can be ignored.
+    assert cs.main(f.name, '--ignore-regex=abandonned') == 0
+    # Ignoring part of the word can result in odd behavior.
+    assert cs.main(f.name, '--ignore-regex=nn') == 0
```
peternewman, line 474: Perhaps test with your default no option flag? `--ignore-regex` Or does OptParser catch that?

codespell_lib/tests/test_basic.py
```diff
@@ -455,6 +455,25 @@ def test_context(tmpdir, capsys):
...
+
+    # Invalid regex.
+    code, stdout, _ = cs.main('--ignore-regex=(', std=True)
+    assert code == EX_USAGE
+    assert 'usage:' in stdout
+
+    with open(op.join(d, 'flag.txt'), 'w') as f:
+        f.write('# Please see http://example.com/abandonned for info\n')
+    # Non-matching regex results in nothing being ignored.
+    assert cs.main(f.name, '--ignore-regex=^$') == 1
```
peternewman, line 470: It would be good to confirm you get the error without that option on the test string too, in case we ever change the main word regex.

jonmeow: Done

codespell_lib/tests/test_basic.py
jonmeow: It's handled by argparse. I'm a bit wary of catching SystemExit, which is what it's throwing. Per https://docs.python.org/3/library/argparse.html#exiting-methods, the right way to prevent exit from argparse is to subclass. That seems like it'd take a fair bit of wiring, though; are you comfortable relying on argparse handling rather than tests here?

codespell_lib/_codespell.py
jonmeow: Done (will comment back on the issue to make sure thinking about URI flags matches)

codespell_lib/tests/test_basic.py
peternewman: Resolving from this PR anyway.

codespell_lib/tests/test_basic.py
peternewman: Resolving from this PR anyway.

codespell_lib/_codespell.py
peternewman: Resolving from this PR anyway.

codespell_lib/tests/test_basic.py
peternewman: Are you sure? Can you not do a `--verbose` or `--verbose=5` type argument? If it definitely complains when you don't pass an option, I'm happy to not deal with testing argparse, although it would still be good to test with a blank one, so `--ignore-regex=""` which should map to the empty string I think, hence false.

jonmeow: > Are you sure? Can you not do a `--verbose` or `--verbose=5` type argument?

Yes, and you can actually see this already with other flags like `codespell -D`.

If both with-argument and no-argument forms were supported simultaneously, `codespell --verbose foo.txt` would be an ambiguous parse: the `foo.txt` could either be an argument to `--verbose` (with-argument `--verbose=foo.txt`) or a file argument to `foo` (no-argument `--verbose`).

Behavior like `-vvv` is used to address this use-case, repeating the flag to indicate a different level (in argparse, this is `action="count"` -- https://docs.python.org/3/library/argparse.html#action).

> If it definitely complains when you don't pass an option, I'm happy to not deal with testing argparse, although it would still be good to test with a blank one, so `--ignore-regex=""` which should map to the empty string I think, hence false.

Added.

codespell_lib/_codespell.py
```diff
@@ -273,6 +273,13 @@ def parse_options(args):
                         'to include (when "-D -" or no "-D" is passed). '
                         'Current options are:' + builtin_opts + '\n'
                         'The default is %(default)r.')
+    parser.add_argument('--ignore-regex',
+                        action='store', type=str,
+                        help='regular expression which is used to find '
+                             'patterns to ignore by treating as whitespace. '
+                             'When writing regexes, consider ensuring there '
+                             'is boundary non-word chars, e.g., '
```
peternewman, line 281: Nit sorry, grammar wise it should be "ensuring there are boundary..."

codespell_lib/tests/test_basic.py
```diff
@@ -455,6 +455,30 @@ def test_context(tmpdir, capsys):
...
+        f.write('# Please see http://example.com/abandonned for info\n')
+    # Test file has 1 invalid entry, and it's not ignored by default.
+    assert cs.main(f.name) == 1
+    # An empty regex is the default value, and nothing is ignored.
+    assert cs.main(f.name, '--ignore-regex=') == 1
+    assert cs.main(f.name, '--ignore-regex=""') == 1
+    # Non-matching regex results in nothing being ignored.
+    assert cs.main(f.name, '--ignore-regex=^$') == 1
+    # A word can be ignored.
+    assert cs.main(f.name, '--ignore-regex=abandonned') == 0
```
peternewman, line 477: It would be good to test the real world functionality, so e.g. add a typo (I guess ideally a different one) elsewhere in the example text, then do a check of a regex match of it (e.g. with a character class) and the word boundary markers and confirm the flag actually works in reality.

peternewman:
@larsoner do you want to review this, or are you happy with mine?

codespell_lib/_codespell.py
jonmeow: Done

codespell_lib/tests/test_basic.py
jonmeow: Done

PR #1594, opened 2020-07-15
Add additional spelling corrections for obtain(s|ed)?.

cfi-gb:
See e.g.

https://github.com/search?q=optained&type=Code

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -18380,7 +18380,7 @@ obstacal->obstacle
 obstancles->obstacles
 obstruced->obstructed
 obsure->obscure
-obtaien->obtained, obtain,
+obtaien->obtain, obtained,
 obtaiend->obtained
 obtaiens->obtains
 obtaion->obtain
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -18352,6 +18352,7 @@ objets->objects
 objext->object
 objtain->obtain
 objtained->obtained
+objtains->obtains
 objump->objdump
 oblitque->oblique
 obnject->object
@@ -18372,10 +18373,15 @@ obstacal->obstacle
 obstancles->obstacles
 obstruced->obstructed
 obsure->obscure
+obtaien->obtain
 obtaiend->obtained
+obtaiens->obtains
 obtaion->obtain
+obtaioned->obtained
+obtaions->obtains
 obtrain->obtain
 obtrained->obtained
+obtrains->obtains
 obusing->abusing
 obvisious->obvious
 obvisous->obvious
@@ -18705,6 +18711,8 @@ opreations->operations
 opression->oppression
 opressive->oppressive
 optain->obtain
+optained->obtained
+optains->obtains
 optaionl->optional
 opten->often, open,
 optening->opening
@@ -18879,6 +18887,8 @@ osffsets->offsets
 osffsetting->offsetting
 osicllations->oscillations
 ot->to, of, or,
+otain->obtain
+otained->obtained
 otains->obtains
 otehr->other
 otehrwice->otherwise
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -18372,10 +18373,15 @@ obstacal->obstacle
 obstancles->obstacles
 obstruced->obstructed
 obsure->obscure
+obtaien->obtain
```
peternewman, line 18376: Grep.app doesn't find any hits for this, I think because what it's presumably based on `obtaiend` is a result of swapping two of the end characters, whereas this is about inserting another character and the e is no-where near the i or n.

codespell_lib/data/dictionary.txt
```diff
@@ -18372,10 +18373,15 @@ obstacal->obstacle
...
 obtaiend->obtained
+obtaiens->obtains
```
peternewman, line 18378: Again

codespell_lib/data/dictionary.txt
cfi-gb: There are a few available via e.g. google though.

Typo in a code comment:
https://stackoverflow.com/questions/35522401/updating-jagged-array-without-losing-the-value-in-previous-rows-in-c-sharp

Some sort of academical stuff:
https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM1538107
https://www.researchgate.net/publication/257058426_Prevencion_del_cancer_de_cervix_Comparacion_de_la_calidad_de_la_muestra_segun_utilicemos_bastoncillo_o_cepillo_cervical
https://www.icao.int/SAM/Documents/2015-SAMAIM7/SAMAIM7_WP07_SistAutomat.pdf

Not sure if these hits justify the addition of this typos.

peternewman: > Typo in a code comment:
> https://stackoverflow.com/questions/35522401/updating-jagged-array-without-losing-the-value-in-previous-rows-in-c-sharp

Grammar wise, this one looks like it should probably be obtained.

> Some sort of academical stuff:
> https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM1538107

Agreed
> https://www.researchgate.net/publication/257058426_Prevencion_del_cancer_de_cervix_Comparacion_de_la_calidad_de_la_muestra_segun_utilicemos_bastoncillo_o_cepillo_cervical

Looks like it should perhaps be obtained again
> https://www.icao.int/SAM/Documents/2015-SAMAIM7/SAMAIM7_WP07_SistAutomat.pdf

Agreed

Okay, personally I'm sold, on the proviso you add obtained too as a second correction.

codespell_lib/data/dictionary.txt
peternewman: Agreed there are a handful of examples on Google of this too, although a lot seem to be poor OCR. They do seem to be obtains in the ones I can find.

codespell_lib/data/dictionary.txt
cfi-gb: Thanks for your second review / opinion about that. I have added `obtained` as a second correction in b0392bc9713a89c56f0815ecad5dc1aa67db2a1a

peternewman: I'd have been happy for obtain to be the primary correction, but it's probably about equal.

peternewman:
Travis has got upset @cfi-gb . It never saw your check. A force push might make it retry, or I can override the build if I faff in the settings a bit.

cfi-gb:
> Travis has got upset @cfi-gb . It never saw your check. A force push might make it retry, or I can override the build if I faff in the settings a bit.

I took the change to make obtain the primary correction as suggested in https://github.com/codespell-project/codespell/pull/1594#discussion_r455879448, let's see what Travis is doing.

peternewman:
Yeah Travis is no longer out to lunch!

PR #1598, opened 2020-07-18
Add more misspelled words

sebweb3r:
Two common mistakes of my group

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -9280,6 +9280,7 @@ disignated->designated
 disinguish->distinguish
 disiplined->disciplined
 disired->desired
+diskretization->discretization
 dislaimer->disclaimer
 dislay->display
 dislayed->displayed
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -8150,6 +8150,7 @@ demudulator->demodulator
 denegrating->denigrating
 densitity->density
 densly->densely
+denstiy->density
 deocder->decoder
 deoes->does
 deoesn't->doesn't
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -9280,7 +9280,9 @@ disignated->designated
 disinguish->distinguish
 disiplined->disciplined
 disired->desired
+diskrete->discrete
 diskretization->discretization
+diskretize->discretize
 dislaimer->disclaimer
 dislay->display
 dislayed->displayed
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -9281,9 +9281,11 @@ disinguish->distinguish
 disiplined->disciplined
 disired->desired
 diskrete->discrete
+diskretion->discretion
 diskretization->discretization
 diskretize->discretize
 diskretized->discretized
+diskrimination->discrimination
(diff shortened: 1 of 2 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -9279,6 +9280,7 @@ disignated->designated
 disinguish->distinguish
 disiplined->disciplined
 disired->desired
+diskretization->discretization
```
peternewman, line 9283: Can you add the sation variant as a separate typo please.

Also the Wikipedia page mentions discretized, so it might be worth adding this typo against that word and maybe also discrete, discretely, discretion etc. Possibly also discuss etc (although grep.app doesn't find examples of the latter).

sebweb3r: I added some of your suggestions. I think I will make/lookup a list of commonly misspelled words by germans and add it in my holidays.

sebweb3r: @peternewman i don't understand the travis checks.
f.e. diskretization is for sure not in aspell, since it is wrong. 
or https://en.wiktionary.org/wiki/discretise exists :-)

codespell_lib/data/dictionary_en-GB_to_en-US.txt
```diff
@@ -26,6 +26,12 @@ customising->customizing
 dialogue->dialog
 digitise->digitize
 digitising->digitizing
+discretisation->discretization
+discretise->discretize
+discretised->discretized
+diskretisation->discretization
+diskretise->discretize
+diskretised->discretized
```
peternewman, line 34: These can be removed from here, this dictionary should only be valid British words mapping to valid American ones, e.g. colour->color, Travis checks that both left and right of these dictionary entries are in the dictionary, hence it's complaining that diskretisation isn't a real word (because it isnt').

codespell_lib/data/dictionary_en-GB_to_en-US.txt
```diff
@@ -26,6 +26,12 @@ customising->customizing
 dialogue->dialog
 digitise->digitize
 digitising->digitizing
+discretisation->discretization
+discretise->discretize
+discretised->discretized
```
peternewman, line 31: It seems these particular British words are too obscure for our current Aspell dictionary. #1485 might resolve that issue, but for now you should probably just remove these ones.

codespell_lib/data/dictionary.txt
peternewman: > I added some of your suggestions. I think I will make/lookup a list of commonly misspelled words by germans and add it in my holidays.

Great, the trick there is probably to add a new dictionary called dictionary-de.txt and if necessary move anything in the main dictionary that's a valid German word (like some of my other suggestions) out into a dictionary-en.txt (and make that default). Then ideally make Travis check all that complication too!

> @peternewman i don't understand the travis checks.
> f.e. diskretization is for sure not in aspell, since it is wrong.
> or https://en.wiktionary.org/wiki/discretise exists :-)

See my comments for some further explanation.

codespell_lib/data/dictionary_en-GB_to_en-US.txt
peternewman: I can't find discretise in any variant of the word list: http://app.aspell.net/lookup?dict=en_GB-large&words=discretise

PR #1605, opened 2020-07-28
Add vulnerab(le|ility|ies) corrections.

cfi-gb:
See https://grep.app/search?q=vulernab


cfi-gb:
nvm, already there: #1603

Issue #1597, opened 2020-07-16
Possible additional data source for the usage dictionary


peternewman:
https://twitter.com/TwitterEng/status/1280188545743745024

I don't know if you want to add these @hadess or someone else can at some point.

vikivivi:
Also see word-list from Google developer documentation style guide
https://developers.google.com/style/word-list

Issue #74, opened 2016-06-06
Unused dictionary entries


jwilk:
Some entries in the dictionary will never be used by codespell, because they contain uppercase letters or spaces or punctuation characters:

```
$ LC_ALL=C grep -v -E "^[a-z0-9_'-]+->" codespell_lib/data/dictionary.txt
Amercia->America
Bernouilli->Bernoulli
Blitzkreig->Blitzkrieg
Bonnano->Bonanno
Brasillian->Brazilian
Britian->Britain
Brittish->British
Buddah->Buddha
Buddist->Buddhist
Cambrige->Cambridge
Capetown->Cape Town
Carmalite->Carmelite
Carnagie->Carnegie
Carnagie-Mellon->Carnegie-Mellon
Carnigie->Carnegie
Carnigie-Mellon->Carnegie-Mellon
Carribbean->Caribbean
Carribean->Caribbean
Carthagian->Carthaginian
Cataline->Catiline, Catalina,
Ceasar->Caesar
Celcius->Celsius
Champange->Champagne
Cincinatti->Cincinnati
Cincinnatti->Cincinnati
Coca Cola->Coca-Cola
Conneticut->Connecticut
Dardenelles->Dardanelles
Dravadian->Dravidian
Enlish->English, enlist,
Europian->European
Europians->Europeans
Eurpean->European
Eurpoean->European
Farenheit->Fahrenheit
Febuary->February
Feburary->February
Flemmish->Flemish
Formalhaut->Fomalhaut
Foundland->Newfoundland
Fransiscan->Franciscan
Fransiscans->Franciscans
Galations->Galatians
Gameboy->Game Boy
Ghandi->Gandhi
Godounov->Godunov
Gothenberg->Gothenburg
Gottleib->Gottlieb
Guaduloupe->Guadalupe, Guadeloupe,
Guadulupe->Guadalupe, Guadeloupe,
Guatamala->Guatemala
Guatamalan->Guatemalan
Guilia->Giulia
Guilio->Giulio
Guiness->Guinness
Guiseppe->Giuseppe
Habsbourg->Habsburg
Hallowean->Hallowe'en, Halloween,
Hatian->Haitian
Heidelburg->Heidelberg
Ihaca->Ithaca
Israelies->Israelis
Janurary->January
Januray->January
Japanes->Japanese
Johanine->Johannine
Jospeh->Joseph
Juadaism->Judaism
Juadism->Judaism
Lybia->Libya
Malcom->Malcolm
[...]

PR #1607, opened 2020-07-28
Add renaming advice based on Twitter engineering guidelines

hadess:
As listed in:
http://web.archive.org/web/20200706172154/https://twitter.com/TwitterEng/status/1280188545743745024

Note that “placeholder” was not known in my dictionary so replaced with
“place-holder”, and that pronouns were not added, as adding renaming
advice based on those would not work very well with codespell.

Closes: #1597


peternewman:
> Note that “placeholder” was not known in my dictionary so replaced with
> “place-holder”

What's your dictionary? It's shown as being in the core US one and both GB ones (let alone large). What does the CI say?

hadess:
Would we want to add support for spaces in the typo term? Or is that going to be a problem because the items won't be found in the dictionary?

peternewman:
> Would we want to add support for spaces in the typo term? Or is that going to be a problem because the items won't be found in the dictionary?

Probably yes in the long term (see e.g. #74 ), I think as part of that you'd then fix the dictionary checker to check each word rather than the whole correction.

It would need a bit of work though, as our regex currently finds words, you'd need to then take pairs or words and see if they match a correction, probably while skipping pairs with e.g. a comma or full stop. So I'd suggest let's get the basic one in first, and if you fancy having a go at the complicated case afterwards would be great.

lurch:
#1657 is relevant here :slightly_smiling_face:

hadess:
I've rebased on top of https://github.com/codespell-project/codespell/pull/1657 for testing.

codespell_lib/_codespell.py
```diff
@@ -28,7 +28,8 @@
 import sys
 import textwrap
 
-word_regex_def = u"[\\w\\-'’`]+"
+# NOTE: flake8 suppression due to it not liking \] escape sequence
+word_regex_def = u"([\\w\\-'’`]+)([.,?!-:;><@#$%^&*()_+=/\]\\[])?"  # noqa W605
```
DimitriPapadopoulos, line 32: I believe flake8 is correct, **`\]`** is **not** a valid escape sequence in Python.

PR #1608, opened 2020-08-01
Add correction for 'srting' -> 'string'

waldyrious:


codespell_lib/data/dictionary.txt
```diff
@@ -24991,6 +24991,7 @@ srinked->shrunk
 srinking->shrinking
 srollbar->scrollbar
 srouce->source
+srting->string
```
peternewman, line 24994: Can you also suggest sorting.

Can you add the plural version of this typo too.

waldyrious: Oh right, it didn't occur to me. Will do both changes shortly.

waldyrious: @peternewman done :)

PR #1613, opened 2020-08-06
Add straighted->straightened

peternewman:


peternewman:
Or possibly this should be in rare, needs more research.

peternewman:
Now in rare based on http://poincare.matf.bg.ac.rs/~cvetana/xml-primene/xml-oed/wod270306.htm

PR #1618, opened 2020-08-10
Add exclusuively->exclusively and friends

peternewman:


larsoner:
CIs red, otherwise LGTM

PR #1622, opened 2020-08-11
Add misspellings of semaphore

lurch:


lurch:
Done. I didn't realise plurals weren't handled automatically.

peternewman:
> Done. I didn't realise plurals weren't handled automatically.

Unfortunately it's not necessarily as simple as that, so I'd imagine it's safer to do it manually:
https://en.wikipedia.org/wiki/English_plurals#Irregular_plurals

I guess the system could have a list of the rules and the exceptions, but then that needs testing and can be missed. Also a typo of a plural might have different additional suggestions (aside from the obvious correction) to it's singular.

lurch:
Yeah, I sometimes forget that the English language has so many odd rules and exceptions :wink:

PR #1623, opened 2020-08-11
Unify acknowledgment

sebweb3r:
According to https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Spelling 
acknowledgment is the preferred us-american spelling.


larsoner:
I think thus far we have tended/tried to do GB_wrong->GB_correct and then US_wrong->US_correct rather than GB_wrong->US_correct in dictionary.txt, then GB_correct->US_correct in the US_GB dictionary

sebweb3r:
@peternewman your argument is very convincing. I reduced my commit to the entry of enGB_to_enUS

PR #1628, opened 2020-08-12
Add interfcae -> interface

lurch:


codespell_lib/data/dictionary.txt
```diff
@@ -15519,6 +15519,7 @@ interessted->interested
 interessting->interesting
 intereview->interview
 interfave->interface
+interfcae->interface
```
peternewman, line 15522: Can you add the plural of this please. I know it's not your change but would you mind doing the same for the one above too.

lurch: Done.

PR #1624, opened 2020-08-11
Fix mistakes

sebweb3r:
I had a longer look on the dictionary and found some mistakes.

PR #1643, opened 2020-08-18
Fix misspell in suggestion word

vikivivi:
- https://en.wiktionary.org/wiki/minusculy
- identificable is Spanish


sebweb3r:
most of them are already in #1624

vikivivi:
No longer relevant since merged in #1624

PR #1645, opened 2020-08-18
Add nedle->needle and neeedle->needle

lurch:


codespell_lib/data/dictionary.txt
```diff
@@ -17735,11 +17735,17 @@ neded->needed
 nedia->media
 nedium->medium
 nediums->mediums
+nedle->needle
+nedles->needles
+nedless->needless
```
peternewman, line 17740: I'd probably suggest the opposite typo too for each one.

codespell_lib/data/dictionary.txt
```diff
@@ -17735,11 +17735,17 @@ neded->needed
...
 neede->needed, need,
 needeed->needed
 neeed->need
 neeeded->needed
 neeeding->needing
+neeedle->needle
+neeedles->needles
+neeedless->needless
```
peternewman, line 17748: Probably again

codespell_lib/data/dictionary.txt
lurch: Are you suggesting:
```
nedles->needles, needless,
nedless->needless, needles,
```
?

I'm not sure if that makes sense, given that the dictionary (obviously) doesn't also have
```
needles->needless
needless->needles
```
? :shrug:

peternewman: Yes, because when you've typed `nedles` you've already made a mistake. Yes needles is the closest Levenstein distance, but needless is only a little further away and could be likely. If we suggest both, it will prompt and let the user pick, rather than auto-replacing with a possibly incorrect suggestion.

lurch: Fair enough!

codespell_lib/data/dictionary.txt
lurch: Done

peternewman (approved):

PR #1647, opened 2020-08-19
Fix dictionary check action and standardise the make command we run

peternewman:


peternewman:
This isn't catching the failure I thought it might. This fails:
https://github.com/codespell-project/codespell/runs/999123609?check_suite_focus=true

This doesn't:
https://github.com/codespell-project/codespell/runs/1003544298?check_suite_focus=true

But the equivalent Travis does.

Is it actually that pytest is missing or something?

peternewman:
It is indeed pytest:
https://github.com/codespell-project/codespell/runs/1039801917

.github/workflows/codespell-private.yml
```diff
@@ -7,6 +7,10 @@ jobs:
   make-check-dictionaries:
     runs-on: ubuntu-latest
     steps:
+      - name: Set up Python
+        uses: actions/setup-python@v2
+      - name: Install dependencies
+        run: pip install pytest pytest-cov pytest-dependency
```
peternewman, line 13: Should we make a requirements-test.txt or something so we can load one file for GitHub and Travis?

Makefile
```diff
@@ -27,6 +27,8 @@ check-dictionaries:
 	done
 	@if command -v pytest > /dev/null; then \
 		pytest codespell_lib/tests/test_dictionary.py; \
+	else \
+		echo "pytest command not present, dictionary tests being skipped"; \
```
peternewman, line 31: I dunno if this should exit or just warn? I guess technically it's "some dictionary tests" being skipped, as trim and sort are being run.

lurch: Might be worth adding `1>&2` on the end to get it echoing to STDERR instead of STDOUT?

peternewman: Yeah I could do, although the other errors in the Makefile are going to STDOUT currently. Personally I'd be tempted to drop the @'s so it generates more output, but I guess that might be too noisy for some people's preference. It certainly would have made finding this issue easier.

larsoner: I would probably make it exit with an informative error so that people are pushed toward installing the necessary dependencies

.github/workflows/codespell-private.yml
larsoner: Feel free to add it if you want

peternewman: This seems to be one way to go:
https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

So I've done it

Makefile
peternewman: Done.

README.rst
```diff
@@ -107,6 +107,12 @@ directly, but instead be manually inspected. E.g.:
    Note that there isn't a comma in the end of the line. The last argument is
    treated as the reason why a suggestion cannot be automatically applied.
 
+Development Setup
+-----------------
+
+You can install required dependencies for development by running the following within a checkout of the codespell source:
```
larsoner, line 113:
```suggestion
You can install required dependencies for development by running the following within a checkout of the codespell source::

```

README.rst
```diff
@@ -181,6 +187,7 @@ with the following terms:
 
 .. _GPL v2: http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
 
-dictionary.txt is a derived work of English Wikipedia and is released under the
-Creative Commons Attribution-Share-Alike License 3.0
+dictionary.txt and the other dictionary_*.txt files are a derived work of
```
larsoner, line 190:
```suggestion
dictionary.txt and the other ``dictionary_*.txt`` files are a derived work of
```

README.rst
peternewman: Oops thanks, I'm not really hot on .rst files.

.travis.yml
```diff
@@ -29,16 +29,17 @@ before_install:
     - source venv/bin/activate
     - python --version  # just to check
     - pip install -U pip wheel # upgrade to latest pip find 3.5 wheels; wheel to avoid errors
-    - retry pip install pytest pytest-cov pytest-dependency flake8 coverage codecov chardet "setuptools!=47.2.0" docutils check-manifest
+    - retry pip install flake8 coverage codecov chardet "setuptools!=47.2.0" docutils
```
larsoner, line 32: flake8 should also be in `dev`

you shouldn't need `coverage` since it's a dependency of `pytest-cov`

peternewman: We don't explicitly have anything which links to flake8 (aside from a few lines of config in setup.cfg), but I can add it in if you'd rather?

I'll drop coverage then.

.travis.yml
```diff
@@ -29,16 +29,17 @@ before_install:
     - source venv/bin/activate
     - python --version  # just to check
     - pip install -U pip wheel # upgrade to latest pip find 3.5 wheels; wheel to avoid errors
-    - retry pip install pytest pytest-cov pytest-dependency flake8 coverage codecov chardet "setuptools!=47.2.0" docutils check-manifest
```
larsoner, line 32: We do use flake8 -- see below the call to `flake8`. In theory we should add a Makefile entry to `Makefile` that runs the steps below. Then `script:` could just be `make test`, where `make test` is an alias for the commands below. So I do think `flake8` should be part of the `requirements_dev` or whatever...

.travis.yml
```diff
@@ -29,16 +29,17 @@ before_install:
...
+    - retry pip install flake8 codecov chardet "setuptools!=47.2.0" docutils
```
larsoner, line 32: `flake8` is still here and shouldn't be necessary...

appveyor.yml
```diff
@@ -13,7 +13,8 @@ cache:
 
 install:
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
-  - "pip install pytest pytest-cov pytest-dependency setuptools flake8 coverage chardet codecov"
+  - "pip install flake8 codecov chardet setuptools"
```
larsoner, line 16: and here

peternewman: Done

.travis.yml
peternewman: Done

.travis.yml
peternewman: Done

.travis.yml
peternewman: Okay I've added make check (rather than test), as the phony was already configured and it's what I've used before, plus more Google hits. I could alias test to check if you'd like?

It runs lots of stuff as you'll see.

Technically the Python dictionary checks get run twice via the basic pytest and the check-dictionaries one, unless we swap to calling sort and trim individually and keeping them in step, there's no easy way round that. I think if you've got a failure it will bomb out anyway.

larsoner:
Thanks @peternewman

PR #1649, opened 2020-08-21
Replace cyrillic letter Es as latin C

Gelma:


bl-ue:
I'll close this for now. We can reopen it if anyone's interested in it.

Issue #1612, opened 2020-08-05
endcode is used intentionally


sebweb3r:
In the dictionary.txt endcode is changed to encode.
However, endcode is often used intentionally: https://grep.app/search?q=endcode

I propose to kick it out of the dictionary or to prevent auto-fix.

peternewman:
Looks like it should be moved to the code dictionary @sebweb3r https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary_code.txt

sebweb3r:
I'm not entirely sure of the purpose of the different dictionaries. Is there an explanation somewhere?

peternewman:
In the help, or here:
https://github.com/codespell-project/codespell/blob/00f8ffd23ddd41bb562c4c1ba2f8df24f0f38cd7/codespell_lib/_codespell.py#L40-L51

So the code one (slightly confusingly) is where we put likely code typos, so you don't use that dictionary if you're checking code (if that makes sense). Or more clearly, the informal one "fixes" informal language, so if you run with that dictionary too, it will remove those "misspellings".

peternewman:
@larsoner writing this, should we add "that might be typos" to the end of code too? Also possibly "for making informal words more formal". Essentially those two (and rare) remove those types of words from the text being checked.

larsoner:
Whatever makes it clearer for people is fine by me

peternewman:
Had a go at this in https://github.com/codespell-project/codespell/pull/1793

PR #1652, opened 2020-08-28
Add storaget->storage

lurch:


codespell_lib/data/dictionary.txt
```diff
@@ -25278,6 +25278,7 @@ stoppping->stopping
 stopps->stops
 stopry->story
 storag->storage
+storaget->storage, target, targets,
```
peternewman, line 25281: Do you have any examples of the latter corrections, they seem rather unlikely.

lurch: Nope, I was only trying to preempt what you'd suggest ;)  Will remove.

sebweb3r: if i use https://grep.app/search?q=storaget&words=true, I only find c++ templates. Wouldn't it be better in the code dictionary?

lurch: I found this typo in a code-comment (on an internal project) as "USB Mass Storaget Interface" which is why I suggested it here.
I've no idea what the "code dictionary" is, and I don't really have time to look into it right now, so I'll let you and @peternewman decide between you what to do.

sebweb3r: To cite Peter "So the code one (slightly confusingly) is where we put likely code typos, so you don't use that dictionary if you're checking code (if that makes sense)." from https://github.com/codespell-project/codespell/issues/1612

peternewman: > Nope, I was only trying to preempt what you'd suggest ;) Will remove.

Heh, I'm not that random am I @lurch ?

> I found this typo in a code-comment (on an internal project) as "USB Mass Storaget Interface" which is why I suggested it here.

Yeah that's great, I was only querying the "target" suggestions as I couldn't see how it would come up.

> if i use https://grep.app/search?q=storaget&words=true, I only find c++ templates. Wouldn't it be better in the code dictionary?

I see a few others dotted around, although the C++ template types do weight it quite heavily that way, so I guess it should go in there to avoid false positives. @sebweb3r do you fancy doing a PR based on this to move it if @lurch doesn't think he'll have time?

sebweb3r: I can hijack this MR, if @lurch is fine with it.

sebweb3r: Or he just puts the line into dictionary_code.txt himself and i add some additional finds. :-)

lurch: Feel free to either add more commits to this PR, or create a new PR and close this one :+1:

lurch:
Replaced by #1663

PR #1626, opened 2020-08-11
More conversions from enGB to enUS (after #1485)

sebweb3r:
The words are in dictionary, but #1485 is needed.

PR #1653, opened 2020-08-28
add corrections for daemon and demon

sebweb3r:


sebweb3r:
how to proceed with words, that are "wrong" in regular english but used in coding?

codespell_lib/data/dictionary_code.txt
```diff
@@ -2,6 +2,15 @@ amin->main
 cas->case, cast,
 clas->class
 cmo->com
+daemon->demon
```
peternewman, line 5: This isn't a typo ever:
https://en.wikipedia.org/wiki/Daemon_(classical_mythology)
https://en.wikipedia.org/wiki/Daemon_(computing)

peternewman:
> how to proceed with words, that are "wrong" in regular english but used in coding?

As per my comment, I'm not convinced they are, certainly one isn't, so I suspect the others probably shouldn't be either if you're using the Greek sense.

sebweb3r:
ok. I reduced it to the enGB->enUS

peternewman (approved):
LGTM. Do you want to add equivalents for daemonize etc? Up to you?

This looks like it exists!
https://linux.die.net/man/1/daemonize

sebweb3r:
I wanted to add them first, but they are not in aspell. Thus, the tests fail. 
I've put them in #1626, which is blocked by the costum #1485.

lurch:
Whilst I agree that `daemon` can be considered a "proper word", I'm not convinced that `daemonize`, `daemonisation`, etc. should ever be considered "proper words", although I accept that they _may_ appear in computing-only contexts?
I.e. in natural-language I think that codespell should suggest `daemonize->demonize`, and only "allow" `daemonize` when checking code. But perhaps that's a level of subtlety that codespell doesn't have yet? :shrug:

codespell_lib/data/dictionary_en-GB_to_en-US.txt
```diff
@@ -33,6 +33,9 @@ crystallisation->crystallization
 customised->customized
 customising->customizing
 defence->defense
+demonise->demonize
+demonised->demonized
```
lurch, line 37: Could also add -es ?

sebweb3r: added

peternewman:
> Whilst I agree that `daemon` can be considered a "proper word", I'm not convinced that `daemonize`, `daemonisation`, etc. should ever be considered "proper words", although I accept that they _may_ appear in computing-only contexts?
> I.e. in natural-language I think that codespell should suggest `daemonize->demonize`, and only "allow" `daemonize` when checking code. But perhaps that's a level of subtlety that codespell doesn't have yet? shrug

They were already in the code dictionary before @sebweb3r deleted them, which would mean if you chose to enable that dictionary (typically **not** in code contexts), it would see them as typos and convert them as you suggested. It does mean you can't have a typo which suggests them if they are corrected to something else elsewhere.

sebweb3r:
There were not in the dictionary. I proposed to add them in this MR (see b3e9127e8594b17fb3090f21d8b78685b79c4ad0), but after your remarks I deleted the addition. :smile:

lurch:
> it would see them as typos and convert them as you suggested. It does mean you can't have a typo which suggests them if they are corrected to something else elsewhere.

I'm afraid you've lost me again :upside_down_face: :laughing:

sebweb3r:
to simplify it:

You are not allowed to have the same word as a right word in one dictionary and as wrong in a different dictionary (except of enGB_to_enUS).

peternewman:
> You are not allowed to have the same word as a right word in one dictionary and as wrong in a different dictionary (except of enGB_to_enUS).

And the others currently defined here (which does need a comment clarifying direction if any):
https://github.com/codespell-project/codespell/blob/master/codespell_lib/tests/test_dictionary.py#L180-L186

> There were not in the dictionary. I proposed to add them in this MR (see [b3e9127](https://github.com/codespell-project/codespell/commit/b3e9127e8594b17fb3090f21d8b78685b79c4ad0)), but after your remarks I deleted the addition. smile

Sorry, I meant you'd added them to achieve what @lurch was suggesting. I only meant and said that `daemon` is never a typo, not the other variants. I haven't done any searching, but I suspect @lurch could be correct about `daemonize` only being valid in computing concepts (and so should be corrected in the code dictionary).

lurch:
I can't find it being used outside of a computing context? https://www.google.com/search?q=daemonize+-computer+-process+-command+-program+-library

peternewman:
> I can't find it being used outside of a computing context? https://www.google.com/search?q=daemonize+-computer+-process+-command+-program+-library

https://www.google.com/search?q=daemonize+-computer+-process+-command+-program+-library+-application+-server+-software+-github+-tool+-bitbucket+-ruby+-java+-php+-perl+-node+-js+-c+-c%2B%2B+-django+-rust+-api+-python+-ubuntu+-debian+-redhat+-ceph+-xen

Finds:
https://www.reddit.com/r/AskLiteraryStudies/comments/8o4zvf/what_is_daemonization/
Also:
https://books.google.co.uk/books?id=ybSsZzPgtgEC&pg=PA153&lpg=PA153&dq=daemonize#v=onepage&q=daemonize&f=false

So it's not exclusively a computing term, but quite rare. But this is rather irrelevant given we don't currently have a computing only dictionary, just the opposite.

lurch:
Wow, you clearly have more google-persistence than me!
Given that you've found it used in a real printed book I'll withdraw my objections. :blue_book:

peternewman:
@sebweb3r do you want to add back the other dea- to dae- words?

sebweb3r:
Which additions exactly do you mean?
For the code dictionary I had those:
```
daemonisation->demonisation, demonization,
daemonise->demonise, demonize,
daemonised->demonised, demonized,
daemonising->demonising, demonizing,
daemonization->demonization, demonisation,
daemonize->demonize, demonise,
daemonized->demonized, demonised,
daemonizing->demonizing, demonising,
```
But you said, that they are real words, that are rarely used outside of the code context.
I would be fine to add them to the code dictionary again (but it is maybe a false positive)

dea- to dae are alreay in.
daemonis*->daemoniz* don't work, due to aspell

peternewman:
Sorry I lost track the other stuff had been merged elsewhere.

> daemonis*->daemoniz* don't work, due to aspell

Okay. They just aren't in aspell at all, so we'd need to have a second dictionary for non-aspell entries or something, so lets give up on them for now.

I'm happy if @lurch is happy with the latest state too.

lurch:
The title of the PR needs updating, but otherwise LGTM!

PR #1656, opened 2020-08-28
Add support for ignoring spelling mistakes in URIs specifically.

jonmeow:
This is per discussion on #676, particularly:
https://github.com/codespell-project/codespell/issues/676#issuecomment-664575379

The URI/email regex is deliberately simple to help with maintainability. False positives are possible but would likely have a low cost associated, particularly as nothing is ignored without setting --uri-ignore-words-list. False negatives would probably cause more trouble, so the regex is over-inclusive.


peternewman:
I suspect we should potentially leave discussion of the regex itself to a separate PR to avoid potentially derailing this one, but here is something interesting to look at/test:
https://mathiasbynens.be/demo/url-regex

codespell_lib/_codespell.py
```diff
@@ -502,8 +520,20 @@ def extract_words(text, word_regex, ignore_word_regex):
...
 
+def apply_uri_ignore_words(check_words, line, word_regex, ignore_word_regex,
+                           uri_regex, uri_ignore_words):
+    if not uri_ignore_words:
+        return
+    for uri in re.findall(uri_regex, line):
+        for uri_word in extract_words(uri, word_regex,
+                                      ignore_word_regex):
+            if "*" in uri_ignore_words or uri_word in uri_ignore_words:
+                check_words.remove(uri_word)
```
peternewman, line 531: Can you not just drop the entire uri if `*` is present? Like your previous PR.

Also won't this currently mean that:
`I like to eat foo you can buy it from my online store at http://example.com/foo`

Would miss foo->food in the free text because it exists in the URI if I set `--uri-ignore-word-list=foo`?

codespell_lib/_codespell.py
```diff
@@ -697,10 +731,8 @@ def main(*args):
             return EX_USAGE
         build_ignore_words(ignore_words_file, ignore_words)
 
-    ignore_words_list = options.ignore_words_list or []
-    for comma_separated_words in ignore_words_list:
-        for word in comma_separated_words.split(','):
-            ignore_words.add(word.strip())
+    uri_regex = re.compile(uri_regex_def)
```
peternewman, line 734: Can you add a try/catch around it like the others and log the failure.

peternewman (changes requested):
A few comments.

codespell_lib/_codespell.py
peternewman: Edit, I see your test says not. Is it actually finding the right one, or just working because there are two hits on the line and it just removes one of them?

jonmeow: It's working because there are two hits on the line, and removing one. codespell doesn't track where the hit was when doing spell corrections, so if there *is* a repeated misspelling on the line, and it's told to fix the spelling, it'll probably fix both. I don't think that's particularly bad though: while something might be ignored in a uri, if it's there in plain text, I'd expect the common case to be that it needs to either be fixed in both or ignored in both. (either way, this is an edge case scenario)

codespell_lib/_codespell.py
jonmeow: Done

jonmeow:
> I suspect we should potentially leave discussion of the regex itself to a separate PR to avoid potentially derailing this one, but here is something interesting to look at/test:
> https://mathiasbynens.be/demo/url-regex

Added tests. I've loosened the regex a little accordingly, too.

codespell_lib/tests/test_basic.py
```diff
@@ -485,7 +486,242 @@ def test_ignore_regex_flag(tmpdir, capsys):
...
+    assert cs.main(f.name, '--uri-ignore-words-list=foo,example') == 1
+    # A word can be ignored.
+    assert cs.main(f.name, '--uri-ignore-words-list=abandonned') == 0
+    assert cs.main(f.name, '--uri-ignore-words-list=foo,abandonned,bar') == 0
+    assert cs.main(f.name, '--uri-ignore-words-list=*') == 0
+    # The match must be for the complete word.
+    assert cs.main(f.name, '--uri-ignore-words-list=abandonn') == 1
+
+    with open(op.join(d, 'flag.txt'), 'w') as f:
+        f.write('abandonned example@abandonned.com\n')
```
peternewman, line 606: I think it would make sense to use abandonned@example.com here, so we've tested both variants.

codespell_lib/_codespell.py
```diff
@@ -28,6 +28,11 @@
 import textwrap
 
 word_regex_def = u"[\\w\\-'’`]+"
+# While we want to treat characters like ( or " as okay for a starting break,
+# these may occur unescaped in URIs, and so we are more restrictive on the
```
peternewman, line 32: Ah I guess my later comment is because of this?

codespell_lib/_codespell.py
```diff
@@ -28,6 +28,11 @@
...
+# endpoint.  Emails are more restrictive, so the endpoint remains flexible.
```
peternewman, line 33: What about `mailto:someone@foo.com?subject=Test`?

codespell_lib/_codespell.py
```diff
@@ -697,10 +738,15 @@ def main(*args):
             return EX_USAGE
         build_ignore_words(ignore_words_file, ignore_words)
 
-    ignore_words_list = options.ignore_words_list or []
-    for comma_separated_words in ignore_words_list:
-        for word in comma_separated_words.split(','):
-            ignore_words.add(word.strip())
+    uri_regex = options.uri_regex or uri_regex_def
+    try:
+        uri_regex = re.compile(uri_regex)
```
peternewman, line 743: I don't know why codecov doesn't think you're testing this,

codespell_lib/_codespell.py
```diff
@@ -697,10 +738,15 @@ def main(*args):
...
+    except re.error as err:
+        print("ERROR: invalid --uri-regex \"%s\" (%s)" %
+              (uri_regex, err), file=sys.stderr)
```
peternewman, line 746: Again, especially as you're testing an invalid regex.

peternewman (changes requested):
A few more comments, I still need to take a final look at the main code changes.

codespell_lib/tests/test_basic.py
jonmeow: Added around line 619

codespell_lib/tests/test_basic.py
```diff
@@ -485,7 +486,242 @@ def test_ignore_regex_flag(tmpdir, capsys):
...
+    assert cs.main(f.name, '--uri-ignore-words-list=*') == 1
+    # The regular --ignore-words-list will ignore both.
+    assert cs.main(f.name, '--ignore-words-list=abandonned') == 0
+
+    variation_option = '--uri-ignore-words-list=abandonned'
+
+    # Variations where an error is ignored.
+    for variation in ('# Please see http://abandonned for info\n',
+                      '# Please see "http://abandonned" for info\n',
+                      '# Please see "http://foo"abandonned for info\n',
```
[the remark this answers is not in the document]
jonmeow: I added this not as an example of what *should* be skipped, but what is *incidentally* skipped. The " is a valid character in URIs after the "/", and I just didn't have two separate regexes for before the / vs after.

I've added a clarifying comment on this line.

codespell_lib/_codespell.py
jonmeow: As above

codespell_lib/_codespell.py
jonmeow: Yeah, I don't know... AFAICT the test `cs.main('--uri-regex=(', std=True)` should be hitting this.

codespell_lib/_codespell.py
jonmeow: Added a test. It's caught with the current regex due to \b matching both : and ?

codespell_lib/_codespell.py
jonmeow: If you mean the `"http://foo"abandonned` example, yes.

jonmeow (commented):
Added a few more tests based on comments.

codespell_lib/_codespell.py
```diff
@@ -502,8 +527,20 @@ def extract_words(text, word_regex, ignore_word_regex):
...
 
+def apply_uri_ignore_words(check_words, line, word_regex, ignore_word_regex,
+                           uri_regex, uri_ignore_words):
+    if not uri_ignore_words:
+        return
+    for uri in re.findall(uri_regex, line):
+        for uri_word in extract_words(uri, word_regex,
+                                      ignore_word_regex):
+            if "*" in uri_ignore_words or uri_word in uri_ignore_words:
+                check_words.remove(uri_word)
```
peternewman, line 538: I wonder if something even more like this might make more sense. Especially because your initial comment was people will just want to skip all typos in URIs wasn't it? Which means that 90% of the time we just want to hide the URI rather than process it twice.

```suggestion
    if "*" in uri_ignore_words:
      text = uri_regex.sub(' ', text)
    return word_regex.findall(text)


def apply_uri_ignore_words(check_words, line, word_regex, ignore_word_regex,
                           uri_regex, uri_ignore_words):
    for uri in re.findall(uri_regex, line):
        for uri_word in extract_words(uri, word_regex,
                                      ignore_word_regex):
            if uri_word in uri_ignore_words:
                check_words.remove(uri_word)
```

@larsoner do you have any strong feelings either way?

codespell_lib/_codespell.py
peternewman: I wonder if something even more like this might make more sense. Especially because your initial comment was people will just want to skip all typos in URIs wasn't it? Which means that 90% of the time we just want to hide the URI rather than process it twice.

```suggestion
    if "*" in uri_ignore_words:
      text = uri_regex.sub(' ', text)
    return word_regex.findall(text)


def apply_uri_ignore_words(check_words, line, word_regex, ignore_word_regex,
                           uri_regex, uri_ignore_words):
    for uri in re.findall(uri_regex, line):
        for uri_word in extract_words(uri, word_regex,
                                      ignore_word_regex):
            if uri_word in uri_ignore_words:
                check_words.remove(uri_word)
```

@larsoner do you have any strong feelings either way?

larsoner: No strong opinion, either way seems fine to me

jonmeow: I've switched the code to check for "*". I don't think it's a performance issue, but whatever's easiest to understand.

I'm not clear what the suggested edit was intended for. If the top half was supposed to be rewriting extract_words, I'd note that extract_words is also used for the options.check_filenames code path, where I felt URI checks shouldn't be done (felt more likely to hit false positives).

jonmeow:
Just checking here, is there still a chance of getting this merged? If not, I'll probably close the PR given its age.

larsoner:
@peternewman did you want to look again?

I'm going to close and reopen so that CIs get restarted -- there are some marked as pending which is weird. But I don't see a problem with merging this given the positive reception from @peternewman before, so feel free to ping me in a few days if you haven't heard anything @jonmeow !

peternewman:
Sorry @jonmeow . I suck, I'll make sure to have a look at this again later today.

codespell_lib/_codespell.py
peternewman: Thanks. Yeah I'm personally generally a fan of not baking config like this in, but providing a sane default e.g. if someone wants a different protocol such as TOR or svn, or wants a more or less restrictive regex like that site lists. Plus with the config file stuff for options we've got now, it's easy for them to just set it once in their repo and forget about it.

codespell_lib/tests/test_basic.py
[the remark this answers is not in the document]
peternewman: Perfect thanks.

codespell_lib/tests/test_basic.py
```diff
@@ -485,7 +486,247 @@ def test_ignore_regex_flag(tmpdir, capsys):
...
+    assert cs.main(f.name, '--ignore-words-list=abandonned') == 0
+
+    variation_option = '--uri-ignore-words-list=abandonned'
+
+    # Variations where an error is ignored.
+    for variation in ('# Please see example@abandonned for info\n',
+                      '# Please see abandonned@example for info\n',
+                      '# Please see abandonned@example.com for info\n',
+                      '# Please see mailto:abandonned@example.com?subject=Test'
+                      ' for info\n'):
```
peternewman, line 624: I don't know if this is actually a match or a miss with the regex, but would be good to get "documented" in a test:
```suggestion
                      '# Please see mailto:abandonned@example.com?subject=Test'
                      ' for info\n',
                      '# Please see mailto:foo@example.com?subject=Test'
                      'abandonned\n',):
```

codespell_lib/_codespell.py
peternewman: It seems to be green here: https://codecov.io/gh/codespell-project/codespell/compare/bc01afd55a91556e21214722ea364b017d2f657e...a5402ec7696c7da5cf05f1d266ff6dd1b8203e6b/diff

codespell_lib/_codespell.py
```diff
@@ -280,7 +285,7 @@ def parse_options(args):
                              'patterns to ignore by treating as whitespace. '
                              'When writing regexes, consider ensuring there '
                              'are boundary non-word chars, e.g., '
-                             '"\\Wmatch\\W". Defaults to empty/disabled.')
+                             '"\\bmatch\\b". Defaults to empty/disabled.')
```
peternewman, line 288: Excellent spot!

I don't know if you fancy fixing the same error here:
https://github.com/codespell-project/codespell/blob/5a8085ac47e4aeeb3a8983c64355a152ff00ad4c/codespell_lib/tests/test_basic.py#L487-L488

Edit, you have already thanks!

peternewman (approved):
Just a few minor comments if you wouldn't mind addressing them.

Really sorry again for the delay.

jonmeow:
Just pinging again, I don't think I can merge this because I don't have write access

codespell_lib/_codespell.py
```diff
@@ -29,6 +29,11 @@
 import textwrap
 
 word_regex_def = u"[\\w\\-'’`]+"
+# While we want to treat characters like ( or " as okay for a starting break,
+# these may occur unescaped in URIs, and so we are more restrictive on the
+# endpoint.  Emails are more restrictive, so the endpoint remains flexible.
+uri_regex_def = (u"(\\b(?:https?|t?ftp|file|git|smb)://[^\\s]+(?=$|\\s)|"
```
bl-ue, line 35: What about `sftp`?
```suggestion
uri_regex_def = (u"(\\b(?:https?|[ts]?ftp|file|git|smb)://[^\\s]+(?=$|\\s)|"
```

codespell_lib/_codespell.py
```diff
@@ -280,7 +285,7 @@ def parse_options(args):
                              'patterns to ignore by treating as whitespace. '
                              'When writing regexes, consider ensuring there '
```
bl-ue, line 286: Can `regexes` be updated to `regular expressions`?

codespell_lib/tests/test_basic.py
peternewman: @jonmeow this one.

codespell_lib/_codespell.py
peternewman: Seems reasonable, although it would be nice to sort the character set:
```suggestion
uri_regex_def = (u"(\\b(?:https?|[st]?ftp|file|git|smb)://[^\\s]+(?=$|\\s)|"
```

peternewman:
> Just pinging again, I don't think I can merge this because I don't have write access

Hi @jonmeow,

I left a few comments in a hopefully final review after you pinged last time, although only one ended up attached to the review, the other was just a comment. Would you mind addressing them?

@bl-ue 's comments don't seem unreasonable (and pretty quick fixes) if you're happy to look after them.

codespell_lib/_codespell.py
jonmeow: Done

jonmeow:
> > Just pinging again, I don't think I can merge this because I don't have write access
> 
> Hi @jonmeow,
> 
> I left a few comments in a hopefully final review after you pinged last time, although only one ended up attached to the review, the other was just a comment. Would you mind addressing them?
> 
> @bl-ue 's comments don't seem unreasonable (and pretty quick fixes) if you're happy to look after them.

Done.

bl-ue (approved):
LGTM.

@jonmeow did you see https://github.com/codespell-project/codespell/pull/1656/files#r619284923?

codespell_lib/tests/test_basic.py
jonmeow: This suggestion is missing a space in `Testabandonned`, but I'm assuming that's an oversight rather than deliberate: added a test around line 633 similar to the test on line 581.

jonmeow (commented):
> LGTM.
> 
> @jonmeow did you see https://github.com/codespell-project/codespell/pull/1656/files#r619284923?

Thanks, I wasn't seeing that.

codespell_lib/tests/test_basic.py
peternewman: Oops, it was indeed @jonmeow . I was actually thinking of the example where the subject of the email was "Test abandonned", which I guess probably needs to be entered as "Test%20abandonned" or similar, which probably breaks any cleverness we have in place currently.

codespell_lib/_codespell.py
peternewman: Now done.

peternewman (approved):
Thanks very much for this @jonmeow it looks great. Sorry it's taken quite so long to get merged!

PR #1658, opened 2020-08-30

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -57,6 +57,7 @@ abnornally->abnormally
 abnove->above
 abnrormal->abnormal
 aboce->above
+aboluste->absolute
```
lurch, line 60: Could also add abolustely->absolutely ?

codespell_lib/data/dictionary.txt
```diff
@@ -14870,6 +14873,7 @@ indicaite->indicate
 indicat->indicate
 indicateds->indicated, indicates,
 indicatee->indicates, indicated,
+indicats->indicates
```
lurch, line 14876: Perhaps this should also suggest `indicate`, as it's only one-letter different from `indicats`?

luzpaz:
@lurch see d6c941f

codespell_lib/data/dictionary.txt
```diff
@@ -6976,7 +6978,9 @@ copstruction->construction
 copuright->copyright
 copurighted->copyrighted
 copurights->copyrights
+copute->compute
 coputer->computer
+coputes->computes
```
peternewman, line 6983: Do you want to add the -ed version of this typo too.

peternewman (changes requested):
Just one comment

codespell_lib/data/dictionary.txt
luzpaz: 3055a10

peternewman (approved):
LGTM

peternewman:
I'm happy. @lurch do you want to re-review. @luzpaz please can you fix the conflicts.

codespell_lib/data/dictionary.txt
```diff
@@ -14954,6 +14960,11 @@ inexpirienced->inexperienced
 infact->in fact
 infalability->infallibility
 infallable->infallible
+infaltable->inflatable, infallible,
+infalte->inflate
+infalted->inflated
+infaltes->inflates
```
lurch, line 14966: Do you think it's also worth adding `infants` here, or is that too different to the original typo?

codespell_lib/data/dictionary.txt
```diff
@@ -28140,6 +28164,8 @@ untoched->untouched
 untqueue->unqueue
 untrached->untracked
 untranslateable->untranslatable
+untrasform->untransform
```
lurch, line 28167: Whilst `untransformed` seems a proper adjective, I'm not convinced that `untransform` is a proper verb? However it seems quite common on grep.app so maybe it should be added to the code dictionary? :shrug:

codespell_lib/data/dictionary.txt
peternewman: It feels rather a stretch to me, it's `lte`->`n` and all of those letters are at least two away from n on QWERTY, so it seems rather unlikely to me.

codespell_lib/data/dictionary.txt
peternewman: > However it seems quite common on grep.app so maybe it should be added to the code dictionary? :shrug:

The code dictionary is just for typos you **DON'T** want replacing within code, for example `uint->unit` would play havoc with unsigned integers.

I'd probably agree with you @lurch , the verb isn't directional like melt for example. If we can think of a better option say `reverse transform`, we should probably add that as a suggestion too, but probably leave `untransform` in as the clear correction of this typo, i.e. when someone typed `untrasform` they clearly meant to type `untransform` even if grammatically they should have used something else.

We don't currently have a dictionary where the correction is questionable/rare!

lurch: All good points. Perhaps "undo transform" is a better suggestion for "untransform" than "reverse transform" ?

luzpaz: clarification, how should the final entry look like?  
`untrasform->untransform, undo transform,`

peternewman: Yes probably @luzpaz , you could also suggest `reverse transform` too, given the other suggestions are getting into grammar and it probably depends on the context of the sentence.

I think this is the second word we've had recently where the correction is questionable, so perhaps its time to add a dictionary for those...

luzpaz: See 00e50e7

peternewman: Thanks @luzpaz , sorry it was just for untransform though, not untransformed so if you can remove the extra suggestions on untransformed please.

sebweb3r: push

luzpaz:
Ok, done.

PR #1663, opened 2020-08-30
Replacement of #1652

sebweb3r:
Hijacking the MR of @lurch, since I cannot push into his branch

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25278,7 +25278,6 @@ stoppping->stopping
 stopps->stops
 stopry->story
 storag->storage
-storaget->storage
 storeage->storage
 stoream->stream
 storeis->stories
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25278,8 +25278,10 @@ stoppping->stopping
 stopps->stops
 stopry->story
 storag->storage
+storeable->storable
 storeage->storage
 stoream->stream
+storeble->storable
 storeis->stories
 storise->stories
 stornegst->strongest
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25278,7 +25278,7 @@ stoppping->stopping
 stopps->stops
 stopry->story
 storag->storage
-storaget->storage, target, targets,
+storaget->storage
 storeage->storage
 stoream->stream
 storeis->stories
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25286,7 +25286,7 @@ storeage->storage
 stoream->stream
 storeble->storable
 storeing->storing
-storeis->stories, stores,
+storeis->stories, stores, storeys,
 storise->stories
 stornegst->strongest
 stoyr->story
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -23100,6 +23100,9 @@ restoding->restoring
 restoiring->restoring
 restor->restore
 restorated->restored
+restoreable->restorable
+restoreble->restorable
+restoreing->restoring
 restors->restores
 restouration->restoration
 restraunt->restraint, restaurant,
@@ -25282,6 +25285,7 @@ storeable->storable
 storeage->storage
 stoream->stream
 storeble->storable
+storeing->storing
 storeis->stories
 storise->stories
 stornegst->strongest
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25278,6 +25278,7 @@ stoppping->stopping
 stopps->stops
 stopry->story
 storag->storage
+storaget->storage, target, targets,
 storeage->storage
 stoream->stream
 storeis->stories
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25286,7 +25286,7 @@ storeage->storage
 stoream->stream
 storeble->storable
 storeing->storing
-storeis->stories, stores, storeys,
+storeis->stories, stores, store is, storeys,
 storise->stories
 stornegst->strongest
 stoyr->story
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25286,7 +25286,7 @@ storeage->storage
 stoream->stream
 storeble->storable
 storeing->storing
-storeis->stories
+storeis->stories, stores,
 storise->stories
 stornegst->strongest
 stoyr->story
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -21,6 +21,7 @@ seeked->sought
 sinc->sync, sink, since,
 sincs->syncs, sinks, since,
 stoll->still
+storaged->storage, stored,
 storaget->storage
 thead->thread
 todays->today's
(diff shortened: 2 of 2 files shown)


codespell_lib/data/dictionary_code.txt
```diff
@@ -21,6 +21,7 @@ seeked->sought
 sinc->sync, sink, since,
 sincs->syncs, sinks, since,
 stoll->still
+storaget->storage, storaged,
```
lurch, line 24: `storaged` seems like a very odd suggestion? (but again perhaps this is because I don't understand the different dictionaries? :confused: )

codespell_lib/data/dictionary.txt
```diff
@@ -25278,8 +25281,11 @@ stoppping->stopping
 stopps->stops
 stopry->story
 storag->storage
+storeable->storable
 storeage->storage
 stoream->stream
+storeble->storable
+storeing->storing
 storeis->stories
```
lurch, line 25289: I guess you could also add a suggestion of `store is` here?

lurch:
Should I go ahead and close #1652 now?

peternewman:
> Should I go ahead and close #1652 now?

Yes please, as @sebweb3r seems to have pulled in all your commits.

codespell_lib/data/dictionary_code.txt
sebweb3r: I found it as a valid word. I will kick it out, If you think it is not valid. 
The dictionary should not influence the suggestions (as far as I have understood it)

peternewman: Do you have a citation @sebweb3r ?

sebweb3r: https://en.wiktionary.org/wiki/storaged

peternewman: A proper citation, from a real dictionary? ( I'm just going to leave this here... https://www.theregister.com/2020/08/26/scots_wikipedia_fake/ )

Also:
```
storaged
    simple past tense and past participle of storage
```

Versus:
`store (<snip> simple past and past participle stored) `

I'd expect stored too, I can't think of a way to use storaged in a sentence).

peternewman: > The dictionary should not influence the suggestions (as far as I have understood it)

Also yes, you're correct, the multiple dictionaries are all about avoiding different classes of typos (e.g. real rare words, where the correction is more likely than the typo, or names or computing terms that would be typos outside those contexts)

sebweb3r: Thats why you are the gatekeeper :-)

I just checked, if there is an other solution possible. I wouldn't use storaged either, but I was not sure, if it is valid.
For english it is quite hard to find a reliable.
daemonize f.e. is also not in real dictionaries like https://dictionary.cambridge.org/spellcheck/english/?q=daemonize

lurch: I've **never** heard "storage" used as a verb, only ever as a noun.

I think "daemonize" is a computing-specific term, derived from the "Unix daemon" sense.
It's probably used in a "making a process into a daemon (background task)" sense, but it's not _really_ a real word.

sebweb3r: I just referenced daemon due to this discussion https://github.com/codespell-project/codespell/pull/1653#issuecomment-683221197 :smile:

I'm fine with your suggestions. You are the native speakers.

I deleted storaged.

sebweb3r: Storaged has 424 hits on grep.app
https://grep.app/search?q=storaged&words=true
I would add this as "storaged->storage, stored," too, if you agree.

lurch: Yeah, I'd guess in code-terms "storaged" is probably an abbreviation for "storage daemon", much like the way "maild" is an abbreviation for "mail daemon" (also ftpd, httpd, etc.).

sebweb3r: Yes, but there is a lot of people, who want to use stored too :-)

codespell_lib/data/dictionary.txt
sebweb3r: I'm not sure if it is wanted to split up words like that. But if it is, I will add more of those combinations.

sebweb3r: Maybe comments/consent of @peternewman

lurch: and perhaps storeis -> stores is also a reasonable suggestion? :shrug:

sebweb3r: added

codespell_lib/data/dictionary.txt
```diff
@@ -25278,9 +25281,12 @@ stoppping->stopping
...
 stopry->story
 storag->storage
+storeable->storable
 storeage->storage
 stoream->stream
-storeis->stories
+storeble->storable
+storeing->storing
+storeis->stories, stores,
 storise->stories
```
lurch, line 25290: Haha, this line suddenly made me think of tortoise ! :turtle: 
No, I'm not suggesting that :grinning:

sebweb3r: :-D

peternewman, line 25290: Suggest storeys too

peternewman (changes requested):
One comment

codespell_lib/data/dictionary.txt
sebweb3r: added

codespell_lib/data/dictionary.txt
peternewman: Also storeys (again).

> I'm not sure if it is wanted to split up words like that. But if it is, I will add more of those combinations.
> Maybe comments/consent of @peternewman

You don't ever need my consent. Worse case if you add it, myself or someone else reviewing may suggest it gets removed if it seems unlikely.

I'd say generally, if it's two valid words, then the typo may be they missed the space bar (rather than another letter), so go for it. If you want to be really keen, try a search for that pair adjacent. I guess there might be some pairs that are unlikely to occur together, but I'm struggling to think of good examples (e.g. book door; perhaps the hypothetical typo meant backdoor, although even then "I want to book door to door service").

sebweb3r: I wanted to ask, before i search for all typos like that.

peternewman: Okay, do you want to add the two specific fixes for this into here so it can be merged, then do another PR for the other stuff.

It might even be worth writing some sort of script that splits each word in the dictionary in two at every possible boundary and then checks each resulting word against aspell. Or just start with a list of common 1-3 letter words (like a, in, is, it) and see if they are prefixes or suffixes of the dictionary words (and check the resulting word is valid too).

sebweb3r: I would put all the split up in a separate PR. It makes it easier to review and revert if necessary.

I'm probably going to make it semi-automatic.

peternewman: > I would put all the split up in a separate PR. It makes it easier to review and revert if necessary.

Can you fix this one now we've found it.

I'm saying don't drop a PR with 300 changes in one PR split it up into chunks, as otherwise it ends up like #1624 that becomes a beast that's a challenge to get over the line.

> I'm probably going to make it semi-automatic.

Good, makes sense.

lurch:
Perhaps `storys->stories, storeys,` should be added too?

codespell_lib/data/dictionary.txt
sebweb3r: I was misleading. I meant "I want to put this in (one of the multiple) new MR(s)

added the suggestion.

peternewman (approved):
LGTM thanks!

peternewman:
@lurch any more comments before I merge it?

lurch:
LGTM!

lurch:
Thanks for taking this on @sebweb3r on my behalf :+1:

sebweb3r:
you're welcome :+1:

PR #1669, opened 2020-09-01
Update README around multi-dictionary changes

peternewman:


README.rst
```diff
@@ -95,38 +93,33 @@ directly, but instead be manually inspected. E.g.:
        fiel->feel, field, file, phial,
 
    Note the last comma! You need to use it, otherwise the last suggestion
-   will be discarded (see below for why). When there is more than one
+   will be discarded (for historic reasons). When there is more than one
```
peternewman, line 96: Is now the time to remove the restriction or re-implement this feature (or are we worried about people running new dictionaries against old code? I guess it's not too arduous a requirement to keep for future re-implementation.

larsoner: It's possible that people have their own custom dictionaries and it would be good not to break them or force them to update them

README.rst
```diff
@@ -95,38 +93,33 @@ directly, but instead be manually inspected. E.g.:
...
    suggestion, an automatic fix is not possible and the best we can do is
    to give the user the file and line where the error occurred as well as
    the suggestions.
 
-3. Entry with one word, but with automatically fix disabled::
```
larsoner, line 103: Did we decide to stop officially supporting this?

README.rst
peternewman: Good point. We could add it back, although I guess if you're using e.g. a code word, you hopefully know what it is and why. I guess showing which dictionary it's come from is more initially valuable. This is just spit-balling anyway.

README.rst
peternewman: :shrug: But this is always telling me off for not adding one:
https://github.com/codespell-project/codespell/blob/master/codespell_lib/tests/test_dictionary.py#L110-L112

I suspect most of the time the risk of missing one and it becoming a reason is higher than the times we need to disable.

Also we now disable by just turning off a whole dictionary or skipping a specific word.

larsoner: It's possible people are using this feature, we should probably have some test that uses the functionality and makes sure it still works / disables autocorrection

larsoner: (i.e., just because internally we don't use it doesn't mean people's custom dictionaries don't. And it seems pretty easy to keep supporting this, assuming it currently does still work.)

peternewman: Yep true. Although by that argument we should possibly allow it via make check-dictionary too? We could also check for multiple words in the reason to catch missing accidental commas (until the multi-word thing comes in via #1657 )

larsoner: What would we need to change in `make check-dictionaries`? I don't see anything. The first steps don't check for commas (do they?), then the `pytest ...` line checks the internal ones, where we don't allow our internal dictionaries to have such entries.

sebweb3r:
Would you mind to add a table with the existing dictionaries and shortly say, which are used as a default?

I know it now, but the question would be helpful for new people using the code.

README.rst
peternewman: I meant to allow the pytest bit to work, while it's not required, it seems rather odd if someone can't just add their own dictionary to our file to easily run all our tests. Perhaps it wants an option like the aspell stuff to disable that test?

sebweb3r:
Adding a reference to the builtin section is fine for me. But if you just read the dictionary part and Readme, it is not really clear.

larsoner:
> I don't really want to add something separately as it's just another thing to forget to update and become stale.

I would add a tiny test (pytest) that looks for `../README.rst`, does a quick parse, and checks that there is a match. That would mitigate the risk.

It would be good to have the dictionary list in human-readable form. We can also tell people to do `codespell --help`, where they all should be listed.

README.rst
larsoner: > it seems rather odd if someone can't just add their own dictionary to our file to easily run all our tests. 

This suggests we should have a separate command like `codespell validate <filename>` that runs the Python-level validation on a given dictionary.

But even if we do that, we'd probably have switches for which things to check for. For our internal ones, we wouldn't allow the `, disabled because ...` because we don't need them, but for other people's dictionaries, we would for backward compatibility reasons.

So I'm still -1 on removing this information

README.rst
```diff
@@ -67,17 +67,19 @@ Display them without terminal colors and with a quiet level of 3. ::
...
 
-We ship a dictionary that is an improved version of the one available
+We ship a collection of dictionaries that are an improved version of the one available
 `on Wikipedia <https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines>`_
 after applying them in projects like Linux Kernel, EFL, oFono among others.
 You can provide your own version of the dictionary, but patches for
 new/different entries are very welcome.
 
-Want to know if a word you're proposing exists in codespell already? It is possible to test a word against the current dictionary that exists in ``codespell_lib/data/dictionary.txt`` via::
+Want to know if a word you're proposing exists in codespell already? It is possible to test a word against the current set dictionaries that exists in ``codespell_lib/data/dictionary*.txt`` via::
```
larsoner, line 76:
```suggestion
Want to know if a word you're proposing exists in codespell already? It is possible to test a word against the current set dictionaries that exist in ``codespell_lib/data/dictionary*.txt`` via::
```

larsoner (commented):
@peternewman do you want to take a stab at a better phrasing inspired by #1535 so we can close that one when we merge this, then cut a release?

larsoner:
Ping @peternewman , would be nice to move on this so we can release

README.rst
```diff
@@ -67,17 +67,19 @@ Display them without terminal colors and with a quiet level of 3. ::
...
 
     echo "word" | codespell -
     echo "1stword,2ndword" | codespell -
 
+You can select the optional dictionaries with the ``--builtin` option.
```
larsoner, line 81:
```suggestion
You can select the optional dictionaries with the ``--builtin`` option.
```

larsoner:
Thanks @peternewman

peternewman:
> Ping @peternewman , would be nice to move on this so we can release

Sorry @larsoner , various stuff got a bit mad, plus trying to hit a release deadline for something else I help maintain.

I'll generate some issues for the other things we discussed.

PR #1602, opened 2020-07-24
Various changes while spellchecking protobuf

peternewman:

PR #1672, opened 2020-09-02
Dedup->dedupe->deduplicate

peternewman:
See also notes in #1602


larsoner:
Travis is not happy

sebweb3r:
Travis will never be happy with that. Informal requires both error and replacement to be in aspell. None of the three words is in aspell.
http://app.aspell.net/lookup?dict=en_US;words=dedup%0D%0Adedupe%0D%0Adeduplicate

larsoner:
> Travis will never be happy with that. Informal requires both error and replacement to be in aspell. None of the three words is in aspell.

Travis will do whatever we say :) Concretely, I'm fine with modifying the test suite to have exceptions (`# do not check aspell for this set of words`) in cases like these if it's justified to do so

sebweb3r:
Ok, let me rephrase it. Without bending travis, it won't accept it.

This basically follows this trend:
https://github.com/codespell-project/codespell/issues/1711

lurch:
I suspect these commits need squashing? :wink:

bl-ue:
@peternewman can I merge this PR? (**SQUASH definitely!!**)

DimitriPapadopoulos:
@peternewman It should be possible to merge this pull request if you resolve the conflicts and put words missing from _aspell_ under `codespell_lib/tests/data`.

PR #1676, opened 2020-09-03
Add atmoically->atomically to dictionary

tinchoabbate:


peternewman (changes requested):
Thanks for your contribution @tinchoabbate !

As per the annotation you need to sort the dictionary either by running `make sort-dictionaries` and then recommitting, or by doing so manually yourself.

codespell_lib/data/dictionary.txt
```diff
@@ -2647,6 +2647,7 @@ atll->all
 atomatically->automatically
 atomical->atomic
 atomicly->atomically
+atmoically->atomically
```
sebweb3r, line 2650: Could you add atmoic->atomic too plz?

PR #1684, opened 2020-09-21
Move correct words with ' to code-dict

### LINKED ISSUE #1682: false positive at boundary of a string literal
```python
def test():
    out = ...
    expected = (
        "The exclude pattern 'foo' for check-useless-excludes "
        'does not match any files'
    )
    assert expected == out
```

```console
$ codespell t.py 
t.py:6: files' ==> file's
```

sebweb3r:
Closes #1682


peternewman (approved):
LGTM

PR #1686, opened 2020-09-21
More words

sebweb3r:


codespell_lib/data/dictionary.txt
```diff
@@ -16564,7 +16574,8 @@ lpatform->platform
 lsat->last
 lsit->list
 lsits->lists
-lukid->likud
+lukid->lucid, likud,
```
sebweb3r, line 16577: likud is the name of an israelian party. should it be a suggestion?

codespell_lib/data/dictionary.txt
```diff
@@ -10068,6 +10070,9 @@ ecret->secret, erect,
 ecspecially->especially
 ect->etc
 ecxept->except
+ecxite->excite
+ecxited->excited
```
lurch, line 10074: -ing ?

codespell_lib/data/dictionary.txt
```diff
@@ -11437,6 +11442,9 @@ existsing->existing
 existung->existing
 existy->exist
 existying->existing
+exitation->excitation
+exitations->excitations
+exite->exit, excite,
```
lurch, line 11447: also exits ?

codespell_lib/data/dictionary.txt
```diff
@@ -15899,6 +15911,7 @@ iresistibly->irresistibly
 iritable->irritable
 iritated->irritated
 ironicly->ironically
+irradation->irradiation
```
lurch, line 15914: also irradate->irradiate ?

(and perhaps also iradiate, iradate, iradiation, iradation etc.)

Could maybe also add `irigation->irrigation` (etc.) if you're feeling enthusiastic? :wink: 

And I've just spotted (sorry!) that the dictionary has `iritated->irritated` but not `iritate->irritate`.

codespell_lib/data/dictionary.txt
```diff
@@ -16564,7 +16577,8 @@ lpatform->platform
 lsat->last
 lsit->list
 lsits->lists
-lukid->likud
+lukid->lucid, likud,
+lumination->lamination, illumination,
```
lurch, line 16581: Hmmm? https://www.collinsdictionary.com/dictionary/english/lumination

codespell_lib/data/dictionary.txt
```diff
@@ -16564,7 +16577,8 @@ lpatform->platform
 lsat->last
 lsit->list
 lsits->lists
-lukid->likud
+lukid->lucid, likud,
```
lurch, line 16580: Maybe also add `lusid->lucid` ?

codespell_lib/data/dictionary_en-GB_to_en-US.txt
```diff
@@ -101,6 +101,8 @@ organisations->organizations
 organise->organize
 organised->organized
 plagiarise->plagiarize
+polarise->polarize
+polarised->polarized
```
lurch, line 105: Also the valid ones from this list? http://app.aspell.net/lookup?dict=en_GB-ise&words=polarise%0D%0Apolariser%0D%0Apolarising%0D%0Apolarised%0D%0Apolarises

codespell_lib/data/dictionary.txt
```diff
@@ -14271,6 +14279,10 @@ illustrasion->illustration
 illution->illusion
 ilness->illness
 ilogical->illogical
+iluminate->illuminate
+iluminated->illuminated
+iluminates->illuminates
+ilumination->illumination
```
lurch, line 14285: illuminations?

codespell_lib/data/dictionary.txt
sebweb3r: All Dictionaries I have looked it up, don't know it.
https://dictionary.cambridge.org/de/spellcheck/englisch/?q=lumination

I will have a look on some physics papers later, if it was/is used by people.

sebweb3r: I think most people, who use it, want something like luminance?!

codespell_lib/data/dictionary.txt
sebweb3r: added stuff, except irigat*

codespell_lib/data/dictionary.txt
sebweb3r: Only use of lusid look intentional 

https://www.grep.app/search?q=lusid&words=true

lurch: Fair enough, I was just thinking that somebody who is bad at spelling, might misspell it that way. But perhaps I was over-reaching.

codespell_lib/data/dictionary.txt
peternewman: Yeah probably. The fact someone added that as the initial entry hopefully means they found it as a typo. I think putting it second is right though. Should it be capitalised?

codespell_lib/data/dictionary.txt
peternewman: Chuck it in rare:
http://app.aspell.net/lookup?dict=en_US&words=lumination%0D%0Aillumination

codespell_lib/data/dictionary.txt
```diff
@@ -20874,6 +20893,7 @@ programable->programmable
 programatic->programmatic
 programatically->programmatically
 programattically->programmatically
+programed->programmed
```
peternewman, line 20896: You could also add `programmd->programmed`.

Also not yours but I've just seen `programm->program` should also suggest the BE `programme`

peternewman (changes requested):
Plus the capitalisation of Likud if necessary

codespell_lib/data/dictionary.txt
sebweb3r: I wanted to add programme->program to enGB_to_enUS, 
but it is not in aspell http://app.aspell.net/lookup?dict=en_US&words=programm%0D%0Aprogramme

codespell_lib/data/dictionary.txt
```diff
@@ -20874,12 +20892,15 @@ programable->programmable
...
+programd->programmed
+programed->programmed
 programemer->programmer
 programemers->programmers
 programers->programmers
 programing->programming
-programm->program
+programm->program, programme,
 programmaticaly->programmatically
+programmd->programmed
```
peternewman, line 20903: I think this should suggest programme too given the d/e proximity.

peternewman (changes requested):
Just one comment from me

codespell_lib/data/dictionary.txt
sebweb3r: added and rebased.

peternewman (approved):

codespell_lib/data/dictionary.txt
lurch: It is if you look in the right dictionary? http://app.aspell.net/lookup?dict=en_GB-ise&words=program%0D%0Aprogramme

The way that I've always understood it, is that in the UK we have television programmes and computer programs, but in the US they use 'program' for both.

sebweb3r: Ok, I relied on the feature "show it to me, if it is in other dictionaries." which doesn't seem to work here. My local aspell test fails too.
let's try, what jenkins says to #1688

PR #1688, opened 2020-09-23
Dict: More words

sebweb3r:


sebweb3r:
@lurch see. it is strange.

sebweb3r:
@peternewman The words are in aspell [1] but not shown if you look for US [2] while other words are found. Jenkins doesn't like it too.

[1] http://app.aspell.net/lookup?dict=en_GB-ise&words=program%0D%0Aprogramme%0D%0Aprogrammes
[2] http://app.aspell.net/lookup?dict=en_US&words=program%0D%0Aprogramme%0D%0Aprogrammes%0D%0Acentres

peternewman:
> @peternewman The words are in aspell [1] but not shown if you look for US [2] while other words are found. Jenkins doesn't like it too.
> 
> [1] http://app.aspell.net/lookup?dict=en_GB-ise&words=program%0D%0Aprogramme%0D%0Aprogrammes
> [2] http://app.aspell.net/lookup?dict=en_US&words=program%0D%0Aprogramme%0D%0Aprogrammes%0D%0Acentres

That's because my naive dictionary implementation only uses one dictionary (in this case en_US), which it seems some GB words (like colour) are in, but not others. The fix will be to have two aspell instances and use the appropriate one for each dictionary, which probably means tracking what language it's in. That would also be problematic for typos for GB words when we start getting stricter about them being in the dictionary. Or the large one might fix it. :crossed_fingers:

sebweb3r:
Ok, that sound like a stupid implementation in aspell. To cite the description of aspell-en in debian "American, British, Canadian and Australian spellings are included." :-D

peternewman:
> Ok, that sound like a stupid implementation in aspell. To cite the description of aspell-en in debian "American, British, Canadian and Australian spellings are included." :-D

Or it might be my usage of it, perhaps we can/need to pass in multiple languages at once.

codespell_lib/data/dictionary.txt
```diff
@@ -15506,6 +15510,8 @@ intenational->international
 intendet->intended
 inteneded->intended
 intension->intention
+intensional->intentional
+intensionally->intentionally
```
lurch, line 15514: Could also add the variants with a single `l` ?

sebweb3r: like abde9cf25a6c7af1b81c4eb027d17884bf0e2228?

lurch: Yep :+1:   Could also add `intentialy` and `itentionaly`? :shrug:

lurch:
The way the aspell lookup works seems a bit weird...

http://app.aspell.net/lookup?dict=en_US&words=program%0D%0Aprogramme%0D%0Acolor%0D%0Acolour says that `program` and `color` are in en_US, `colour` is in en_GB-ise, but `programme` is only in SCOWL.
(which implies that `programme` isn't in en_GB-ise?).

http://app.aspell.net/lookup?dict=en_GB-ise&words=program%0D%0Aprogramme%0D%0Acolor%0D%0Acolour says that `program`, `programme` and `colour` are in en_GB-ise, but `color` is only in SCOWL.
(which implies that `color` isn't in en_US?) :confused:

peternewman (approved):

PR #1692, opened 2020-09-26
Add spelling corrections for resoruce

fishilico:
Many projects use "Resoruce" instead of "Resource": https://grep.app/search?q=resoruce

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -23128,6 +23128,8 @@ resopnses->responses
 resorce->resource
 resorces->resources
 resore->restore
+resoruce->resource
+resoruces->resources
 resouce->resource
 resouces->resources
 resoultion->resolution
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -22037,7 +22037,9 @@ recordproducer->record producer
 recored->recorded
 recoriding->recording
 recource->resource
+recourced->resourced
 recources->resources
+recourcing->resourcing
 recquired->required
 recrational->recreational
 recreateation->recreation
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -23128,6 +23128,8 @@ resopnses->responses
 resorce->resource
 resorces->resources
 resore->restore
+resoruce->resource
+resoruces->resources
```
peternewman, line 23132: Perhaps add resourced and resourcing equivalent typos too?

peternewman (changes requested):
Just one comment

codespell_lib/data/dictionary.txt
fishilico: I agree and added these typos for all existing `->resource` and `->resources`.

lurch:
I wonder if it'd be worth also adding the same typos but for 'source' (and variants) instead of 'resource'? :man_shrugging:

fishilico:
> I wonder if it'd be worth also adding the same typos but for 'source' (and variants) instead of 'resource'?

The dictionary already contains corrections for `soruce` and `soruces`, so there is no need to add them: https://github.com/codespell-project/codespell/blob/89be12128e05a3bcc98194c7115991ac975ea1e4/codespell_lib/data/dictionary.txt#L24781-L24782

But the dictionary does not contain `soruced` nor `sorucing` and some open-source projects did these misspellings: https://grep.app/search?q=soruced and https://grep.app/search?q=sorucing. Would you prefer that I add these corrections to this PR, or that I open a new PR about "source and its derived words"?

lurch:
> Would you prefer that I add these corrections to this PR, or that I open a new PR about "source and its derived words"?

That's a question for @peternewman ...

peternewman:
> > Would you prefer that I add these corrections to this PR, or that I open a new PR about "source and its derived words"?
> 
> That's a question for @peternewman ...

Sorry for the delay. A separate PR is probably best as this one is already relatively long and those new words may bring up other alternative suggestions that want looking at.

PR #1693, opened 2020-09-26
Add spelling corrections for tamplate

fishilico:
Many projects use "Tamplate" (and derivatives) instead of "Template": https://grep.app/search?q=tamplate


peternewman (approved):
Good find thanks @fishilico .

PR #1695, opened 2020-09-27
Dict: Add more words

sebweb3r:


codespell_lib/data/dictionary.txt
```diff
@@ -5324,6 +5324,7 @@ coloum->column
 coloumn->column
 coloumns->columns
 coloums->columns
+colourfull->colourful
```
peternewman, line 5327: Suggest colourfully too.

Does this exist in the US too, or do they use two l's?

peternewman (changes requested):
One comment

codespell_lib/data/dictionary.txt
sebweb3r: colorfull->colorful was already in.

I added the ly variant for all.

https://dictionary.cambridge.org/de/worterbuch/englisch/colorful

peternewman (approved):

PR #1699, opened 2020-10-13
Add misspelled words from EC reviews

thughes:
See https://chromium-review.googlesource.com/c/chromiumos/platform/ec/+/2092492/2/chip/stm32/registers-stm32g4.h#550


thughes:
Woops, these are already in the dictionary.

PR #1700, opened 2020-10-14

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -1371,6 +1371,7 @@ allwys->always
 almostly->almost
 almsot->almost
 alo->also
+aloable->allowable
```
peternewman, line 1374: This seems quite a stretch!

The only hit grep.app had was avaloable which should be to available so would you mind adding that?

codespell_lib/data/dictionary.txt
```diff
@@ -23963,6 +23966,7 @@ self-comparisson->self-comparison
 self-contianed->self-contained
 self-referencial->self-referential
 self-refering->self-referring
+selt->set, self, sold,
```
peternewman, line 23969: Suggest seat.

I'm struggling to see sold though?

From looking at grep.app I think it should be self and then set.

codespell_lib/data/dictionary.txt
```diff
@@ -28809,6 +28814,7 @@ vissible->visible
 visted->visited
 visting->visiting
 vistors->visitors
+visualyse->visualize
```
peternewman, line 28817: Can you suggest visualise first.

Can you also add an equivalent typo for visualiser/visualizer.

peternewman (changes requested):
A few comments

codespell_lib/data/dictionary.txt
luzpaz: stretch, you say? github says otherwise: https://github.com/search?q=%22aloable%22&type=code

peternewman: Okay, although they seem to be one typo in lots of forks of the same repo. I guess it's fixed upstream hence why grep.app isn't returning it or something?

Anyway fair enough.

codespell_lib/data/dictionary.txt
luzpaz: `He selt it last week` a foreign speaker could make this mistake.

luzpaz:
Revisions added to 8b02264, please confirm

codespell_lib/data/dictionary.txt
peternewman: Okay, I thought that would be more likely to be "selled it" (do you want to add that too if it's not there.

Can you swap self and set though please.

codespell_lib/data/dictionary.txt
peternewman: Thanks, can you add the `visualyser` equivalent please.

PR #1702, opened 2020-10-22
Add ehen->when

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -10188,6 +10188,7 @@ ehance->enhance
 ehanced->enhanced
 ehancement->enhancement
 ehancements->enhancements
+ehen->when
```
peternewman, line 10191: Suggest hen too as a typo.

Add ehenever as another typo too perhaps?

peternewman (changes requested):
One comment

codespell_lib/data/dictionary.txt
lurch: Could also suggest ehen->even ?

luzpaz: Added in bb06c83

lurch:
@peternewman Hmm, should `ehen->then` also be suggested, or is that "too distant"? :shrug:

peternewman:
> ehen

Yeah seems like it could possibly happen on a Qwerty keyboard.

luzpaz:
Added in 355a784

peternewman (approved):

PR #1705, opened 2020-10-22
Add corrections for update*

cfi-gb:
See e.g.:

https://grep.app/search?q=udated&words=true

I have put udate/udates into the dictionary_code.txt because this seems to be a common variable name:

https://grep.app/search?q=udate&words=true

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -27667,8 +27667,9 @@ ubitquitous->ubiquitous
 ublisher->publisher
 ubunut->Ubuntu
 ubutunu->Ubuntu
-udated->updated
-udater->updater
+udated->updated, dated,
+udater->updater, dater,
+udating->updating, dating,
 udno->undo, uno,
 udpatable->updatable
 udpate->update
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -27667,6 +27667,8 @@ ubitquitous->ubiquitous
 ublisher->publisher
 ubunut->Ubuntu
 ubutunu->Ubuntu
+udated->updated
+udater->updater
 udno->undo, uno,
 udpatable->updatable
 udpate->update
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -39,7 +39,7 @@ storaget->storage
 subprocess'->subprocess's
 thead->thread
 todays->today's
-udate->update
-udates->updates
+udate->update, date,
+udates->updates, dates,
 uint->unit
 were'->we're
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -39,5 +39,7 @@ storaget->storage
 subprocess'->subprocess's
 thead->thread
 todays->today's
+udate->update
+udates->updates
 uint->unit
 were'->we're


sebweb3r:
Maybe add date* as suggestions too?

codespell_lib/data/dictionary_code.txt
```diff
@@ -39,5 +39,7 @@ storaget->storage
 subprocess'->subprocess's
 thead->thread
 todays->today's
+udate->update
+udates->updates
```
peternewman, line 43: Yeah as @sebweb3r said:
```suggestion
udate->update, date,
udates->updates, dates,
```

codespell_lib/data/dictionary.txt
```diff
@@ -27667,6 +27667,8 @@ ubitquitous->ubiquitous
 ublisher->publisher
 ubunut->Ubuntu
 ubutunu->Ubuntu
+udated->updated
+udater->updater
```
peternewman, line 27671:
```suggestion
udater->updater, dater,
udating->updating, dating,
```

codespell_lib/data/dictionary.txt
```diff
@@ -27667,6 +27667,8 @@ ubitquitous->ubiquitous
 ublisher->publisher
 ubunut->Ubuntu
 ubutunu->Ubuntu
+udated->updated
```
peternewman, line 27670:
```suggestion
udated->updated, dated,
```

peternewman (changes requested):
A few comments

codespell_lib/data/dictionary_code.txt
cfi-gb: @peternewman @sebweb3r Thanks for your review.

Please could you elaborate why you think that `udate` could be a misspelling of `date`?

The `udate` misspelling of `update` for example is originating from some one (i have done this typo myself already a few times) by not (fully) hitting the `p` button while typing.

But typing `udate` instead of `date` would require to hit one additional button `u` which is not even near the `d` key on the keyboard.

I don't have any strong reasons against not adding these suggestions but would first like to understand the rationale behind these suggestions.

Thanks

sebweb3r: Hey, 

f.e. "leftovers" from the word in front. Date is just a very common word, that is only one (extra) letter away.

I just had a look on [grep.app](https://grep.app/search?current=64&q=udate&words=true) to look if that mistake happened.
It looks like udate itself is a quite common variable name. Thus, udate should go to the code dictionary.

cfi-gb: Thanks for the explanation. As i don't have any strong opinion about that i have submitted the suggestions in 2a9f5c3ecfa70556e03e2a8c45ef9fe3f3374f02

Note that `udate` and `udates` is / was already added to the code dictionary from the beginning.

sebweb3r: Oh. Mea culpa. It seems like I was not yet awake this morning.

peternewman: E.g. "people you date" becomes "people yo udate". The other part of the reasoning is to make sure that automatic mode doesn't leave us with "people yo update" without warning you. In comparison, "comparrison" as I just typed is pretty much guaranteed to be a typo for comparison so wouldn't particularly benefit from other suggestions. I guess broadly the longer the word the greater the chance you can be confident it's only got one logical correction.

peternewman (approved):
Thanks @cfi-gb .

codespell_lib/data/dictionary_code.txt
cfi-gb: @peternewman Thank you very much for taking the time to explain this in such a level of detail. This really explains a lot :+1:

PR #1709, opened 2020-10-23

luzpaz:


lurch:
Should this be rebased / squashed ?

luzpaz:
rebased. ready for merge

codespell_lib/data/dictionary.txt
```diff
@@ -15107,6 +15113,8 @@ informaiton->information
 informatation->information
 informatations->information
 informatikon->information
+informatin->information, informing,
```
lurch, line 15116: _Maybe_ also informant ? :man_shrugging:

luzpaz: @peternewman ?

luzpaz: bump

peternewman: From looking at it, I'd favour informant over informing if it needs an alternative.

Grep.app doesn't seem to find anything but information typos from a reasonable check, did you find some real world examples of the other two?

luzpaz: `informatin` would be a potential non-native speaker error  
As for adding `informant` i haven't seen a usecase for it.

lurch: I've not seen any real-world examples of `informatin` as a typo for `informant`, I was merely suggesting it as a "looks-like" word. I'm happy to defer to @peternewman 's decision.

> informatin would be a potential non-native speaker error for informing

Hmm, perhaps it also worth suggesting `informating->informing,` ? Although https://www.google.com/search?q=informating suggests it's potentially a (niche?) word, so maybe that should be added to the rare dictionary?

larsoner: I think what's here is good enough for now, let's merge and revise if someone hits a use case that causes us to modify it

PR #1717, opened 2020-10-29
Dict: Add plural of comma

sebweb3r:
Hey, 
I'm not sure about this one. Wiktionary proposes commata as plural of comma[1]. cambridge[2] and aspell[3] don't know commata. 

What is the opinion of our natives? @peternewman @lurch 

[1] https://en.wiktionary.org/wiki/commata
[2] https://dictionary.cambridge.org/us/spellcheck/english/?q=commata
[3] http://app.aspell.net/lookup?dict=en_US;words=commata

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5467,6 +5467,7 @@ commansd->commands
 commant->command, comment,
 commanted->commanded, commented,
 commants->commands, comments,
+commata->commas
 commect->connect
 commected->connected
 commecting->connecting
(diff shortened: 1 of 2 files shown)


lurch:
I've _never_ heard of "commata" before, and a quick search finds https://english.stackexchange.com/questions/151257/usage-of-commas-vs-commata
:wink:

sebweb3r:
okay, thats what I found too. :-)

peternewman (changes requested):
Grep.app also finds commatas!

I think this should go in usage or rare though, it's not technically a spelling mistake from the looks of things, just a native speaker wouldn't use it.

sebweb3r:
But commatas would be double plural :smile:

PR #1721, opened 2020-11-03
Dict: Add more corrections

sebweb3r:


sebweb3r:
I've spotted secant, which is a valid word in the code dictionary. Shall I remove it?

codespell_lib/data/dictionary_code.txt
```diff
@@ -24,6 +25,7 @@ nto->not
 objext->object
 od->of
 process'->process's
+protecten->protection
 referer->referrer
 rela->real
 secant->second
```
sebweb3r, line 31: secant:
https://dictionary.cambridge.org/de/worterbuch/englisch/secant

larsoner: Indeed `secant` is real. Maybe move to `rare`?

sebweb3r: Rare is used in default. This would annoy much more people than code.

larsoner: Let's just delete it for now then. It's not even that close a misspelling...

PR #1740, opened 2020-11-12
Add correction for exploit and it's variants.

cfi-gb:
See e.g.

https://grep.app/search?q=expoit&words=true
https://grep.app/search?q=expoits&words=true
https://grep.app/search?q=exploition&words=true
https://grep.app/search?q=exploitions&words=true
https://grep.app/search?q=expoitation&words=true
https://grep.app/search?q=exlpoit&words=true

Not sure if we should suggest a few other alternatives and if the alternatives i have added makes sense. Feedback is welcome.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -11606,6 +11606,9 @@ exlicite->explicit
 exlicitely->explicitly
 exlicitly->explicitly
 exliled->exiled
+exlpoit->exploit
+exlpoited->exploited
+exlpoits->exploits
 exlude->exclude, exude,
 exluded->excluded, exuded,
 exludes->excludes, exudes,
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -12219,6 +12222,8 @@ explitit->explicit
 explitly->explicitly
 explizit->explicit
 explizitly->explicitly
+exploition->exploitation, exploit,
+exploitions->exploitation, exploit,
```
luzpaz, line 12226: My suggestions:
```
exploition->explosion, exploitation, exploit,
exploitions->explosions, exploitations, exploits,
```

cfi-gb: Did a follow-up PR for this suggestions in #1754

PR #1746, opened 2020-11-14
Fix capitalisation of Portuguese to match other occurrences

lurch:


sebweb3r:
There are a lot of words, that should be capital. From what I learned f.e. Asian has always to be capital. 
Same holds for names (f.e. Gaussian)

lurch:
Okay, I've added those suggestions, and I'm happy to add any similar suggestions provided. But with almost 30,000 entries in the dictionary, I don't have time to search through it for "names that ought to be capitalised"! :stuck_out_tongue_winking_eye: 

Interestingly, https://en.wikipedia.org/wiki/Gauss_(unit) writes `gauss` with a lower-case `g` ? :shrug:

sebweb3r:
I forgot the strange rule of the people doing the SI standard :-D

"They are printed in lower-case letters unless they are derived from a proper name, in which case the first letter is a capital letter" [1]

Most other words like Gaussian should be capital.

[1] https://www.bipm.org/utils/common/pdf/si-brochure/SI-Brochure-9.pdf

lurch:
So... does that mean that `gauss` or `Gauss` is correct? :shrug:  (obviously it wouldn't be the first time that Wikipedia is wrong :wink: )

sebweb3r:
It means Gauss is the name and correct 90% of the times (since a lot of stuff is named after him). But if the unit gauss is meant, you should spell it small (even if some people don't do it)

sebweb3r:
The easiest way to process this would propbably multiple merge requests. 
* names
* countries
* calendar stuff

otherwise it will grow big real quick. I can take over some of it, if you want.

lurch:
> I can take over some of it

Please feel free! :smiley:

PR #1748, opened 2020-11-16
Add misspell: classified, classifier, maintainers

vikivivi:


vikivivi:
These words are first entry: maintainable, classifier, classifiers.
Code search: 
https://github.com/search?q=mantainable&type=code
https://github.com/search?q=classifer&type=code

PR #1764, opened 2020-11-20
Move 'commitish->committish' to code dictionary

luzpaz:
This is a programming term and therefore moved to the relevant code dictionary.

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -6,6 +6,7 @@ cas->case, cast,
 clas->class
 cloneable->clonable
 cmo->com
+commitish->committish
 copyable->copiable
 define'd->defined
 dof->of, doff,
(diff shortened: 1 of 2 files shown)


codespell_lib/data/dictionary_code.txt
```diff
@@ -6,6 +6,7 @@ cas->case, cast,
 clas->class
 cloneable->clonable
 cmo->com
+commitish->committish
```
larsoner, line 9: When is `commitish` valid in a computing context? In my quick searches I see `committish` as the preferred spelling, or `commit-ish`, but not `commitish`.

peternewman:
> This is a programming term and therefore moved to the relevant code dictionary.

Yeah just to clarify what @larsoner said @luzpaz , the code dictionary is just for VALID programming terms, which are typos in other contexts, that correct to other words `uint->unit` (for unsigned integer) being the clearest example I can think of. It's not for all programming related typos, so `programing->programming` would still live in the main dictionary. This is specifically a place to "hide" the things that would break a codebase if they were applied as fixes to it (e.g. when correcting typos in the comments).

PR #1794, opened 2020-12-02
acceleartion->acceleration

thughes:
Example: https://crrev.com/c/2566866


peternewman (changes requested):
Looked good, would you mind adding -ed, -e, -or and -ing equivalent typos please?

lurch:
Could also add the -es variant when adding the others too?

peternewman (approved):
Thanks @thughes !

PR #1795, opened 2020-12-02
Add custon->custom

AirQuick:
I made this mistake myself, and it seems I'm not the only one: https://grep.app/search?q=custon&words=true


peternewman (changes requested):
LGTM, do you want to add some of the equivalent custom* words too, e.g. custoner definitely exists as a typo.

Also while looking for that, I found that custoized is a grep.app hit, so perhaps add some of it's friends too?

AirQuick:
Added some. Not sure if too many or few. Let me know.

peternewman:
Thanks @AirQuick . Sorry I got tied up with some other bits.

PR #1798, opened 2020-12-04
Add refinement 'to' to 'te' typo

luzpaz:
Usecase: `He has te do it`


peternewman (approved):
Seemed unlikely but I found at least one hit on grep.app.

PR #1807, opened 2020-12-07
Add privaye->private

fishilico:
This a misspelling found in some projects: https://grep.app/search?q=privaye


larsoner:
FYI you can include more than one word in your PRs

Issue #1805, opened 2020-12-07
Travis CI Credits


peternewman:
https://travis-ci.com/organizations/codespell-project/plan

Mixed messages online, we might be able to ask for credits but may not get them. We probably need to look at moving the remaining bits to GitHub Actions...

larsoner:
I suppose we should fix Travis before merging any other PRs. I can make a quick PR to switch to Azure tomorrow probably.

peternewman:
Turns out switching to GH Actions for everything was pretty easy, see #1810. Curiously none of my Travis repos have had the issue so it might be worth reaching out to them, or we could just give up and move on.

larsoner:
> Curiously none of my Travis repos have had the issue so it might be worth reaching out to them, or we could just give up and move on.

It's because I switched us over to travis-ci.com on a beta about a week before they put these limits in place. I think projects on travis-ci.org still have open-source credits available. So not great timing by me with the switch :(

peternewman:
> It's because I switched us over to travis-ci.com on a beta about a week before they put these limits in place. I think projects on travis-ci.org still have open-source credits available. So not great timing by me with the switch :(

Ah you might just have got unlucky then. I'd moved/created stuff before and after and they all seem to have dodge the issue. So maybe they'll sort codespell-project eventually.

PR #1810, opened 2020-12-08
Run full testing via GH Actions

peternewman:
Works around #1805

### DIFF
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -5,7 +5,9 @@ name: codespell
 on: [push, pull_request]
 jobs:
   test:
-    runs-on: ubuntu-latest
+    env:
+      REQUIRE_ASPELL: true
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         python-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/codespell-private.yml
```diff
@@ -4,6 +4,40 @@
...
+          pip install -e ".[dev]" # install the codespell dev packages
+      - run: python setup.py install
+      - run: codespell --help
+      - run: make check
+      - run: codespell --check-filenames --skip="./.git/*,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*"
+      # this file has an error
+      - run: "! codespell codespell_lib/tests/test_basic.py"
+      - run: codecov
+    
   make-check-dictionaries:
```
peternewman, line 41: I've left this separate so we only comment on any issues once rather than once per language generating a lot of noise.

.github/workflows/codespell-private.yml
```diff
@@ -4,6 +4,40 @@
 name: codespell
 on: [push, pull_request]
 jobs:
+  test:
+    env:
+      REQUIRE_ASPELL: true
+    runs-on: ubuntu-20.04
```
sebweb3r, line 10: Now it is ubuntu-20.04 here and ubuntu-latest below.

sebweb3r: codespell.yml has ubuntu-latest as well.

peternewman: Yeah it's only the version of the aspell dictionary it actually matters for. ubuntu-latest will become ubuntu-20.04 relatively soon, then we can revert this.

sebweb3r: I am just a fan of consistency. ;-)

larsoner:
I just killed the 3.5 entry, I don't think we need it anymore. Will merge once CIs come back happy. Thanks @peternewman !

peternewman:
I was going to update the protection but I see you've done it!

PR #1727, opened 2020-11-05
Add agancy->agency and friends

peternewman:

PR #1815, opened 2020-12-08
Add typos of align, allocate, block, docker, given

bl-ue:
With `-ing` and `-ed` form

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1303,7 +1303,7 @@ alignrigh->alignright
 alikes->alike, likes,
 aline->align, a line, line, saline,
 alined->aligned
-aling->align, along, a line, ailing, sling, 
+aling->align, along, a line, ailing, sling,
 alinged->aligned
 alinging->aligning
 alingment->alignment
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1314,8 +1314,9 @@ alising->aliasing
 aliver->alive, liver, a liver, sliver,
 allcate->allocate
 allcation->allocation
-allcator->allocator
+allcoate->allocate
 allcoated->allocated
+allcoating->allocating
 allcommnads->allcommands
 alle->all, alley,
 alled->called, allied,
@@ -1345,9 +1346,6 @@ alllocation->allocation
 alllow->allow
 alllows->allows
 allmost->almost
-allcoate->allocate
-allcoated->allocated
-allcoating->allocating
 alloacate->allocate
 alloate->allocate, allotted, allot,
 alloated->allocated, allotted,
@@ -3824,13 +3822,13 @@ blockin->blocking
 bloddy->bloody
 blodk->block
 blohted->bloated
-blokcer->blocker
-blokcs->blocks
 blokc->block
-blokcing->blocking
-blokcss->blocks
+blokcer->blocker
 blokchain->blockchain
 blokchains->blockchains
+blokcing->blocking
+blokcs->blocks
+blokcss->blocks
 bloking->blocking
 blong->belong
 blonged->belonged
@@ -7896,17 +7894,6 @@ dcoker->docker
 dcokerd->dockerd
 dcoking->docking
 dcoks->docks
-dockre->docker
-dockred->dockerd
-dokc->dock
-dokced->docked
-dokcer->docker
-dokcerd->dockerd
-dokcing->docking
-dokcre->docker
-dokcred->dockerd
-dokcs->docks
-dcument->document
 dcumented->documented
 dcumenting->documenting
 dcuments->documents
@@ -10133,6 +10120,14 @@ doign->doing
 doiing->doing
 doiuble->double
 doiubled->doubled
+dokc->dock
+dokced->docked
+dokcer->docker
+dokcerd->dockerd
+dokcing->docking
+dokcre->docker
+dokcred->dockerd
+dokcs->docks
 doller->dollar
 dollers->dollars
 dollor->dollar
@@ -13864,8 +13859,8 @@ gitatributes->gitattributes
 gived->given, gave
 giveing->giving
 givem->given, give them, give 'em,
-givven->given
 givveing->giving
+givven->given
 givving->giving
 glamourous->glamorous
 glight->flight
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -7894,6 +7894,7 @@ dcoker->docker
 dcokerd->dockerd
 dcoking->docking
 dcoks->docks
+dcument->document
 dcumented->documented
 dcumenting->documenting
 dcuments->documents
@@ -13856,7 +13857,7 @@ gisers->geysers
 gitar->guitar
 gitars->guitars
 gitatributes->gitattributes
-gived->given, gave
+gived->given, gave,
 giveing->giving
 givem->given, give them, give 'em,
 givveing->giving
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1328,7 +1328,6 @@ allcoaters->allocators
 allcoating->allocating
 allcoation->allocation
 allcoator->allocator
-allcoator->allocator
 allcoators->allocators
 allcommnads->allcommands, all commands,
 alle->all, alley,
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1303,6 +1303,9 @@ alignrigh->alignright
 alikes->alike, likes,
 aline->align, a line, line, saline,
 alined->aligned
+aling->align
+alinged->aligned
+alinging->aligning
 alingment->alignment
 alinment->alignment
 alinments->alignments
@@ -1341,6 +1344,9 @@ alllocation->allocation
 alllow->allow
 alllows->allows
 allmost->almost
+allcoate->allocate
+allcoated->allocated
+allcoating
 alloacate->allocate
 alloate->allocate, allotted, allot,
 alloated->allocated, allotted,
@@ -3818,6 +3824,10 @@ bloddy->bloody
 blodk->block
 blohted->bloated
 blokcer->blocker
+blokcs->blocks
+blokc->block
+blokcing->blocking
+blokcss->blocks
 blokchain->blockchain
 blokchains->blockchains
 bloking->blocking
@@ -7880,6 +7890,9 @@ daty->data, date,
 daugher->daughter
 DCHP->DHCP
 dcoker->docker
+dokcer->docker
+dokcre->docker
+dockre->docker
 dcument->document
 dcumented->documented
 dcumenting->documenting
@@ -13835,9 +13848,12 @@ gisers->geysers
 gitar->guitar
 gitars->guitars
 gitatributes->gitattributes
-gived->given
+gived->given, gave
 giveing->giving
 givem->given, give them, give 'em,
+givven->given
+givveing->giving
+givving->giving
 glamourous->glamorous
 glight->flight
 gloab->globe
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3838,8 +3838,8 @@ blockin->blocking
 bloddy->bloody
 blodk->block
 bloek->bloke
-bloeks->blokes
 bloekes->blokes
+bloeks->blokes
 bloekss->blokes
 blohted->bloated
 blokc->block, bloke,
@@ -10144,10 +10144,10 @@ doiubled->doubled
 dokc->dock
 dokced->docked
 dokcer->docker
-dokcerd->dockerd
+dokcerd->dockerd, docked, docker,
 dokcing->docking
 dokcre->docker
-dokcred->dockerd
+dokcred->dockerd, docked, docker,
 dokcs->docks
 doller->dollar
 dollers->dollars
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1303,10 +1303,11 @@ alignrigh->alignright
 alikes->alike, likes,
 aline->align, a line, line, saline,
 alined->aligned
-aling->align
+aling->align, along, a line, ailing, sling, 
 alinged->aligned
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -1303,6 +1303,9 @@ alignrigh->alignright
 alikes->alike, likes,
 aline->align, a line, line, saline,
 alined->aligned
+aling->align
+alinged->aligned
+alinging->aligning
```
lurch, line 1308: Could also add `alings->aligns` ?

Hmmm, I wonder if e.g. `aling` should also suggest `sling` as a possible correction? And maybe `ailing`? :man_shrugging: Oooh, and `along` is only a single-key different too...

codespell_lib/data/dictionary.txt
```diff
@@ -1341,6 +1344,9 @@ alllocation->allocation
 alllow->allow
 alllows->allows
 allmost->almost
+allcoate->allocate
+allcoated->allocated
+allcoating
```
lurch, line 1349: You missed out the corrected-version here??

codespell_lib/data/dictionary.txt
```diff
@@ -7880,6 +7890,9 @@ daty->data, date,
 daugher->daughter
 DCHP->DHCP
 dcoker->docker
+dokcer->docker
+dokcre->docker
```
lurch, line 7894: Could also add `dcok->dock` and `dokc->dock` ?
And if you feel like it, I guess you could add similar corrections for the `docks`, `docked` and `docking` variants?

bl-ue: @lurch done @ 04eec2b4f5c5198bc97e08a230b096a934dd6dcb

codespell_lib/data/dictionary.txt
bl-ue: @lurch done @ 04eec2b4f5c5198bc97e08a230b096a934dd6dcb

codespell_lib/data/dictionary.txt
bl-ue: @lurch done @ 04eec2b4f5c5198bc97e08a230b096a934dd6dcb

bl-ue:
@lurch done, PTAL.

codespell_lib/data/dictionary.txt
```diff
@@ -7889,10 +7890,22 @@ datsets->datasets
 daty->data, date,
 daugher->daughter
 DCHP->DHCP
+dcok->dock
+dcoked->docked
 dcoker->docker
+dcokerd->dockerd
```
lurch, line 7896: Is dockerd a reference to the Docker software package? I'm not sure it belongs in the "regular" dictionary? ping @larsoner

lurch:
You need to fix the sorting of the entries to make the CI tests happy :wink:

codespell_lib/data/dictionary.txt
bl-ue: Yes @lurch it is a reference to Docker.

bl-ue:
Fixed CI issues. @lurch @larsoner PTAL.

codespell_lib/data/dictionary.txt
```diff
@@ -1303,15 +1303,20 @@ alignrigh->alignright
...
+alinging->aligning
 alingment->alignment
+alings->aligns, slings,
 alinment->alignment
 alinments->alignments
 alising->aliasing
 aliver->alive, liver, a liver, sliver,
 allcate->allocate
 allcation->allocation
-allcator->allocator
```
lurch, line 1313: Probably shouldn't be removing existing corrections? :confused:

lurch:
Apart from the small query about `dockerd` appearing in the main dictionary, this LGTM now :+1:

bl-ue:
@larson https://github.com/codespell-project/codespell/pull/1815#discussion_r539350092

lurch:
(@bl-ue But just to clarify - my query _may_ be totally invalid... I can never remember how all the different dictionaries are supposed to work together :wink: )

bl-ue:
@lurch what about `macosx`:

https://github.com/codespell-project/codespell/blob/d133befb82db8c5b6f491ae0021a9392f60c4a48/codespell_lib/data/dictionary.txt#L17455

and Mozilla:

https://github.com/codespell-project/codespell/blob/d133befb82db8c5b6f491ae0021a9392f60c4a48/codespell_lib/data/dictionary.txt#L17459

which are in the main dictionary?

lurch:
It's hard to imagine a normal word that `maxosx` or `mazilla` would correct to in non-code-usage.
But IMHO it's entirely plausible that `dockerd` might be considered a typo for `docker` or `docked` in regular language usage? :shrug:

bl-ue:
Good point.  So instead I'll add `dockerd` as a possible replacment for `docked`, etc.  WDYT?

lurch:
I've got no absolutely no idea what the "correct" action is in this situation. ping @larsoner @peternewman @sebweb3r

codespell_lib/data/dictionary.txt
sebweb3r:
```suggestion
dcokerd->dockerd, docker,
```

Would probably the easy solution.

sebweb3r:
There is (from my understanding) no better fitting dictionary. 
One solution could be 
`
dcokerd->dockerd, docked, docker,
`
in the main dictionary and
`
dockerd->docker, docked, 
`
in the code dictionary (contains all checks that would be annoying with code)
However, the main dictionary is not allowed to correct to something, that is a mistake 
in the code dictionary (i don't know why).

"Correct" solutions can only be provided by the other two mentioned.

codespell_lib/data/dictionary.txt
peternewman: > Is `dockerd` a reference to the Docker software package? I'm not sure it belongs in the "regular" dictionary? ping @larsoner

It only matters if the typo is in the main dictionary. What we don't want to do is have:
`dockerd->docker`

In the main dictionary, everyone goes and runs that on their codebases as it's enabled by default and suddenly everything breaks.

Having `dokerd->dockerd` and `doker->docker` are both fine though as they won't break anything.

peternewman:
> It's hard to imagine a normal word that `maxosx` or `mazilla` would correct to in non-code-usage.
> But IMHO it's entirely plausible that `dockerd` might be considered a typo for `docker` or `docked` in regular language usage?

Again it doesn't matter; ALL we care about is that the typo isn't a valid word in some context, then it gets shunted out to another dictionary (see also rare e.g. `chancel->chance`, running that across a religious document would cause chaos, they aren't currently a big enough market so we just pop it in rare and hopefully people are careful or only turn on rare if they are). We wouldn't put e.g. macos->macros in the main dictionary as it would make a mess, it would go in code so only people not writing "code" would get that correction with hopefully no unintended consequences.

> There is (from my understanding) no better fitting dictionary.
> One solution could be
> `dcokerd->dockerd, docked, docker,`
> in the main dictionary and
> `dockerd->docker, docked, `
> in the code dictionary (contains all checks that would be annoying with code)

I think the only downside of this is we don't currently chain dictionaries (see also e.g. #1672), although as long as there is more than one correction in the main one (i,e. it duplicates the corrections from either side of the code one) it shouldn't have any negative side effects.

> However, the main dictionary is not allowed to correct to something, that is a mistake
> in the code dictionary (i don't know why).

We just need to add this commit back in (or make an equivalent edit):
https://github.com/codespell-project/codespell/pull/1727/commits/fc203b40ce40440eeefbe288456f0224b140cfec

The list of correcting to a typo in another dictionary is kept as small as practically possible because you shouldn't generally be correcting to a word that is itself a typo in another dictionary, instead you should correct directly to the final correction. The exceptions are allowed where required. The obvious case being colur->colour in the main dictionary and colour->color in GB to US is obviously acceptable and intended.

If anyone can word this in a more friendly way and put it in a readme would be great as it and similar variations do seem to keep coming up. Or better descriptions for the various dictionaries. Or do we just need a note about it generally only being the typo side of the dictionary we care about (apart from GB to US).

codespell_lib/data/dictionary.txt
```diff
@@ -1303,15 +1303,21 @@ alignrigh->alignright
...
 alinments->alignments
 alising->aliasing
 aliver->alive, liver, a liver, sliver,
 allcate->allocate
 allcation->allocation
 allcator->allocator
+allcoate->allocate
 allcoated->allocated
+allcoating->allocating
 allcommnads->allcommands
```
peternewman, line 1321: Not yours, but can this also suggest `all commands`

codespell_lib/data/dictionary.txt
```diff
@@ -3817,9 +3823,13 @@ blockin->blocking
 bloddy->bloody
 blodk->block
 blohted->bloated
+blokc->block
```
peternewman, line 3826: Suggest bloke too.

codespell_lib/data/dictionary.txt
```diff
@@ -3817,9 +3823,13 @@ blockin->blocking
...
 blokcer->blocker
 blokchain->blockchain
 blokchains->blockchains
+blokcing->blocking
+blokcs->blocks
```
peternewman, line 3831: Suggest blokes too.

codespell_lib/data/dictionary.txt
```diff
@@ -3817,9 +3823,13 @@ blockin->blocking
...
+blokcss->blocks
```
peternewman, line 3832: Probably again.

codespell_lib/data/dictionary.txt
```diff
@@ -13835,9 +13858,12 @@ gisers->geysers
 gitar->guitar
 gitars->guitars
 gitatributes->gitattributes
-gived->given
+gived->given, gave,
```
peternewman, line 13861: Nice :+1:

peternewman (changes requested):
A few more comments

codespell_lib/data/dictionary.txt
bl-ue: > In the main dictionary, everyone goes and runs that on their codebases as it's enabled by default and suddenly everything breaks.

Very good point.

bl-ue:
@lurch @sebweb3r @peternewman done @ 307fe9e922ae219185d38923a4f613bbe9c3039d, PTAL.

codespell_lib/data/dictionary.txt
```diff
@@ -10107,6 +10141,14 @@ doign->doing
 doiing->doing
 doiuble->double
 doiubled->doubled
+dokc->dock
+dokced->docked
+dokcer->docker
+dokcerd->dockerd
+dokcing->docking
+dokcre->docker
+dokcred->dockerd
```
lurch, line 10150: Should this (and `dokcerd`) have the same `->dockerd, docked, docker,` correction as `dcokerd` ?

codespell_lib/data/dictionary.txt
```diff
@@ -3816,10 +3837,18 @@ blockhains->blockchains
 blockin->blocking
 bloddy->bloody
 blodk->block
+bloek->bloke
+bloeks->blokes
+bloekes->blokes
```
lurch, line 3842: Need to fix sorting. (this keeps biting me in the bum whenever I submit a PR too :laughing: )

bl-ue: I was sure it was sorted! 😄

bl-ue: Well @lurch if we could just change the order of the lines with special symbols (like `'` and `;`), we could use tools like `sort`, or VS Code extensions to sort the lines.

lurch: https://github.com/codespell-project/codespell#sending-pull-requests

bl-ue: The issue there is mainly `'` in contractions getting mixed up with `->`:

```
aircrafts->aircraft
aircrafts'->aircraft's
```

becomes

```
aircrafts'->aircraft's
aircrafts->aircraft
```

this is because `'` (apostrophe) has char code 39, and `-` (hyphen) has char code 45, so `aircrafts'` is moved before `aircrafts`. 

Do we even have sorting rules related to such cases?  I guess `aircrafts'` is supposed to be after `aircraft` because of the extra character?

codespell_lib/data/dictionary.txt
```diff
@@ -1303,16 +1303,35 @@ alignrigh->alignright
...
 alising->aliasing
 aliver->alive, liver, a liver, sliver,
 allcate->allocate
+allcateing->allocating
+allcater->allocator
+allcaters->allocators
+allcating->allocating
 allcation->allocation
 allcator->allocator
+allcator->allocator
```
lurch, line 1322: CI is complaining that `allcator` appears twice :wink:

bl-ue: oops.....thank you

bl-ue: https://github.com/codespell-project/codespell/pull/1815/commits/4677b714757cf9d98dbbdef6a3d2f358634f9a3c

codespell_lib/data/dictionary.txt
```diff
@@ -1303,16 +1303,34 @@ alignrigh->alignright
...
 allcoated->allocated
-allcommnads->allcommands
+allcoateing->allocating
+allcoateng->allocating
+allcoater->allocator
+allcoaters->allocators
+allcoating->allocating
+allcoation->allocation
+allcoator->allocator
+allcoator->allocator
```
lurch, line 1331: `allcoator` is duplicated too :confused:

bl-ue: I almost checked for more duplicates! 😢 

Thank you @lurch.  This should be all.  This has been a very involved PR! 😄  Thank you for all of your support!

bl-ue: Next time I open a PR I'll work locally so I can run `make check` and avoid all of this trouble!! 🤣

codespell_lib/data/dictionary.txt
bl-ue: @lurch I didn't see that link you sent.  I guess I should've read that section... 😄

bl-ue:
@peternewman review?

peternewman (approved):
LGTM thanks @bl-ue

PR #1816, opened 2020-12-09
Add misspell: implicit. Not a noun with plural

vikivivi:
* Add misspell: implicit
* Typical code search:
  * https://github.com/search?q=implicits&type=code
  * https://grep.app/search?q=implicits&words=true

Not a word in hunspell, aspell, ispell and major online dictionaries.
But it was found in many projects especially Scala.
According to Word Hippo, "The word implicit is not a noun with a plural"

Do you suggest to withdraw PR if it impacted Scala language?


vikivivi:
I have decided to close and withdraw this PR. Thank you.

PR #1817, opened 2020-12-09
Add misspell: ciphertext, ciphersuite

vikivivi:
* Add misspell: ciphertext, ciphertexts
* New entry word: ciphertexts
* Typical code search:
  * https://github.com/search?q=cyphertext&type=code
  * https://github.com/search?q=cyphertexts&type=code

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5036,6 +5036,8 @@ ciontrol->control
 ciper->cipher
 cipers->ciphers
 cipherntext->ciphertext
+ciphersuit->ciphersuite
+ciphersuits->ciphersuites
 ciphersute->ciphersuite
 ciphersutes->ciphersuites
 cipheruite->ciphersuite
@@ -7780,6 +7782,8 @@ cylnder->cylinder
 cylnders->cylinders
 cylynders->cylinders
 cymk->CMYK
+cyphersuite->ciphersuite
+cyphersuites->ciphersuites
 cyphertext->ciphertext
 cyphertexts->ciphertexts
 cyprt->crypt
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5048,7 +5048,7 @@ cipersuites->ciphersuites
 cipertext->ciphertext
 cipertexts->ciphertexts
 ciph->cipher, chip,
-ciphe->ciphe
+ciphe->cipher
 cipherntext->ciphertext
 ciphersuit->ciphersuite
 ciphersuits->ciphersuites
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5041,6 +5041,9 @@ cintaner->container
 ciontrol->control
 ciper->cipher
 cipers->ciphers
+cipersuite->ciphersuite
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -7776,6 +7776,8 @@ cylnder->cylinder
 cylnders->cylinders
 cylynders->cylinders
 cymk->CMYK
+cyphertext->ciphertext
+cyphertexts->ciphertexts
```
lurch, line 7780: I was _going_ to suggest that you could add `cypher->cipher` too, but it looks like cypher is an accepted (archaic?) variant spelling https://www.google.com/search?q=cypher+or+cipher (although not one I've ever seen myself) :wink: 
It seems `cyphertext` is definitely not a valid spelling though :+1: 

I wonder if it might by worth doing `cyphertext->ciphertext, hypertext,` ? (or perhaps that's "too different")

vikivivi: > I was going to suggest that you could add cypher->cipher too, but it looks like cypher is an accepted (archaic?)

I preferred not to add `cypher` as misspell. Both `cipher` and `cypher` are the same word and same valid noun.
See https://www.oxfordlearnersdictionaries.com/definition/english/cypher?q=cypher

Although `cypher` is a valid word, but `cyphertext` is not a valid word in major online dictionaries (Cambridge, Collins, Merriam-Webster, Oxford Learners, dictionary.com). Except Wikipedia redirect `Cyphertext` to `Ciphertext`.

> I wonder if it might be worth doing 'cyphertext->ciphertext, hypertext,' ?

It would take 2 misspell mistakes to arrive with `hypertext`. IMO, almost all code developers who wrote `cyphertext` actually mean `ciphertext`. The percentage of those who mean `hypertext` would be negligible. I would add `hypertext` if I get more votes.

lurch:
> Although cypher is a valid word, but cyphertext is not a valid word

Yeah, that's exactly what I was saying, sorry if I didn't explain myself very well.

> It would take 2 typo mistakes to arrive with hypertext. IMO, almost all code developers who wrote cyphertext actually mean ciphertext.

Yes I admit it's a bit of a stretch, feel free to ignore my suggestion.

vikivivi:
Very common with software developers writing `cipher` and `suite` as a combined word.
See `ubuntu$ grep -r ciphersuite /usr/include/openssl/`
* Typical code search:
  * https://github.com/search?q=ciphersutes&type=code
  * https://github.com/search?q=cipheruites&type=code

codespell_lib/data/dictionary.txt
```diff
@@ -5036,6 +5036,10 @@ ciontrol->control
 ciper->cipher
 cipers->ciphers
 cipherntext->ciphertext
+ciphersute->ciphersuite
+ciphersutes->ciphersuites
```
lurch, line 5040: Could maybe also add
```
ciphersuit->ciphersuite
ciphersuits->ciphersuites
```
?

And perhaps it's also worth adding
```
cyphersuite->ciphersuite
cyphersuites->ciphersuites
```
:man_shrugging:

vikivivi: Done.

peternewman (approved):
LGTM

PR #1823, opened 2020-12-11
Dict: Unneeded

sebweb3r:
I'm not sure about this one. 

Cambridge doesn't accept unneeded
https://dictionary.cambridge.org/spellcheck/english/?q=unneeded
However aspell does
http://app.aspell.net/lookup?dict=en_US;words=unneeded

Whats the opinion of our natives @lurch @peternewman

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -28879,7 +28879,7 @@ uneccesary->unnecessary
 uneccessarily->unnecessarily
 unecessarily->unnecessarily
 unecessary->unnecessary
-uneeded->unneeded, needed,
+uneeded->unnecessary, needed,
 uneforceable->unenforceable
 uneform->uniform
 unencrpt->unencrypt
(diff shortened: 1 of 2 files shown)


lurch:
It does sound a bit odd. I think I'd suggest `unneeded->unnecessary, not needed,` ?

peternewman:
The GB large aspell has it too:
http://app.aspell.net/lookup?dict=en_GB-large;words=unneeded

It's in Collins too:
https://www.collinsdictionary.com/dictionary/english/unneeded

So I think it's a usage/style issue at best.

sebweb3r:
So how to proceed?
Put it in informal or usage? Or close the PR?

codespell_lib/data/dictionary.txt
```diff
@@ -28879,7 +28879,7 @@ uneccesary->unnecessary
 uneccessarily->unnecessarily
 unecessarily->unnecessarily
 unecessary->unnecessary
-uneeded->unneeded, needed,
```
peternewman, line 28882: Suggest unheeded too

peternewman (commented):
Another suggestion for the original

peternewman:
> So how to proceed?
> Put it in informal or usage? Or close the PR?

Up to you, informal would be the right choice if at all I think, usage was the wrong one. I'm not convinced how informal it is personally.

I did just find one improvement to the existing entries though...

sebweb3r:
I will put it in my local dictionary. I created a new MR for your suggestion to not flood the git history.

peternewman:
> I will put it in my local dictionary. I created a new MR for your suggestion to not flood the git history.

Fair  enough, thanks for that.

I didn't realise a PR could close another!

PR #1834, opened 2020-12-23
Add a couple of spelling corrections

gumb0:


codespell_lib/data/dictionary.txt
```diff
@@ -28873,6 +28874,7 @@ undeflow->underflow
 undeflows->underflows
 undefuned->undefined
 undependend->independent, nondependent,
+underfiend->undefined
```
peternewman, line 28877: Nice, it would make sense to add underfiend and also derfine(d) if they don't exist.

peternewman (changes requested):

codespell_lib/data/dictionary.txt
gumb0: Added.

peternewman (approved):
LGTM thanks

PR #1835, opened 2020-12-23
Add several spelling corrections + refinements

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -25642,6 +25656,10 @@ skeep->skip
 skelton->skeleton
 skept->skipped
 sketchs->sketches
+sketck->skate, sketch,
+sketcked->skated, sketched,
+sketckes->skates, sketches,
+sketcking->skating, sketching,
```
peternewman, line 25662: Swap these round as the second term is clearly the most likely one.

peternewman: In fact I can't really see how the first term is likely in any of these cases...

peternewman (changes requested):

codespell_lib/data/dictionary.txt
luzpaz: I'll swap them but i think we should keep both. I think I saw the original typo somewhere where the correct word was `sketck`.

luzpaz: 79a4879

peternewman: As in `sketck->skate`?

codespell_lib/data/dictionary.txt
```diff
@@ -24637,6 +24650,7 @@ rmove->remove
 rmoved->removed
 rmoving->removing
 rnage->rage, range,
+ro->to, row,
```
peternewman, line 24653: Also rob, rod, roe and rot

codespell_lib/data/dictionary.txt
luzpaz: yes. that is what I meant

peternewman: Okay fair enough.

codespell_lib/data/dictionary.txt
bl-ue:
```suggestion
ro->to, row, rob, rod, roe, rot,
```

bl-ue (approved):

PR #1836, opened 2020-12-25
Add refinement

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -19765,7 +19765,7 @@ osffset->offset
 osffsets->offsets
 osffsetting->offsetting
 osicllations->oscillations
-ot->to, of, or,
+ot->to, of, or, not,
```
arm-in, line 19768: possibly also capital "OT" for internet slang "off topic"

peternewman: Although I think it will upset the corrects to itself test...

bl-ue: Yeah, we can't do that 🤔

bl-ue (approved):

PR #1855, opened 2021-01-15
Add doamin->domain

fishilico:
This is a misspelling found in some projects: https://grep.app/search?q=doamin

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -10074,7 +10074,8 @@ dnamically->dynamically
 dne->done
 do'nt->don't
 doalog->dialog
-doamin->domain
+doamin->domain, dopamine,
+doamine->dopamine
 doamins->domains
 doas->does
 doble->double
@@ -26726,6 +26727,8 @@ subdirectorys->subdirectories
 subdirecty->subdirectory
 subdivisio->subdivision
 subdivisiond->subdivisioned
+subdoamin->subdomain
+subdoamins->subdomains
 subelemet->subelement
 subelemets->subelements
 subexperesion->subexpression
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -10075,7 +10075,7 @@ dne->done
 do'nt->don't
 doalog->dialog
 doamin->domain, dopamine,
-doamine->dopamine
+doamine->dopamine, domain,
 doamins->domains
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -10074,6 +10074,8 @@ dnamically->dynamically
 dne->done
 do'nt->don't
 doalog->dialog
+doamin->domain
```
peternewman, line 10077: Suggest dopamine too

codespell_lib/data/dictionary.txt
```diff
@@ -10074,6 +10074,8 @@ dnamically->dynamically
...
+doamins->domains
```
peternewman, line 10078: Perhaps add the subdomain(s) variants of these too (if they don't already exist).

peternewman (changes requested):
A few suggestions

codespell_lib/data/dictionary.txt
fishilico: I added `dopamine` suggestion for both `doamin` and `doamine`

codespell_lib/data/dictionary.txt
fishilico: This is indeed a common mispelling (https://grep.app/search?q=subdoamin), which was not present. Added, thanks!

codespell_lib/data/dictionary.txt
peternewman: Thanks, I suspect the former is less likely, but it can't hurt.

peternewman: Ignore that, I get what you mean now.

codespell_lib/data/dictionary.txt
```diff
@@ -10074,6 +10074,9 @@ dnamically->dynamically
 dne->done
 do'nt->don't
 doalog->dialog
+doamin->domain, dopamine,
+doamine->dopamine
```
peternewman, line 10078: It's probably worth suggesting domain here too as a secondary one I'd have thought.

peternewman (changes requested):
Just one outstanding comment

codespell_lib/data/dictionary.txt
fishilico: I agree

peternewman (approved):
Thanks @fishilico

PR #1857, opened 2021-01-18
Add spelling corrections for render and variants.

cfi-gb:
See e.g.:

https://github.com/search?q=rendiring&type=code (has more hits then grep.app)
https://grep.app/search?q=rendiring

https://github.com/search?q=rendired&type=code (grep.app doesn't have any hits for this one)

https://grep.app/search?q=rendir&words=true

I have put "rendir" and "rendirs" into the code dictionary because it could also be an variable name for "rename dir" or similar.

Not absolutely sure if all checks / corrections makes sense, any feedback is welcome.


peternewman (approved):
LGTM

PR #1858, opened 2021-01-19
Add arch-independet and related typos

peternewman:


bl-ue (approved):
`databaes` would be cool to add too — https://github.com/search?q=databaes

peternewman:
> `databaes` would be cool to add too — https://github.com/search?q=databaes

Good shout @bl-ue . Do you want to open a PR for that, given it's not really related to the bits in here (and that will probably get it in sooner).

bl-ue:
Yeap I'll open a PR soon for that 👍🏻

PR #1861, opened 2021-01-21
Dict: Add specifi*

sebweb3r:
https://grep.app/search?q=specifig&words=true

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25973,6 +25973,7 @@ specifclly->specifically
 specifed->specified
 speciffic->specific
 speciffically->specifically
+specifi->specific, specify,
 specifially->specifically
 specificallly->specifically
 specificaly->specifically
@@ -25994,6 +25995,8 @@ specifiction->specification
 specifictions->specifications
 specificy->specify, specificity, specifically,
 specifid->specified
+specifig->specific
+specifigation->specification
 specifing->specifying
 specifiy->specify
 specifiying->specifying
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -25994,6 +25995,8 @@ specifiction->specification
 specifictions->specifications
 specificy->specify, specificity, specifically,
 specifid->specified
+specifig->specific
```
peternewman, line 25998: Suggest specifying too maybe or do we feel that's too much of a stretch?

peternewman (commented):
Just one thought

codespell_lib/data/dictionary.txt
sebweb3r: I looks a bit like a stretch. But I don't mind adding it, if you would like to have it in.

peternewman: I guess it's a bit phonetic too maybe?

bl-ue: I think it's too much of a stretch too — let's just leave it as-is for right now. wdyt @peternewman?

codespell_lib/data/dictionary.txt
```diff
@@ -25994,6 +25995,8 @@ specifiction->specification
...
+specifigation->specification
```
bl-ue, line 25999:
```suggestion
specifigation->specification
specifigations->specifications
```

bl-ue (approved):

PR #1869, opened 2021-01-30
en-GB to en-US dict: add licenced->licensed

luzpaz:


codespell_lib/data/dictionary_en-GB_to_en-US.txt
```diff
@@ -121,6 +121,7 @@ legalised->legalized
 legalises->legalizes
 legalising->legalizing
 licence->license
+licenced->licensed
```
peternewman, line 124: I think this should be licensed->licensed i.e. a No-Op

luzpaz: you just wrote `licensed` twice :point_up:

peternewman: Yes, I think for a quick check that the British English is `licensed` too, unless you can point me to a source I've missed. I did try and double-check online...

peternewman: Sorry, I probably didn't explain very well, what I meant is that the British word is also licensed, so as you say @vikivivi it should be removed because it's a no-op. I think our tests should catch that anyway.

bl-ue:
So what's the status of this PR?  `licenced` is definitely British-only.  @peternewman can we merge? I see no blockers 🤔

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Cannot merge because `licenced` is missing from the _aspell_ directory.

Added to the https://github.com/codespell-project/codespell/issues/2657#issuecomment-1365202259 list instead...

PR #1877, opened 2021-02-08
Add exitst -> exits/exists

cfi-gb:
See e.g. https://grep.app/search?q=exitst&words=true

```
# password is null. Return 1 if the user $user\@$domain exitst.
```

Note that i have added it to the `code` dictionary because it seems that it is commonly used as an abbreviation for `exit status`. If you disagree with that please let me know.


peternewman (approved):
LGTM and code seems like a safe precaution.

PR #1878, opened 2021-02-08
Add spawining->spawning

cfi-gb:
See https://github.com/search?q=spawining&type=code (Note that https://grep.app/search?q=spawining doesn't have that much hits)

Not sure about other variants like `re(-)?spawining`, please let me know if i should add them.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -22684,8 +22684,7 @@ re-negoziations->re-negotiations
 re-negoziator->re-negotiator
 re-negoziators->re-negotiators
 re-realease->re-release
-re-spawining->respawning
-re-spawning->respawning
+re-spawining->re-spawning, respawning,
 re-uplad->re-upload
 re-upladad->re-upload, re-uploaded,
 re-upladed->re-uploaded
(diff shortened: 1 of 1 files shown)


peternewman:
> Not sure about other variants like `re(-)?spawining`, please let me know if i should add them.

It certainly won't hurt to have them in, the more the merrier!

peternewman (commented):
My comment should have been a comment review really.

cfi-gb:
> My comment should have been a comment review really.

Thanks a lot for the review. I have added a few variants, hope these are correct.

codespell_lib/data/dictionary.txt
```diff
@@ -22684,6 +22684,8 @@ re-negoziations->re-negotiations
 re-negoziator->re-negotiator
 re-negoziators->re-negotiators
 re-realease->re-release
+re-spawining->respawning
+re-spawning->respawning
```
peternewman, line 22688: It looks like re-spawning is valid too, so perhaps suggest both.

cfi-gb: Ah, indeed. Updated in ea13c9205bf06e0682bf0298bff80a9b8c8d6a49, thanks a lot.

peternewman (approved):

Issue #1824, opened 2020-12-11
Case sensitive typos


bl-ue:
Is it possible to have typos that only match a certain case?  I'm thinking of the following scenario: I would like to add `Github->GitHub`, but that typo has a specific case, because if it is all lowercase or all uppercase, the correction will not be applicable.

peternewman:
Unfortunately this is probably overly simplistic. Consider this (following https://docs.microsoft.com/en-us/windows/win32/stg/coding-style-conventions):
`pGithub = Github`

With this change this will become:
`pGitHub = GitHub`

Which we probably don't want!

Technically currently a constant called say `GITTHUB` might currently get corrected to `github` or `GitHub` if we chained dictionaries, which isn't ideal but at least we've fixed the spelling in the constant.

I'm not sure there's actually an easy fix for this in non-interactive mode, apart from just flagging it as a potential capitalisation issue. Obviously we can offer it in interactive mode.

_Adapted from what I originally posted in https://github.com/codespell-project/codespell/pull/1880#discussion_r579392070_

PR #1880, opened 2021-02-16
Support for capitalization

cristiklein:
I'm sending this PR to brainstorm how support for #1824 could look like. I hereby propose adding '!' in front of words where capitalization matters.

Feedback appreciated!

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -61,8 +61,8 @@ _builtin_dictionaries = (
         None, None, None, None,),
     ('en-GB_to_en-US', 'for corrections from en-GB to en-US', '_en-GB_to_en-US',  # noqa: E501
         True, True, ('en_GB',), ('en_US',)),
-    ('capitalization', 'for using uppercase/lowercase properly', '_capitalization',
-        None, None, None, None,),
+    ('capitalization', 'for using uppercase/lowercase properly',
+        '_capitalization', None, None, None, None,),
 )
 _builtin_default = 'clear,rare'
 
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -61,6 +61,8 @@ _builtin_dictionaries = (
         None, None, None, None,),
     ('en-GB_to_en-US', 'for corrections from en-GB to en-US', '_en-GB_to_en-US',  # noqa: E501
         True, True, ('en_GB',), ('en_US',)),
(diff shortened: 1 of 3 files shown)


codespell_lib/_codespell.py
```diff
@@ -429,10 +431,7 @@ def build_dict(filename, misspellings, ignore_words):
     with codecs.open(filename, mode='r', encoding='utf-8') as f:
         for line in f:
             [key, data] = line.split('->')
-            # TODO for now, convert both to lower. Someday we can maybe add
-            # support for fixing caps.
             key = key.lower()
-            data = data.lower()
```
peternewman, line 435: Unfortunately this is overly simplistic. Consider:
`m_host_ip = ip`

With this change this will become:
`m_host_IP = IP`

Which we probably don't want!

Technically currently a constant called say `GITTHUB` might currently get corrected to `github`, which isn't ideal but at least we've fixed the constant.

I'm not sure there's actually an easy fix for this in non-interactive mode, apart from just flagging it as a potential capitalisation issue. Obviously we can offer it in interactive mode.

codespell_lib/data/dictionary_capitalization.txt
```diff
@@ -0,0 +1,4 @@
+aws->!AWS
+github->!GitHub
+ip->!IP
+tcp->!TCP
```
peternewman, line 4: For some others you could search the dictionary for where the RHS is in all caps, e.g. CMYK.

peternewman (changes requested):
I don't think we particularly need ! unless people want it in their own custom dictionary. We could just add another field where we allow something to replace to itself (but with different case).

Also prefixing with ! would preclude potential corrections which actually began with an exclamation mark.

codespell_lib/_codespell.py
cristiklein: Wait? This project also does spellchecking in source code, i.e., function names and variables? I was operating under the assumption that it's mainly aimed at comments.

peternewman: Yeah hence the name @cristiklein . :smile: 

Broadly by default it's not clever enough to differentiate, although it may be possible to hook it into existing parsers to handle that sort of thing.

For example if you search here for sic you'll see most of the external facing/API changes that have been caught (primarily) by codespell (it's also caught a load of internal ones which aren't mentioned):
https://github.com/OpenLightingProject/ola/blob/master/NEWS

PR #1881, opened 2021-02-24
Add additional spelling correction for considere.

cfi-gb:
`considere` might be also a spelling mistake of `consider`, see e.g. https://github.com/greenbone/gvm-libs/pull/450

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -6705,7 +6705,7 @@ considderation->consideration
 considdered->considered
 considdering->considering
 considerd->considered
-considere->considered
+considere->considered, consider,
 consideren->considered
 consideres->considered, considers,
 considert->considered, consider,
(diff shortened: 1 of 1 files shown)


peternewman (approved):
Would you ~considere~ consider just checking grep.app/GitHub search to see which is most likely and putting that first, or have you already ~considere~ considered that?

peternewman:
`consider` just wins on the first five pages counting the English typos on grep.app.

codespell_lib/data/dictionary.txt
```diff
@@ -6705,7 +6705,7 @@ considderation->consideration
 considdered->considered
 considdering->considering
 considerd->considered
-considere->considered
+considere->considered, consider,
```
cfi-gb, line 6708: As suggested in https://github.com/codespell-project/codespell/pull/1881#issuecomment-787955872:

```suggestion
considere->consider, considered,
```

cfi-gb:
> `consider` just wins on the first five pages counting the English typos on grep.app.

Thanks a lot for checking. I have changed the order in 26972d10a9f7f30828ff3d9ebe5c164dc6386334, having `consider` as the first suggestion sounds better and it is also the first suggestion of my current Firefox dictionary.

PR #1890, opened 2021-03-15
Add spelling correction for errot(s)->error(s).

cfi-gb:
The `t` key is next to the `r` key so this might be a reason for a common typo like this.

See e.g.:

https://grep.app/search?q=errot&words=true
https://grep.app/search?q=errots&words=true

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -11336,6 +11336,8 @@ errornously->erroneously
 errorprone->error-prone
 errorr->error
 erros->errors
+errot->error
+errots->errors
 errro->error
 errror->error
 errrors->errors


peternewman (approved):

peternewman:
Curiously etror and the other variants don't have any hits on grep.app!

cfi-gb:
> Curiously etror and the other variants don't have any hits on grep.app!

From time to time i have seen grep.app lacking a lot results in comparison to the plain github search, see e.g. #1878

peternewman:
> From time to time i have seen grep.app lacking a lot results in comparison to the plain github search, see e.g. #1878

Ah yeah true. In which case do you fancy opening one to add ertor and etror then? From a bit of hunting I didn't find any relevant hits for ettor though (it seems to be a genuine word in a Nordic language, which perhaps floods the results).

cfi-gb:
> > From time to time i have seen grep.app lacking a lot results in comparison to the plain github search, see e.g. #1878
> 
> Ah yeah true. In which case do you fancy opening one to add ertor and etror then? From a bit of hunting I didn't find any relevant hits for ettor though (it seems to be a genuine word in a Nordic language, which perhaps floods the results).

Sure, here we go: #1907

PR #1896, opened 2021-03-22
Add finishied->finished.

cfi-gb:
See e.g.:

https://grep.app/search?q=finishied (https://github.com/search?q=finishied&type=code has quite a lot more hits)

Not sure if we should suggest "finish" as well?

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -13135,6 +13135,7 @@ finilizes->finalizes
 finisch->finish, Finnish,
 finisched->finished
 finishe->finished, finish,
+finishied->finished
 finishs->finishes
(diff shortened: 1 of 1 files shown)


peternewman (approved):

peternewman:
> Not sure if we should suggest "finish" as well?

finishied->finish seems unlikely, if that's what you mean?

cfi-gb:
> > Not sure if we should suggest "finish" as well?
> 
> finishied->finish seems unlikely, if that's what you mean?

Ah, yes and sorry. I should be more explicit on this. Thought about that one:

```
finishied->finished, finish,
```

but after having another look that would be wrong and i think the current MR is fine.

PR #1902, opened 2021-03-30
Remove crate from the list of words

sylvestre:
I propose that we remove crate from codespell.
Two reasons:
* it is an english word: https://en.wikipedia.org/wiki/Crate
* it is the name of a library in Rust: https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html

One of my project
```
$ codespell src|grep crate|wc -l
519
```
It had to be excluded from Firefox codespell integration:
https://searchfox.org/mozilla-central/source/tools/lint/spell/exclude-list.txt#10

### DIFF
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -31,7 +31,6 @@ copping->coping, copying, cropping,
 covert->convert
 crasher->crash
 crashers->crashes
-crate->create
 crated->created
 creche->crèche
 cristal->crystal


peternewman:
>     * it is an english word: https://en.wikipedia.org/wiki/Crate

All of the words in the rare dictionary are English words, hence it's the rare dictionary, they are just (supposed to be) rarer than what they correct to.

>     * it is the name of a library in Rust: https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html

That does mean it gets more hits than most.

> It had to be excluded from Firefox codespell integration:
> https://searchfox.org/mozilla-central/source/tools/lint/spell/exclude-list.txt#10

That means the system works!

Rust has pretty much flooded the market, with it's term and the idea of crates, but after a bit of searching I did find this example (going the other way):
https://github.com/Deph0/CrateDB-scripts/blame/5a1d914303dbadf153f76ba72caba9a8f19af3e1/README.md#L24-L25

Compare the two highlighted lines, so people do make that typo!

So personally I think we should keep it. The rare dictionary is disabled by default.

I did ponder a while ago if the rare dictionary should be split up more:
https://github.com/codespell-project/codespell/issues/1469

But words from there will always have issues for some people, e.g. chancel appears in OpenStreetMap

waldyrious:
>  The rare dictionary is disabled by default.

Hmm, interesting. I don't recall enabling the rare dictionary explicitly, but crate does get reported for Rust projects if I run `codespell` on them 🤔. What gives?

peternewman:
> > The rare dictionary is disabled by default.
> 
> What gives?

Oops, what gives is I lied:
https://github.com/codespell-project/codespell/blob/079b649d9ba381c0cc7a62f5f88d043e5d81c5ea/codespell_lib/_codespell.py#L65

Although I've still had to ignore stuff for various reasons in every sizeable code base I've run it against:
https://github.com/OpenLightingProject/ola/blob/29e38800e59b13bcee2c63e30605c77b0b0d4812/.codespellignorewords
https://github.com/openstreetmap/iD/blob/2e885699142f2e508fdc2339005e5fb3a58bdc42/.github/workflows/codespell.yml#L16
https://github.com/OpenLightingProject/rdm-app/blob/2874fe7bf9550687b131062113089fa9eadcf9ba/.travis-ci.sh#L57

So while removing crate would help some rust users, it would hinder all the create typo makers.

bl-ue:
With the given discussion and reasonings, I'll close those this PR.  We can reopen if anyone sees fit.

PR #1903, opened 2021-03-31
Add focued->focused

AirQuick:
https://grep.app/search?q=focued&words=true


peternewman (approved):
Looks good, I'm wondering if we should add focu->focus too?

AirQuick:
> I'm wondering if we should add focu->focus too?

Added it in 25a2394b1fd3f675e972816f896a60a61e0f519a.

peternewman (approved):

PR #1904, opened 2021-03-31
Remove heterogenous->heterogeneous

waldyrious:
"heterogenous" is both a valid word with a distinct meaning, and an accepted alternative spelling of "heterogeneous": https://en.wiktionary.org/wiki/heterogenous.

In place of this correction, this PR adds hetrogeneous->heterogeneous.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -14558,7 +14558,7 @@ hertzs->hertz
 hesiate->hesitate
 hesistant->hesitant
 hestiate->hesitate
-hetrogenous->heterogeneous
+hetrogenous->heterogenous, heterogeneous,
 heuristc->heuristic
 heuristcs->heuristics
 heursitics->heuristics
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -14558,7 +14558,6 @@ hertzs->hertz
 hesiate->hesitate
 hesistant->hesitant
 hestiate->hesitate
-heterogenous->heterogeneous
 hetrogenous->heterogeneous
 heuristc->heuristic
 heuristcs->heuristics
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -14558,6 +14558,7 @@ hertzs->hertz
 hesiate->hesitate
 hesistant->hesitant
 hestiate->hesitate
+hetrogeneous->heterogeneous
 hetrogenous->heterogenous, heterogeneous,
 heuristc->heuristic
 heuristcs->heuristics
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -63,6 +63,7 @@ guarantied->guaranteed
 guerilla->guerrilla
 guerillas->guerrillas
 hart->heart, harm,
+heterogenous->heterogeneous
 hided->hidden, hid,
 hist->heist, his,
 hove->have, hover, love,


sebweb3r:
I'm against it. Homogenous seems to have valid, but pretty rare usage in biology (maybe rare dictionary). Homogenous is just a pretty, pretty common misspelling.

bl-ue:
@sebweb3r this PR is about `heterogenous`, not `homogenous` :)

peternewman:
This seems to agree and say they are different words:
https://radiopaedia.org/articles/heterogeneous-vs-heterogenous?lang=gb

waldyrious:
@peternewman btw, should `hetrogenous->heterogeneous` also be replaced by `hetrogenous->heterogenous`?

peternewman:
> @peternewman btw, should `hetrogenous->heterogeneous` also be replaced by `hetrogenous->heterogenous`?

I wonder if it should suggest both? :shrug:

waldyrious:
Yeah, I thought the same. I'll add both for now, and we can change later if there's a decision otherwise.

codespell_lib/data/dictionary.txt
```diff
@@ -14558,8 +14558,8 @@ hertzs->hertz
 hesiate->hesitate
 hesistant->hesitant
 hestiate->hesitate
-heterogenous->heterogeneous
-hetrogenous->heterogeneous
+hetrogeneous->heterogeneous
```
peternewman, line 14561: Maybe this should suggest the other one too, with *no* second?

peternewman:
If you're going to do some force pushing etc, it would probably make sense to make this two commits. One to remove `heterogenous->heterogeneous` as they are both valid words. Then a second separate commit to add the new typo/improve/extend the existing ones.

waldyrious:
Yeah, makes sense :) I'll split this into two commits as soon as we agree on what approach to use for the inline discussion.

codespell_lib/data/dictionary.txt
waldyrious: I'm not sure that's warranted. My impression is that "heterogeneous" is significantly more common than "heterogeneous", and the data seems to corroborate it ([google ngram viewer](https://books.google.com/ngrams/graph?content=heterogenous%2Cheterogeneous); [4M](https://www.google.com/search?q=heterogenous) vs [111M](https://www.google.com/search?q=heterogeneous) search results). I'd even venture that "heterogenous" might belong in the rare dictionary.

peternewman: I'm just always a bit wary about it automatically fixing stuff; if you've made a typo and there is another plausible word, you could just has easily have made more of/a different typo!

peternewman: > I'd even venture that "heterogenous" might belong in the rare dictionary.

The rare dictionary is for real words that may be typos of other words, e.g. bellow->below, so we could add this to rare:
`heterogenous->heterogeneous`

I think it allows three way corrections (i.e. having typos for them individually at the same time).

waldyrious: Agreed. So just to make sure we're aligned, what you're suggesting is:

- add `heterogenous->heterogeneous` to the rare dictionary
- _not_ change `hetrogeneous->heterogeneous` into `hetrogeneous->heterogeneous,heterogenous,` in this line

Is that right?

peternewman: > Agreed. So just to make sure we're aligned, what you're suggesting is:
> 
> * add `heterogenous->heterogeneous` to the rare dictionary

Yes, assuming it doesn't get confused by the resultant chain of hetrogeneous->heterogenous->heterogeneous

> * _not_ change `hetrogeneous->heterogeneous` into `hetrogeneous->heterogeneous,heterogenous,` in this line

I think (well I think that I think), that we should suggest both and not automatically convert hetrogeneous to heterogeneous. Because hetrogenEous to hetErogenous is only the move of one E a few places.

waldyrious: > Because hetrogenEous to hetErogenous is only the move of one E a few places.

Hmm, not sure I agree with that reasoning. That's a very mechanistic way to look at the issue — other than accidentally swapping the order of a pair of letters, it's not common for a human to move a letter several places across a word by accident. Semantically-wise, I would say those correspond to different, independent typos: "hetro-" instead of "hetero-", and "-genous" instead of "-geneous". 

With that in mind, IMO it's quite likely that one would (1) commit both typos simultaneously, and (2) actually mean the much rarer "-genous" suffix as opposed to "-geneous". I'd say catering to that case is being excessively cautious and making the tool more cumbersome to the vast majority of users.

peternewman: Okay, fair enough. I'm not hugely heavily wedded to it for something this niche.

waldyrious: Ok, I've updated the PR.

waldyrious:
@peternewman the PR has been updated with split commits — please check if how they're divided makes sense to you.

peternewman (approved):
LGTM, thanks for the work on this one @waldyrious , it's making me go a bit cross-eyed looking at the options!

PR #1906, opened 2021-04-02
Add variants of octothorpe (the '#' character)

bl-ue:
https://en.wiktionary.org/wiki/octothorpe


codespell_lib/data/dictionary.txt
```diff
@@ -19550,6 +19550,28 @@ octect->octet
 octohedra->octahedra
 octohedral->octahedral
 octohedron->octahedron
+octothhorpes->octothorpes
+octothorp->octothorpe
+octothorpe->octothorpe
```
peternewman, line 19555: This one just corrects to itself!

peternewman (changes requested):
Are these actual or generated?

codespell_lib/data/dictionary.txt
```diff
@@ -19550,6 +19550,28 @@ octect->octet
...
+octothrpes->octothorpes
+octotohrpe->octothorpe
+octotohrpes->octothorpes
+octotorpe->octothorpe
+octotorpes->octothorpes
+octotrhope->octothorpe
+octotrhopes->octothorpes
+octotthhorpe->octothorpe
+octotthhorpes->octothorpes
+octotthorpes->octothorpe
```
peternewman, line 19573: This should probably be removed!

bl-ue:
>Are these actual or generated?

Both, in a way. 🙂  I manually typed some likely typos as this is rather long word, and then I used VS Code multi-cursor to mess them up a bit more.

>This should probably be removed!

Yeah, they're rather silly I admit.  There were actually more before I opened PR but I trimmed some down... 😄

codespell_lib/data/dictionary.txt
bl-ue, line 19555:
```suggestion
```

codespell_lib/data/dictionary.txt
bl-ue, line 19573:
```suggestion
```

bl-ue, line 19573:
```suggestion
```

codespell_lib/data/dictionary.txt
```diff
@@ -19550,6 +19550,24 @@ octect->octet
 octohedra->octahedra
 octohedral->octahedral
 octohedron->octahedron
+octothhorpes->octothorpes
```
peternewman, line 19553: We should at least have the singular

peternewman:
> > Are these actual or generated?
> 
> Both, in a way. slightly_smiling_face I manually typed some likely typos as this is rather long word, and then I used VS Code multi-cursor to mess them up a bit more.

I'm not sure we should necessarily be flooding the dictionary with random potential typos, it would be nice to see some real world examples that some of these exist as a base, even if we then add the plural etc for example.

bl-ue:
Okay, I'll keep the simple obvious ones and get rid of the less likely ones.  How about that?

peternewman:
> Okay, I'll keep the simple obvious ones and get rid of the less likely ones. How about that?

In case my :+1: wasn't explicit enough, that sounds good to me.

peternewman:
What happened @bl-ue ? I thought you were going to just sort the simple/obvious ones?

bl-ue:
Since I opened this PR, I realized that it's better to provide known typos, rather than possible ones.  None of these I've ever actually seen, so I figured it's just a waste of time.  There's much more interesting work that can be done in terms of adding to the database :wink:

PR #1908, opened 2021-04-14
Add correction for bypass and variants.

cfi-gb:
See e.g.

https://grep.app/search?q=bypas&words=true
https://grep.app/search?q=bypasing
https://grep.app/search?q=bypased

Not 100% sure about the `bypas` as it is used a few times as a variable name but IMHO this is a too common typo for the code dictionary.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -4282,6 +4282,9 @@ buuild->build
 buuilds->builds
 bve->be
 bwtween->between
+bypas->bypass
+bypased->bypassed
+bypasing->bypassing
 byteoder->byte order
 cacahe->cache
 cacahes->caches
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -4282,6 +4282,9 @@ buuild->build
 buuilds->builds
 bve->be
 bwtween->between
+bypas->bypass
+bypased->bypassed
+bypasing->bypassing
 byteoder->byte order
```
peternewman, line 4288: Sorry  I know it's not your change, but can this be changed as follows:
```suggestion
byteoder->byteorder, byte order,
```

Otherwise it feels like it's quite risky to code.

peternewman (commented):
LGTM, just one (unrelated) change.

codespell_lib/data/dictionary.txt
cfi-gb: Sure :+1:

peternewman (approved):
Thanks for picking that unrelated change up.

PR #1920, opened 2021-05-19
Add spelling correction for authenticate and variants.

cfi-gb:
See e.g.:

https://grep.app/search?q=authentciation

I have also added a few variants of existing typos (there currently should be 7 for each authenticate* variant), hope i was able to identify and update all correctly.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -2932,7 +2932,13 @@ attruibutes->attributes
 atttached->attached
 atttribute->attribute
 atttributes->attributes
+atuhenticate->authenticate
+atuhenticated->authenticated
+atuhenticates->authenticates
+atuhenticating->authenticating
 atuhentication->authentication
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -2961,6 +2970,8 @@ autenticated->authenticated
 autenticates->authenticates
 autenticating->authenticating
 autentication->authentication
+autenticatior->authenticatior
```
bl-ue, line 2973: Is `authenticatior` a realy word? Do you mean `authenticator` (minus the last `i`)?

cfi-gb: Ah, yes. I completely messed this up, thanks a lot for the review / catching this :+1: 

Fixed in the recent push.

bl-ue (approved):

bl-ue (approved):

PR #1921, opened 2021-05-19
Add a couple of refinements to preexisting entries

luzpaz:


bl-ue (commented):
Good idea.  I've now seen a lot of misspellings because people try to write what they hear. 👍🏻

bl-ue (approved):

bl-ue (approved):

PR #1931, opened 2021-06-02
Add 79 (completely real!) variants of "vehicle"

bl-ue:
It can be a hard word to write!


bl-ue:
Ping @peternewman.  What do you think about this PR?  I'd really like to know as it will guide me for similar PRs in the future.

bl-ue:
Some time I'll go through these an pick only the ones that are (1) real, and (2) likely to appear in real source code.

PR #1933, opened 2021-06-04
Refine preexisting entry

luzpaz:


bl-ue (approved):

codespell_lib/data/dictionary.txt
```diff
@@ -7082,7 +7082,7 @@ contrl->control
 contrller->controller
 contro->control
 controlable->controllable
-controle->control, controlled, controller,
+controle->control, controlled, controls, controller,
```
bl-ue, line 7085: I think it's more likely that the user would mean `controls` than `controlled` — what do you think?
```suggestion
controle->control, controls, controlled, controller,
```

bl-ue (approved):

PR #1937, opened 2021-06-07
Add parsef->parsed

luzpaz:


bl-ue (approved):

luzpaz:
I guess I can also add `parsec` as an option as well since it's one keystroke removed from 'd'

PR #1939, opened 2021-06-09
Add more corrections for variants of promt

jrosdahl:


codespell_lib/data/dictionary.txt
```diff
@@ -22063,6 +22063,8 @@ promotted->promoted
 promprted->prompted
 promps->prompts
 promt->prompt
+promted->prompted
+promting->prompting
```
bl-ue, line 22067: May I suggest `promter->prompter` as well?
```suggestion
promted->prompted
promter->prompter
promting->prompting
```

bl-ue (changes requested):

bl-ue (approved):

bl-ue (approved):

PR #1942, opened 2021-06-10
Remove cancelability from the dictionary

enjoy-binbin:
cancelability and cancellability both are correct spellings. 

Remove it from the dictionary
And add it to dictionary_rare.

We can see many uses here: https://linux.die.net/man/3/pthread_setcanceltype

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -4478,7 +4478,6 @@ camparing->comparing
 can;t->can't
 canadan->canadian
 canbe->can be
-cancelability->cancellability
 cancelaltion->cancellation
 cancelation->cancellation
 cancelations->cancellations
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -16,6 +16,7 @@ busses->buses
 calculatable->calculable
 calender->calendar
 calenders->calendars
+cancelability->cancellability
 cant->can't
 chack->check, chalk, cheque,
 chancel->cancel


bl-ue:
@enjoy-binbin just because it's used in a real, large (and old) project doesn't mean necessarily that it's not a misspelling.  For example, the V8 JavaScript engine (which powers Google Chrome and Node.js) has several C++ types defined which have `Convertable` in their names, which is a misspelling.  The correct spelling is `Convertible`. https://github.com/v8/v8/search?q=convertable

enjoy-binbin:
I am not really sure about that. So i am going to get outside help...

Pinging... @yoav-steinberg

bl-ue:
I think you're right actually @enjoy-binbin: https://en.wiktionary.org/wiki/cancellability (lists https://en.wiktionary.org/wiki/cancelability as an alternate spelling).

bl-ue (approved):

peternewman:
> I think you're right actually @enjoy-binbin: https://en.wiktionary.org/wiki/cancellability (lists https://en.wiktionary.org/wiki/cancelability as an alternate spelling).

It would be nice to see it in a commercial dictionary which probably spends more time and effort vetting it's entries.

I wonder if it should be rare rather than dropped entirely?

bl-ue:
@enjoy-binbin will you please remove it from `dictionary.txt` and add it instead to https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary_rare.txt?

enjoy-binbin:
@bl-ue Sure. Already updated. Please check again for me.
ps sorry for the delay

bl-ue (approved):
LGTM 👍🏻

Issue #1949, opened 2021-06-14
Rework of #1374


bl-ue:
I've just finished my rework of #1374 — it's quite large, adding 2,449 new typo.  I'm going to close #1374 and open 25 new PRs—one for each letter of the alphabet except for X for which there's no typos—instead.

- A → #1950
- B → #1951
- C → #1952
- D → #1953
- E → #1954
- F → #1955
- G → #1956
- H → #1957
- I → #1958
- J → #1959
- K → #1960
- L → #1961
- M → #1962
- N → #1963
- O → #1964
- P → #1965
- Q → #1966
- R → #1967
- S → #1968
- T → #1969
- U → #1970
- V → #1971
- W → #1972
- Y → #1973
- Z → #1974

bl-ue:
ping @luzpaz @c72578 @arm-in @warmsocks @waldyrious @Gelma @anatol @lfelipe @codespell-project (@tldr-pages?) can you help review the load? 😆

bl-ue:
Okay, all 25 PRs' checks are green 💚

peternewman:
Can't we close this and just track it in the PRs?

bl-ue:
Yep, sure!

warmsocks:
I noticed that the title of the original, #1374 is okay, but in all the individual, by-letter PRs, the word "misspell" is misspelled as "mispell".  In this case, that is a functional breakage, since it is in a URI, and as written, when copied and pasted, results in 404/not found. 

Suggest correcting the PR titles.

warmsocks:
> ping @luzpaz @c72578 @arm-in @warmsocks @waldyrious @Gelma @anatol @lfelipe @codespell-project (@tldr-pages?) can you help review the load?

@bl-ue Sure. I'm working through them as time permits.  If I manage to get through them all, I'll post here.

bl-ue:
Haha, that's a bit embarassing 😳 :D
I'll fix those now.

PR #1374, opened 2019-12-10
Batch import from marvin.cs.uidaho.edu/misspell.html

Gelma:
Stolen from

http://marvin.cs.uidaho.edu/misspell.html

PR #1962, opened 2021-06-14
Typos from marvin.cs.uidaho.edu/misspell.html (rework of #1374): M

bl-ue:
For #1949


skangas:
I believe this can be closed now that #2221 has been merged.

PR #1965, opened 2021-06-14
Typos from marvin.cs.uidaho.edu/misspell.html (rework of #1374): P

bl-ue:
For #1949

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -20721,6 +20721,7 @@ paceholder->placeholder
 pach->patch, path,
 pachage->package
 paches->patches
+pachooly->patchouli
 pacht->patch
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -21400,24 +21542,73 @@ phylosophical->philosophical
...
 phython->python
 phyton->python
 phy_interace->phy_interface
 piblisher->publisher
 pich->pitch
+picknic->picnic
+pickniced->picnicked
+picknicer->picnicker
+picknicing->picnicking
+picknick->picnic
```
skangas, line 21565: picknick is an accepted variant spelling according to https://www.merriam-webster.com/dictionary/picknick

I'm not sure what we prefer to do in such cases.

PR #1969, opened 2021-06-14
Typos from marvin.cs.uidaho.edu/mispell.html (rework of #1374): T

bl-ue:


codespell_lib/data/dictionary.txt
```diff
@@ -29649,7 +29657,12 @@ trempoline->trampoline
 treshhold->threshold
 treshold->threshold
 tressle->trestle
+tresure->treasure
+tresures->treasures
 treting->treating
+trew->true
```
skangas, line 29663: `trew->true, threw,`

PR #1970, opened 2021-06-14
Typos from marvin.cs.uidaho.edu/misspell.html (rework of #1374): U

bl-ue:
For #1949


codespell_lib/data/dictionary.txt
```diff
@@ -30194,6 +30207,8 @@ unhilighted->unhighlighted
 unhilights->unhighlights
 Unicde->Unicode
 unich->unix
+unick->eunuch, unix,
+unicks->eunuchs
```
luzpaz, line 30211: `unicks->eunuchs, unix,` is appropriate

PR #1975, opened 2021-06-14
Typos from salsa.debian.org/lintian/lintian

bl-ue:


vikivivi:
comment-only: The github debian/lintian code is stale. The latest lintian code for `correction` (last commit 2 months ago) is available from https://salsa.debian.org/lintian/lintian/-/blob/master/data/spelling/corrections

bl-ue:
Nice catch @vikivivi, I didn't even think to check for a newer version.  Imported typos from there (albeit there weren't many).

PR #1980, opened 2021-06-16
Allow multiple lines in config files

wjsi:
As it is common to use multiple lines in `setup.cfg`, esp. when some paths are long. It would be nice to support config files like

```ini
[codespell]
skip =
    bad1.txt
    bad2.txt  # this is a comment
count =
```

When parsed by `configparser`, multiline values are returned as `\n` separated strings. What we need is to strip comments and concatenate these lines together.


codespell_lib/_codespell.py
```diff
@@ -413,8 +413,17 @@ def parse_options(args):
...
             # Add option as arg.
             cfg_args.append("--%s" % key)
+            # get lines of value
+            val_lines = config['codespell'][key].splitlines()
+            # strip comments from lines
+            filtered_lines = []
+            for line in val_lines:
+                line = line.split('#', 1)[0].strip()
+                if line != "":
+                    filtered_lines.append(line)
```
bl-ue, line 423: This manual solution concerns me a bit.  Stdlib functions always account for more possible errors and other things going wrong.  For example, given the following:
```ini
[codespell]
skip = "##d.txt##"
```
`val` below will be `"`, the result of manually stripping `"##d.txt##"` of the first `#` and everything after it.

wjsi: I decide to remove support for inline comments as `configparser` in Python stops handling these comments by default, and other libraries (like `coveragepy`) do not process these comments either.

.gitignore
```diff
@@ -1,4 +1,5 @@
 .coverage
+.hypothesis
```
DimitriPapadopoulos, line 2: This is unrelated, best left for a different merge request.

codespell_lib/_codespell.py
```diff
@@ -413,8 +413,15 @@ def parse_options(args):
         for key in config['codespell']:
             # Add option as arg.
             cfg_args.append("--%s" % key)
+            # strip comments from lines
+            filtered_lines = []
+            for line in config['codespell'][key].splitlines():
+                line = line.strip()
+                if line != "":
+                    filtered_lines.append(line)
```
DimitriPapadopoulos, line 421:
```python
# strip whitespace from lines
filtered_lines = filter(None, line.strip() for line in config['codespell'][key].splitlines())
```
I don't understand `strip comments from lines`. This code does not strip comments, `configparser` does. This code strips whitespace.

codespell_lib/_codespell.py
```diff
@@ -413,8 +413,15 @@ def parse_options(args):
...
+            # concatenate non-empty lines
+            val = ','.join(filtered_lines)
```
DimitriPapadopoulos, line 423: This isn't a mere concatenation. Lines are separated by `,`.

This works with `skip`, will it work with `dictionary`?

PR #1981, opened 2021-06-16
Make all dictionary entries lowercase

bl-ue:
This PR is a proposal to, for consistency, make all dictionary entries (typo, suggestions, and hints) lowercase.

I'm doing this for two reasons:
1. because I've observed that proper capitalization of various proper nouns is hard enforce everywhere, and all over the dict there are messed up capitalizations.

2. There is **no** point to having them capitalized, because `build_dict()` converts them to lowercase before it uses them:
   https://github.com/codespell-project/codespell/blob/c6293a665b64d7d19e31b907a28335619a0b4923/codespell_lib/_codespell.py#L458-L461

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -558,10 +558,10 @@ achor->anchor
 achored->anchored
 achoring->anchoring
 achors->anchors
-ACI->ACPI
+aci->acpi
 acident->accident
 acient->ancient
-ACII->ASCII
+acii->ascii
 acition->action
 acitions->actions
 acitivate->activate
@@ -981,7 +981,7 @@ agaisnt->against
 agaist->against
 agancies->agencies
 agancy->agency
-aganda->agenda, Uganda,
+aganda->agenda, uganda,
 aganist->against
 agant->agent
 agants->agents, against,
@@ -1073,7 +1073,7 @@ airlfow->airflow
 airloom->heirloom
 airporta->airports
 airrcraft->aircraft
-aisian->Asian
+aisian->asian
 aixs->axis
 aizmuth->azimuth
 ajacence->adjacence
@@ -1656,7 +1656,7 @@ amdgput->amdgpu
 amendement->amendment
 amendmant->amendment
 amened->amended, amend,
-Amercia->America
+amercia->america
 amerliorate->ameliorate
 amgle->angle
 amgles->angles
@@ -1765,7 +1765,7 @@ andoid->android
 androgenous->androgynous
 androgeny->androgyny
 androidextra->androidextras
-androind->Android
+androind->android
 andthe->and the
 ane->and
 anevironment->environment
@@ -1938,7 +1938,7 @@ apender->appender
 apendices->appendices
 apending->appending
 apendix->appendix
-apenines->Apennines
+apenines->apennines
 aperatures->apertures
 aperure->aperture
 aperures->apertures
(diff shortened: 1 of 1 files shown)


vikivivi:
Please do not remove capitalise suggestion been build in the dictionaries. See:
https://www.grammarly.com/blog/capitalization-countries-nationalities-languages/
https://www.scribendi.com/academy/articles/capitalization.en.html

In my opinion, the capitalise suggestion in dictionary.txt is supposed for codespell to suggest in capitalised manner in-according to the dictionary. But currently it does not seem to work as expected on my system to suggest in capitalised manner.

Expectation example:
```sh
$ echo amercia | codespell -
amercia ==> America
$ echo wedensday | codespell -
wedensday ==> Wednesday
```

bl-ue:
Yeah, that's because of the lines I linked above — the capitalisation in the dictionaries serves _no_ purpose.

bl-ue:
Ideally, we could make it support capitalisation as @larsoner's todo says.  I'll try it one of these days.

peternewman (changes requested):
Straight reject, no changes possible.

From a technical perspective you've only done one of the dictionaries, unless there is no capitalisation in all the others, which seems unlikely.

"consistency" is not a sufficient reason to undo all the effort that has been built into the dictionary so far for this future feature. You also don't know if other people aren't using the codespell dictionaries in other tools and making use of this functionality.

peternewman:
> Yeah, that's because of the lines I linked above — the capitalisation in the dictionaries serves _no_ purpose.

No, those 282 are the groundwork for the implementation of that TODO, there's no reason not to get the dictionary into shape before the code can use it. Yes the remainder of the 32,000 entries may need reviewing and capitalisation adding, but the existing content would make the functionality immediately useful and means it's just a coding problem, not a language one too.

codespell_lib/data/dictionary.txt
```diff
@@ -18797,6 +18797,7 @@ mofification->modification
 mofified->modified
 mofifies->modifies
 mofify->modify
+mohammedan->muslim
```
peternewman, line 18800: You should add this in a separate PR though.

codespell_lib/data/dictionary.txt
```diff
@@ -17423,7 +17423,7 @@ laod->load
 laoded->loaded
 laoding->loading
 laods->loads
-Laotion->Laotian
```
peternewman, line 17426: Suggest lotion too

peternewman (changes requested):
I couldn't work out why the additions and removals didn't match...

bl-ue:
😆 I thought it probably wasn't a very good idea, but I thought I'd get more feedback by opening a PR with some real changes than an issue.  I agree, we should fix the code to respect the capitalisation, which is something that I'm working on.

bl-ue:
>I couldn't work out why the additions and removals didn't match...

Yeah, I meant for that line to go in another PR 🙂

Thank you for the input @peternewman — I'll add your suggestions in a new PR I'm going to open.

peternewman:
> laughing I thought it probably wasn't a very good idea, but I thought I'd get more feedback by opening a PR with some real changes than an issue. I agree, we should fix the code to respect the capitalisation, which is something that I'm working on.

If it's integrated with the function that currently tries to guess the case of the suggestion should be half way there, but it might need an option to ignore the capitalisation anyway (or offer both suggestions in interactive mode if it's mixed case), we probably need to play around with a few scenarios.

PR #1985, opened 2021-06-17
Add ore->or to dictionary_rare.txt

cfi-gb:
There is a `ORE` library which seems to be commonly used. `ore` is also a valid word but also might be a spelling mistake of `or`, hope the placement in dictionary_code.txt is correct to handle such edge cases.

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -26,7 +26,6 @@ mut->must, mutt, moot,
 nto->not
 objext->object
 od->of
-ore->or
 process'->process's
 protecten->protection
 reday->ready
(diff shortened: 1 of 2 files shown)


bl-ue (approved):

peternewman (commented):
So as cfi-gb noted, ore is a valid word, so it shouldn't/can't go in the main dictionary, which means it needs to go in rare or elsewhere.

Personally I'd probably lean towards rare for the same reasons as #1902, most people won't be using the ORE library, so would benefit from not miss-spelling `or`. The other stuff in the code dictionary are programming specific terms which are typos in other cases e.g. stdio whereas this feels more like crate, if I'm writing something that's not code, I can't turn on the code dictionary and be guaranteed safe if things like this are in there.

cfi-gb:
Thanks for the reviews and the suggestion to move it into the rare dictionary. Sounds sensible so moved it in 19664a73537de21623ec76ef28edb9c23722a71b

peternewman (approved):
LGTM

PR #1991, opened 2021-06-21
dictionary: fix suggested words with capitalisation

vikivivi:


luzpaz:
Thanks for the effort. Though honestly I'm not a big fan of this idea. The reason is that a lot of source code comments are meant to be less intrusive. Forcing caps might defeat this purpose.

peternewman:
> Edit: I like the idea of leaving the decision up to the original authors. This is what codespell currently does. If they capitalized it, then it corrects it with the capitalized typo fix, if not, then correct it without. Basically, it corrects it the way it found it.

I think there was an issue for this as it does a upper-case first match too, which didn't work with two word corrections or something.

IIRC the system currently ignores the case of the stuff in the dictionary, but I've pushed for changes similar to what @vikivivi has done when reviewing stuff, working on the principle if it's right in the dictionary we can add an option to enable that behaviour in future. If it's not right in the dictionary the option can't do anything. So I'm in favour of this change (and it matches some other ones already present).

vikivivi:
@luzpaz In case you are not aware, this PR is also related to another PR https://github.com/codespell-project/codespell/pull/1981 which wanted to convert all entries in dictionary to lowercase and it was rejected. I fully agreed that it is important to get the dictionary in shape as reference. Whether the codespell engine is capable of using the dictionary to the max is another TODO effort. Also see my comment on capitalization rules https://github.com/codespell-project/codespell/pull/1981#issuecomment-862388808

vikivivi:
@peternewman for your approval

codespell_lib/data/dictionary.txt
```diff
@@ -18459,13 +18459,13 @@ maxinum->maximum
 maxium->maximum
 maxiumum->maximum
 maxmimum->maximum
-maxosx->macosx
+maxosx->macOS X
```
peternewman, line 18462: We need to offer the correction without spaces first, or you'll break a lot of code related stuff:
```suggestion
maxosx->maxosx, macOS X,
```

You could probably put this on it's own in the code dictionary:
`maxosx->macOS X`

codespell_lib/data/dictionary.txt
```diff
@@ -30791,7 +30791,7 @@ unhilight->unhighlight
 unhilighted->unhighlighted
 unhilights->unhighlights
 Unicde->Unicode
-unich->unix
+unich->Unix
```
peternewman, line 30794: Possibly suggest eunuch too?

codespell_lib/data/dictionary.txt
```diff
@@ -30855,7 +30855,7 @@ uniqe->unique
 uniqu->unique
 uniquness->uniqueness
 uniterrupted->uninterrupted
-UnitesStates->UnitedStates
+UnitesStates->United States
```
peternewman, line 30858: Again:
```suggestion
UnitesStates->UnitedStates, United States,
```

peternewman (changes requested):
Just as a sanity check, have you confirmed this doesn't currently force corrections of things to the correct case?

peternewman:
Thinking about this, the code fix should be fairly easy I think, we've got the existing logic where it suggests lower/UPPER/Title as appropriate, I think it could just add a secondary suggestion, either if the typo was lower/Title (or perhaps if the dictionary typo and correction have different case), where it would offer the fixed version, so for example:
asci would suggest ascii and ASCII
fance would suggest france and France
Fance might just suggest France

I think this would result in a good balance of offering the fix without breaking your code.

vikivivi:
@peternewman 
> We need to offer the correction without spaces first, or you'll break a lot of code related stuff:

Could you explain how it will break a lot of code related stuff?
Breaking 3rd party using codespell dictionary or codespell code?
My demonstration show below. 
```sh
host:~/tmp$ wget https://raw.githubusercontent.com/vikivivi/codespell/patch22/codespell_lib/data/dictionary.txt
host:~/tmp$ mkdir -p test ; cd test
host:~/tmp/test$ grep -i "Amercia\|UnitesStates" ../dictionary.txt
Amercia->America
UnitesStates->United States

host:~/tmp/test$ echo "The United States of Amercia, commonly known as the UnitesStates or Amercia." > docu.txt
host:~/tmp/test$ codespell -D ../dictionary.txt -w -i 0
FIXED: ./docu.txt
host:~/tmp/test$ cat ./docu.txt 
The United States of America, commonly known as the united states or America.
```

```text
maxosx->maxosx, macOS X,
```
I am confuse by your comment on this `maxosx`, why suggestion is the same as misspell? Why in both dictionary.txt and dictionary_code.txt?

peternewman:
> Could you explain how it will break a lot of code related stuff? Breaking 3rd party using codespell dictionary or codespell code? My demonstration shown below.

Let it fix this code and then see how it runs...
```
var UnitesStates = 50
print "There are " . $UnitesStates . " states";
```

> 
> ```
> maxosx->maxosx, macOS X,
> ```
> 
> I am confuse by your comment on this `maxosx`, why suggestion is the same as misspell? Why in both dictionary.txt and dictionary_code.txt?

Sorry I've fixed the typo in my suggestion :facepalm: 

I meant ma**c**osx as used as a compiler flag amongst others...

PR #2006, opened 2021-07-05
Add mane->main, many to rare dictionary

luzpaz:


gpotter2:
What if my program has something to do with horse manes? 🐴

PR #2019, opened 2021-07-17
Add fixes from https://github.com/ggtracker/sc2reader/pull/146

timgates42:


luzpaz (approved):
LGTM, though i'm not sure about the commit title. Should we keep or modify?

peternewman (approved):
LGTM. It tells us where they come from which works for me.

TBH the commit titles are somewhat irrelevant here IMHO, git blame (where it works) is probably just as useful).

I guess ideally the title would be more descriptive with the link in the description, but :shrug:

PR #2040, opened 2021-08-11
README: add --ignore-words-list

kolyshkin:
It took me a while to figure out how to specify a list of ignored words
in a configuration file, as --ignore-words-list was missing from README
(I eventually found it elsewhere).

* add --ignore-words-list long option next to -L;
* add ignore-words-list to the .codespellrc example.

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>


README.rst
```diff
@@ -43,15 +43,15 @@ Some noteworthy flags::
 
 The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
 
-    codespell -I FILE, --ignore-words=FILE
+    codespell -I, --ignore-words FILE
```
peternewman, line 46: I assume the original style with an equals worked here? I'd probably favour reverting the change of this line so it still matches grep's man page.

README.rst
```diff
@@ -43,15 +43,15 @@ Some noteworthy flags::
...
 
-The ``-I`` flag can be used for a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said list of allowed words. **Important note:** The list passed to ``-I`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
+The ``-I`` flag can be used to supply a file with a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said list of allowed words. **Important note:** The list passed to ``-I`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
```
peternewman, line 48: :+1:

peternewman (changes requested):
Thanks for this.

Regarding the options with arguments, looking at the grep man page for example, it looks like this:
`-e PATTERN, --regexp=PATTERN`

It would seem sense to me to follow the same style so would you mind updating them to match

README.rst
kolyshkin: Redone to adhere to grep(3) syntax.

kolyshkin:
@peternewman PTAL

README.rst
```diff
@@ -45,15 +45,15 @@ The ``-w`` flag will actually implement the changes recommended by codespell. No
 
     codespell -I FILE, --ignore-words=FILE
 
-The ``-I`` flag can be used for a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said list of allowed words. **Important note:** The list passed to ``-I`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
+Get a list of words to ignore from _FILE_. The _FILE_ format is one word per line. **Important note:** Words are case-sensitive, based on how they are listed in the codespell dictionaries. ::
 
-    codespell -L word1,word2,word3,word4
+    codespell -L LIST, --ignore-words-list=LIST
```
peternewman, line 50: Maybe turn this into an example like the skip one?

README.rst
```diff
@@ -96,12 +96,13 @@ be specified in this file (without the preceding dashes), for example::
 
     [codespell]
     skip = *.po,*.ts,./src/3rdParty,./src/Test
+    ignore-words-list = creat,clos
```
peternewman, line 99: :+1: Great idea!

peternewman (approved):
I like this new more succinct wording, but I've asked @luzpaz for a review too as they had tweaked this in the past.

README.rst
kolyshkin: Since you mentioned grep(3) man page, I tried to follow its style, which mentions the argument twice (this is why I originally changed it to `-o, --option example,argument` -- to avoid repeating the example argument).

I'm afraid something like
```
codespell -L word1,word2,..., --ignore-words-list=word1,word2,...
```
will look way worse. Or do you mean something different here?

peternewman: Yeah I think I'd agree. I was meaning more like this (rather than dropping the example config entirely):

https://github.com/codespell-project/codespell/blob/eabccd736fb56fb3dc5cc27c219f70f967892a22/README.rst#L58

kolyshkin: done

README.rst
```diff
@@ -45,15 +45,19 @@ The ``-w`` flag will actually implement the changes recommended by codespell. No
...
+Get a list of words to ignore from ``FILE``. The ``FILE`` format is one word per line. **Important note:** Words are case-sensitive, based on how they are listed in the codespell dictionaries. ::
 
-    codespell -L word1,word2,word3,word4
+    codespell -L LIST, --ignore-words-list=LIST
 
-The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it.  **Important note:** The list passed to ``-L`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
+Ignore words from the comma-separated ``LIST``. **Important note:** Words are case-sensitive, based on how they are listed in the codespell dictionaries. Example:
 
-    codespell -S, --skip=
+* to ignore the words "fle" and "clos", invoke ``codespell --ignore-words-list="fle,clos"``.
```
peternewman, line 54: Can we make these examples match the config file examples below please?

peternewman (changes requested):
Just one minor comment

README.rst
kolyshkin: done

README.rst
```diff
@@ -96,12 +100,13 @@ be specified in this file (without the preceding dashes), for example::
...
     [codespell]
     skip = *.po,*.ts,./src/3rdParty,./src/Test
+    ignore-words-list = creat,clos
     count =
     quiet-level = 3
 
 This is equivalent to running::
 
-    codespell --quiet-level 3 --count --skip "*.po,*.ts,./src/3rdParty,./src/Test"
+    codespell --quiet-level 3 --count --skip "*.po,*.ts,./src/3rdParty,./src/Test" --ignore-words-list creat,clos
```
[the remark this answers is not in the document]
kolyshkin: This is unrelated to this PR -- the order was wrong before this.

Fixed the config file order to match that of CLI

kolyshkin:
Rebased; addressed review comments; PTAL @peternewman

README.rst
```diff
@@ -41,23 +41,41 @@ Some noteworthy flags::
 
     codespell -w, --write-changes
 
-The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
+The ``-w`` flag will actually implement the changes recommended by codespell.
+Not running with ``-w`` flag is the same as with doing a dry run. It is
```
peternewman, line 45: I think grammatically it should probably be like this:
```suggestion
Not running with ``-w`` flag is the same as doing a dry run. It is
```

peternewman (approved):
LGTM, just one minor tweak

peternewman:
@luzpaz do you want to take a look too as you've tweaked this in the past?

README.rst
kolyshkin: This is not something that I have added. Anyway, I agree that "with" should not be here.

I should have changed that description to something more concise and clear. Done.

kolyshkin:
I am pretty sure there are many more improvements that can be done to the README in this repo. Originally my intention was to merely add `--ignore-words-list` (because I spent quite some time looking for it, and wanted others to find it faster/easier). It is getting out of hand now as I am adding more and more fixes on top, and already had to rebase this once. The fact that I am not very familiar with ReST syntax doesn't help either 😓. 

Can this be merged as is (unless there are some incorrect things that this PR *adds*), and implement more improvements  separately on top of this?

PS if you want to see how this is rendered (by Github, that is), see https://github.com/kolyshkin/codespell/blob/patch-1/README.rst.

peternewman:
> I am pretty sure there are many more improvements that can be done to the README in this repo. Originally my intention was to merely add `--ignore-words-list` (because I spent quite some time looking for it, and wanted others to find it faster/easier). It is getting out of hand now as I am adding more and more fixes on top, and already had to rebase this once.

Yeah sorry, just spotting related bits in the course or reviewing your PR.

> Can this be merged as is (unless there are some incorrect things that this PR _adds_), and implement more improvements separately on top of this?

Yes certainly. I'd like to give @luzpaz a little while to take a look, but happy to merge if they've not reviewed soon.

> PS if you want to see how this is rendered (by Github, that is), see https://github.com/kolyshkin/codespell/blob/patch-1/README.rst.

Yeah, I'd had a look already and it looks fine to me thanks!

luzpaz (approved):

luzpaz:
Thanks for the refinements. LGTM

Issue #1578, opened 2020-06-27
Can't ignore an unknown acronym only when it's written in upper case


vadz:
I have a lot of occurrences of the word [SEH](https://en.wikipedia.org/wiki/Structured_Exception_Handling) in my code and I'd like codespell to leave it alone, but this doesn't seem to work, using neither `-I` nor `-L` option, e.g.:
```
% echo SEH | codespell -L SEH -
1: SEH
        SEH ==> SHE
```

The problem seems to be the case, i.e. if I use `-L seh`, it does recognize "SEH", but I'd prefer not to have to do it, as "seh" is much more likely to be a misspelling of "she" than anything else -- "SEH" itself is always written in upper case.

I'm not sure if it's the same as #1128, as it happens for camel case identifier there, while my problem doesn't seem to be related to camel case (or an identifier, for that matter, as the problematic acronym only occurs inside comments), sorry in advance if it's the same bug and so an already known problem.

peternewman:
Hi @vadz ,

Currently codespell is case-insensitive, so as per the readme:
https://github.com/codespell-project/codespell/blob/master/README.rst

> The -I flag can be used for a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: codespell -I path/to/file.txt to execute codespell referencing said list of allowed words. Important note: The list passed to -I is case-sensitive based on how it is listed in the codespell dictionaries.

So you need to use `-L seh`.

Our dictionary checking needs to be case-insensitive because for example a C define is normally written in upper case so we'd want to catch `SPELING` as being an error without the dictionary needing to have both speling AND SPELING in it. I guess we could theoretically make the ignores match the case of the word in the code, but if you're using a term which codespell thinks is a typo you might have it as a define in capitals, a variable and possible also in camel case, so I think it would get rather complicated. I'll leave this open though, as it is different to the camel case request and another possible feature request.

Holzhaus:
I have the same problem. I want to exclude `InOut`, but not `inout` (because that's possibly a typo and supposed to say `input`). Is there some way?

Holzhaus:
How about filtering matches after looking them up in the dictionary?

peternewman:
> I have the same problem. I want to exclude `InOut`, but not `inout` (because that's possibly a typo and supposed to say `input`). Is there some way?

So you want to do it based on the case of the typo?

Currently `-x FILE, --exclude-file FILE` would be your only option.

[...]

PR #2049, opened 2021-08-30
Tweak --ignore-regex to ignore consecutive matches

peterjc:
Consider this example for ignoring 2 or 3 letter upper case words (typical of acronyms):

```
codespell --ignore-regex '\W([A-Z]{2,3})\W'
```

If there are two consecutive matches in a string like "text TE OTU text" only one gets removed - the problem is the space needs to be used twice, the end of `` TE `` and the start of `` OTU `` as the second but overlapping regex match.

Cross reference #1578.


peternewman (changes requested):
That makes sense thanks @peterjc .

Please can you add some tests for it so we won't regress in future (even if we change to a different workaround).

peterjc:
I wasn't sure if this change was in the spirit of the original regex implementation, but since you like the idea, I agree a test or two is a good idea.

peterjc:
(I would suggest doing a squash-merge since the commit history is less than ideal; I was lazy and did this via the GitHub editor which makes trailing whitespace too easy)

peterjc:
I've tested this locally now:

```
$ codespell flag.txt 
flag.txt:1: TE ==> THE, BE, WE, TO
flag.txt:1: OTU ==> OUT
$ codespell --ignore-regex '\W([A-Z]{2,3})\W' flag.txt 
$ codespell --ignore-regex "\W([A-Z]{2,3})\W" flag.txt 
$ codespell --ignore-regex="\W([A-Z]{2,3})\W" flag.txt 
$ codespell --ignore-regex='\W([A-Z]{2,3})\W' flag.txt 
$ codespell --ignore-regex \W([A-Z]{2,3})\W flag.txt 
-bash: syntax error near unexpected token `('
$ codespell --ignore-regex=\W([A-Z]{2,3})\W flag.txt 
-bash: syntax error near unexpected token `('
```

Looks like the protective quotes are being lost somewhere in the test framework...

peterjc:
Solved, rather than doing what works at the command line, I needed to match what Python receives from the command line.

Also rebased for a clearer history.

Issue #2046, opened 2021-08-22
format of FILE as in --exclude-file=FILE


DimitriPapadopoulos:
What is the format of **FILE** as in `--exclude-file=FILE`?

Is it perhaps of the form "file:line"? If so, I suggest fixing the help message from:
```
  -x FILE, --exclude-file=FILE
                        FILE with lines that should not be changed
```
to:
```
  -x FILE, --exclude-file=FILE
                        FILE with file:line pairs that should not be changed
```
That would avoid misunderstandings such as https://github.com/codespell-project/codespell/issues/1601#issuecomment-663634656 and https://github.com/codespell-project/actions-codespell/issues/33#issuecomment-903241243.

DimitriPapadopoulos:
After reading the source code, It appears it's really simple line numbers in the file, not specific to a file. So if you want to exclude line 100 of a specific file, you cannot but exclude line 100 of all files.

peternewman:
It's the whole line, not the line number. See #2052 for work to fix this.

DimitriPapadopoulos:
Ah, I see. Sorry about that. Then let's follow up in #2052.

PR #2052, opened 2021-09-01
Add note about --exclude-file in README

bwitt:
Add a quick note about using `--exclude-file` to exclude whole lines


peternewman (changes requested):
I think I prefer your wording, but it would probably make sense if it matched the help file:
https://github.com/codespell-project/codespell/blob/d598d5b/codespell_lib/_codespell.py#L347-L349

Or can you at least add your second sentence to the command help please.

bwitt:
@peternewman I've updated both sets of text to try and bring them closer together; how does that look?

README.rst
```diff
@@ -51,6 +51,10 @@ The ``-I`` flag can be used for a list of certain words to allow that are in the
 
 The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it.  **Important note:** The list passed to ``-L`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
 
+    codespell -x FILE, --exclude-file=FILE
+
+The ``-x`` flag can be used to specify a file FILE with lines to be excluded from being checked for errors or changed.  These lines in FILE should match the to-be-excluded lines exactly.
```
peternewman, line 56: I liked your "whole lines" from the earlier revision.

peternewman (changes requested):
Perhaps also take a look at @kolyshkin work in https://github.com/codespell-project/codespell/pull/2040/files , I suspect their formatting of the METAVAR FILE should make things more readable (in the README file).

README.rst
peternewman: I don't know what @DimitriPapadopoulos thinks, but whole might help for #2046 too.

codespell_lib/_codespell.py
```diff
@@ -345,8 +345,10 @@ def parse_options(args):
                              'you\'d give "*.eps,*.txt" to this option.')
 
     parser.add_argument('-x', '--exclude-file', type=str, metavar='FILE',
-                        help='FILE with lines that should not be checked for '
-                             'errors or changed')
+                        help='specify a file FILE with whole lines to be '
+                             'excluded from being checked for errors or changed. '  # noqa: E501
```
peternewman, line 349: I don't think we need the noqa here, we're just joining strings together already so we can wrap it ourselves so it fits.

peternewman:
> I notice https://github.com/codespell-project/codespell/pull/2040/files also rewords the descriptions in the README a bit (removing `The ... flag can be used ...`). Should we do the same here?

I suspect fewer words makes it easier to read, so that's probably a good idea.

peternewman (changes requested):

bwitt:
thanks! I've updated it; how does that look?

kolyshkin:
@bwitt can you please squash your commits?

bwitt:
squashed

README.rst
```diff
@@ -51,6 +51,10 @@ The ``-I`` flag can be used for a list of certain words to allow that are in the
 
 The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it.  **Important note:** The list passed to ``-L`` is case-sensitive based on how it is listed in the codespell dictionaries. ::
 
+    codespell -x FILE, --exclude-file=FILE
+
+Ignore whole lines that match those in ``FILE``.  These lines in ``FILE`` should match the to-be-excluded lines exactly.
```
kolyshkin, line 56: nit:
```suggestion
Ignore whole lines that match those in ``FILE``.  The lines in ``FILE`` should match the to-be-excluded lines exactly.
```

codespell_lib/_codespell.py
```diff
@@ -345,8 +345,9 @@ def parse_options(args):
                              'you\'d give "*.eps,*.txt" to this option.')
 
     parser.add_argument('-x', '--exclude-file', type=str, metavar='FILE',
-                        help='FILE with lines that should not be checked for '
-                             'errors or changed')
+                        help='ignore whole lines that match those '
+                             'in the file FILE. These lines in FILE '
```
peternewman, line 349: @kolyshkin 's improvement didn't make it into the main program help:
```suggestion
                             'in the file FILE. The lines in FILE '
```

bwitt:
@peternewman rebased and squashed to one commit; please let me know if there's anything else I need to do.

kolyshkin (approved):
LGTM

PR #2059, opened 2021-09-10
Fix LGTM recommendation: Module is imported with 'import' and 'import from'

DimitriPapadopoulos:
Module '`os`' is imported with both '`import`' and '`import from`'

I feel we don't need this abbreviation here:
- Lines are not too long.
- `os.path` is known by all Python developers, while the `op` abbreviation is not (I believe the `osp` abbreviation is more common).

And actually we had both `os.path.exists()` and `op.join()` in the code. That's confusing.

Here is the LGTM alert:
https://lgtm.com/projects/g/codespell-project/codespell/


peternewman (approved):
LGTM :laughing: Seems legit.

PR #2061, opened 2021-09-13
Fix a couple trailing spaces

DimitriPapadopoulos:


codespell_lib/data/linux-kernel.exclude
```diff
@@ -73,7 +73,7 @@ S: 1326 De Val-Brillant
 infinit:
 	bra	infinit
 the read on an 8-byte boundary (e.g., if you seeked an odd number of bytes
-#define ISNT		   12 
+#define ISNT		   12
```
peternewman, line 76: We should leave this as is, I think the line matching will fail otherwise:
https://github.com/torvalds/linux/blob/b24413180f5600bcb3bb70fbed5cf186b60864bd/arch/parisc/include/asm/perf.h#L31

peternewman (changes requested):
Just the one change thanks.

DimitriPapadopoulos:
Ah, if it's a copy from upstream, it should be kept as is. I'll look into it.

codespell_lib/data/linux-kernel.exclude
DimitriPapadopoulos: Fixed now.

peternewman (approved):
LGTM thanks!

PR #2066, opened 2021-09-16
chmod -x

DimitriPapadopoulos:
File `codespell_lib/_codespell.py` does not start with `#!` anymore, therefore it should not be identified as executable by Git.


peternewman (approved):
LGTM

PR #2067, opened 2021-09-16
Improve the definition of the "rare" builtin dictionary

DimitriPapadopoulos:
For an end user who is not familiar with the inner details of the script, the meaning of the old definition was ambiguous. It wasn't
clear whether using this dictionary flags the rare words as errors (false positives) or on the contrary does not flag these rare words
(false negatives).

I do reckon the leading `for` was consistent with the rest of the definitions, but cannot suggest a good definition starting with `for`.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -54,7 +54,7 @@ _builtin_dictionaries = (
     # realistic for obscure words
     ('clear', 'for unambiguous errors', '',
         False, None, supported_languages_en, None),
-    ('rare', 'for rare but valid words', '_rare',
+    ('rare', 'for rare (but valid) words which are likely to be errors', '_rare',  # noqa: E501
         None, None, None, None),
     ('informal', 'for making informal words more formal', '_informal',
         True, True, supported_languages_en, supported_languages_en),


codespell_lib/_codespell.py
```diff
@@ -54,7 +54,7 @@
     # realistic for obscure words
     ('clear', 'for unambiguous errors', '',
         False, None, supported_languages_en, None),
-    ('rare', 'for rare but valid words', '_rare',
+    ('rare', 'to categorize rare but valid words as errors', '_rare',
```
peternewman, line 57: How about one of these?:
```suggestion
    ('rare', 'for rare (but valid) words which are likely to be errors', '_rare',
```
```suggestion
    ('rare', 'for rare (but valid) words which are likely to be errors where someone meant the correction', '_rare',
```

I would say we're not really categorising. Also can you try running the --help, I think it uses this output and not having for will make for some odd phrasing.

peternewman (changes requested):
LGTM, can you fix the lint issue it's introduced though please.

DimitriPapadopoulos:
The linter issue has been fixed, with a `noqa`. It could also be fixed by breaking lines on the commas separating arguments. Which do you prefer?

peternewman:
> The linter issue has been fixed, with a `noqa`. It could also be fixed by breaking lines on the commas separating arguments. Which do you prefer?

That looks perfect as it matches the others a few lines down.

peternewman (approved):

PR #2071, opened 2021-09-21
Add configulation->configuration correction.

ryo-tagami:
Should be beneficial: https://github.com/search?q=configulation&type=code

Signed-off-by: Ryo Tagami <rtagami@airstrip.jp>


peternewman (changes requested):
The mind baffles, but nice find!

Can you add `configulate` and `configulations` too.

ryo-tagami:
Good point, added `configulate` and `configulations`.

peternewman (approved):
LGTM thanks

PR #2074, opened 2021-09-22
WIP: Add --uri-ignore-words

DimitriPapadopoulos:
This is in addition to `--uri-ignore-words-list`, similar to the `--ignore-words` / `--ignore-words-list` pair.


peternewman:
I've converted this to draft as it's still WIP and needs some tests added at least.

DimitriPapadopoulos:
This option is really needed, without it _codespell_ reports errors in its own setup!

I have a file which contains:
```
... <...@usal.es>
```

After attempting to ignore **`usal`** in URIs  in **`setup.cfg`**:
```
uri-ignore-words-list = usal
```
_codespell_ reports errors in its own setup:
```
./setup.cfg:79: usal ==> usual

There are two ways to fix this:
1. have codespell remember the location of its setup in `setup.cfg`, but the changes would be quite extensive,
2. use this new option `--uri-ignore-words` and make sure it skips the file given as argument (I need to make sure of that in the tests).
```

Once this is done, I recommend we add a warning or even bail out when **`uri-ignore-words-list`** is used in a config file, as it doesn't work properly.

PR #2077, opened 2021-09-22
Fix spelling in snap config

a1346054:


peternewman (approved):
LGTM, but we should probably try and get to the bottom of why we haven't caught this already before we merge  it.

a1346054:
There's no rush merging this PR :)

DimitriPapadopoulos:
Just add `mispellings->misspellings` to the dictionary, [here](https://github.com/codespell-project/codespell/blob/23f2cd1/codespell_lib/data/dictionary.txt#L18953), and you should be all set.

Explanation:
Currently codespell does not check the plural inflection when only the singular of a typo is given. It'sprobably not  easy to improve, espacially when it comes to typos: I suspect [inflect](https://github.com/jaraco/inflect) wlll  fail on typos. What could be done, not in this PR of course, would be to make sure the plural inflection of a correction can be found in the dictionary when the singular form is.

peternewman:
> Just add `mispellings->misspellings` to the dictionary, [here](https://github.com/codespell-project/codespell/blob/23f2cd1/codespell_lib/data/dictionary.txt#L18953), and you should be all set.

Thanks @DimitriPapadopoulos good spot, I'd not gone looking. Added in #2080.

> Explanation:
> Currently, codespell does not catch the plural inflection when only the singular form of a typo is given. It's probably not easy to improve, especially when it comes to typos: I suspect [inflect](https://github.com/jaraco/inflect) might fail on typos. What could be done, not in this PR of course, would be to make sure the plural inflection of a suggestion can be found in the dictionary when the singular form is.

I see you've opened #2079 for this thanks.

PR #2078, opened 2021-09-23
Simpler regex and consistent use in tests

DimitriPapadopoulos:
* Some regex'es were compiled, others not. Not convinced it's useful to pre-compile for such simple tests. Defining the regex locally simplifies future maintenance.
* No need to match "anything" (`.*`) at the end of a regex.
* No need to match "anything" (`.*`) at the start of a regex, use `find()` instead of `match()`.
* A regex to find a comma is overkill.


codespell_lib/tests/test_dictionary.py
```diff
@@ -87,13 +84,13 @@ def _check_aspell(phrase, msg, in_aspell, fname, languages):
 
 
 def _check_err_rep(err, rep, in_aspell, fname, languages):
-    assert ws.match(err) is None, 'error %r has whitespace' % err
-    assert comma.match(err) is None, 'error %r has a comma' % err
+    assert not re.search(r'\s', err), 'error %r has whitespace' % err
```
peternewman, line 87: Have you benchmarked this? There's a rather unclear note in the docs:
https://docs.python.org/3/library/re.html#re.compile

But if this is being processed from scratch each time it will do it for each match in each line (>32,000 lines alone)...

peternewman (commented):
Just one comment

codespell_lib/tests/test_dictionary.py
DimitriPapadopoulos: And there's a bunch of other regexes that are not optimized:
```python
    for (r, msg) in [
            (r'^,',
             '%s starts with a comma'),
            (r'\s,',
             '%s contains a whitespace character followed by a comma'),
            (r',\s\s',
             '%s contains a comma followed by multiple whitespace characters'),
            (r',[^ ]',
             '%s contains a comma *not* followed by a space'),
            (r'\s+$',
             '%s has a trailing space'),
            (r'^[^,]*,\s*$',
             '%s has a single entry but contains a trailing comma')]:
        assert not re.search(r, rep), (msg % (prefix,))
```
Honestly, I don't see how optimizing the first regex helps in any way. EIther we optimize all regexes, or we optimzie none.

codespell_lib/tests/test_dictionary.py
```diff
@@ -86,30 +83,40 @@ def _check_aspell(phrase, msg, in_aspell, fname, languages):
         assert not this_in_aspell, '%s should not %s' % (msg, end)
 
 
+whitespace = re.compile(r'\s')
+start_whitespace = re.compile(r'^\s')
+start_comma = re.compile(r'^,')
+whitespace_comma = re.compile(r'\s,')
+comma_whitespaces = re.compile(r',\s\s')
+comma_without_whitespace = re.compile(r',[^ ]')
```
peternewman, line 91: Given this isn't all whitespace, just an actual space, can I suggest this change (and the equivalent below):
```suggestion
comma_without_space = re.compile(r',[^ ]')
```

PR #2082, opened 2021-09-23
Add examlpe->example

clickthisnick:
Finally a typo I made that is so bad its missing from codespell 😉

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -12074,6 +12074,7 @@ examinated->examined
 examing->examining
 examinining->examining
 examles->examples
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -12074,6 +12074,7 @@ examinated->examined
 examing->examining
 examinining->examining
 examles->examples
+examlpe->example
```
vikivivi, line 12077: If you don't mind adding the plural `examlpes->examples`, see https://github.com/search?q=examlpes&type=code

vikivivi (changes requested):

PR #2085, opened 2021-09-24
Fix Deepsource issues

DimitriPapadopoulos:


DimitriPapadopoulos:
Perhaps the `list()`/`set()` part is the most debatable. Links to relevant [DeepSource.io](https://deepsource.io/) pages:
* [7. Not using literal syntax to initialize empty `list`/`dict`/`tuple`](https://deepsource.io/blog/8-new-python-antipatterns/#7-not-using-literal-syntax-to-initialize-empty-listdicttuple)
  > It is relatively slower to initialize an empty dictionary by calling `dict()` than using the empty literal, because the name `dict` must be looked up in the global scope in case it has been rebound. Same goes for the other two types — `list()` and `tuple()`.
* [Consider using literal syntax to create the data structure — PTC-W0019](https://deepsource.io/gh/DimitriPapadopoulos/codespell/issue/PTC-W0019/occurrences)
  > **DESCRIPTION**
  > 
  > Using the literal syntax can give minor performance bumps compared to using function calls to create `dict`, `list` and `tuple`.
  > [...]
  > This is because here, the name `dict` must be looked up in the global scope in case it has been rebound. Same goes for the other two types `list()` and `tuple()`.

I am happy to discard debatable commits - that's actually why I have multiple commits.

jdufresne (approved):

DimitriPapadopoulos:
Some issues, such as `list()`/`set()` → `[]`/`{}`, have been fixed in #2364.

The remaining issues are less debatable. I would appreciate someone reviewing them.

larsoner:
Thanks @DimitriPapadopoulos

PR #2087, opened 2021-09-27
Add `tox.ini` to the list of default config files

### LINKED ISSUE #2086: tox.ini config file
What about adding `tox.ini` to the default locations of the config file, in addition to `setup.cfg` and `.codespellrc`?

Some projects such as [python-versioneer](https://github.com/python-versioneer/python-versioneer) have a `tox.ini` file but no `setup.cfg` file.

DimitriPapadopoulos:
Closes #2086.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,9 +94,10 @@ Using a config file
 Command line options can also be specified in a config file.
 
 When running ``codespell``, it will check in the current directory for a file
-named ``setup.cfg`` or ``.codespellrc`` (or a file specified via ``--config``),
-containing an entry named ``[codespell]``. Each command line argument can
-be specified in this file (without the preceding dashes), for example::
+named ``.codespellrc``, ``setup.cfg`` or ``tox.ini`` (or a file specified via
+``--config``), containing an entry named ``[codespell]``. Each command line
+argument can be specified in this file (without the preceding dashes), for
+example::
 
     [codespell]
     skip = *.po,*.ts,./src/3rdParty,./src/Test
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -400,7 +400,7 @@ def parse_options(args):
     options = parser.parse_args(list(args))
 
     # Load config files and look for ``codespell`` options.
-    cfg_files = ['setup.cfg', '.codespellrc']
+    cfg_files = ['tox.ini', 'setup.cfg', '.codespellrc']
     if options.config:
         cfg_files.append(options.config)
     config = configparser.ConfigParser()


README.rst
```diff
@@ -94,9 +94,10 @@ Using a config file
 Command line options can also be specified in a config file.
 
 When running ``codespell``, it will check in the current directory for a file
-named ``setup.cfg`` or ``.codespellrc`` (or a file specified via ``--config``),
-containing an entry named ``[codespell]``. Each command line argument can
-be specified in this file (without the preceding dashes), for example::
+named ``setup.cfg``, ``tox.ini`` or ``.codespellrc`` (or a file specified via
```
peternewman, line 97: I think the order in the readme should match the order they are loaded in the code:
```suggestion
named ``setup.cfg``, ``.codespellrc`` or ``tox.ini`` (or a file specified via
```

peternewman (commented):
Looks fine in principle.

This does raise one interesting question, for backwards compatibility it would be nice if tox.ini (i.e. the new one) had the lowest priority. I don't currently know how that works, given it renders fake command line data, does `--foo 1 --foo 2` mean that 1 or 2 wins, or we get both?

If we don't currently have a test for that we ought to add one (I don't know if you fancy trying that, feel free to do another PR), or I'll have a go. If we do, we should extend it to check tox.ini works too.

README.rst
DimitriPapadopoulos: Fixed the order of configuration files in the documentation.

DimitriPapadopoulos: The priority of configuration files is an issue in theory, but is it worth addressing in practice? Because `tox.ini` was not a configuration file (until now) what are the chances that existing `tox.ini` files in the wild contain a `[codespell]` section?

Additionally, the documentation of [`configparser.ConfigParser.read()`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read) does not specify how it behaves when passed an iterable of potential configuration file locations. Which file has priority? My guess is that the latest file actually has priority, as its options will overwrite identical options read from previous files. That's only a guess of course, but if so, `tox.ini` should be added first instead of last to have the lowest priority. As an alternative, we could read each possible configuration file separately and merge them manually in a well-defined order. But then, I feel we shouldn't address such details if even the [`configparser`](https://docs.python.org/3/library/configparser.html) modules does not.

DimitriPapadopoulos: I have actually modified my commit to use the reverse order from [configparser.ConfigParser.read()](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read) in the documentation.

DimitriPapadopoulos: I had a look at [`configparser.RawConfigParser._read()`](https://github.com/python/cpython/blob/53a54b781d1f05f2d0b40ce88b3da92d5d23e9d2/Lib/configparser.py#L973) and indeed the last file takes precedence over previous files.

I ended up opening Python bug  https://github.com/python/cpython/issues/96765.

DimitriPapadopoulos:
I think `tox.ini` will be replaced by `pyproject.toml` in the future. It's probably not worth pursuing this.

PR #2092, opened 2021-10-04
laer->later

DimitriPapadopoulos:
neurospin/neurospin_to_bids@9bb4941


luzpaz:
Could also be `layer`

DimitriPapadopoulos:
Indeed. Are there automated methods to help find suggestions?

DimitriPapadopoulos:
We could scan a dictionary (which one?) and calculate the [Damerau–Levenshtein distance](https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance) between the typo and all words in the dictionary (and keep those at a distance of 1 or maybe even 2).

luzpaz:
@DimitriPapadopoulos that would be pretty amazing

PR #2100, opened 2021-10-07
Remove encoding declaration

### LINKED ISSUE #2099: Remove `# -*- coding: utf-8 -*-`?
In Python 3, the [default source file encoding is UTF-8](https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations). As far as the Python interpreter is concerned we could remove `# -*- coding: utf-8 -*-`.

Yet the encoding declaration might be useful to some IDEs, especially on Windows which still seems to be enforcing Windows-1252 as the default encoding on "Western" locales. On the other hand, VS tends to write into **UTF-8-BOM** (with an initial BOM) so we're doomed anyway!

DimitriPapadopoulos:
In Python 3, the default source file encoding is UTF-8. As far the
Python interpreter is concerned, we can remove the encoding declaration:
```
# -*- coding: utf-8 -*-
```

Closes #2099.


bl-ue (approved):

jdufresne (approved):

DimitriPapadopoulos:
Fixed in #2364.

PR #2102, opened 2021-10-10
Add spelling corrections

luzpaz:


vikivivi:
It would be great if you could add these.
Another alternate plural of formula is formulae.
https://github.com/search?q=formuale&type=code
```text
formuale->formulae
```

PR #2105, opened 2021-10-11
fix flaky test on test_dictionary.py::test_dictionary_looping

lonly7star:
This PR aims to fix the flaky test on **test_dictionary.py::test_dictionary_looping**  so the test could pass for multiple test runs 

#### The result

The test can pass when running only once, but fail when running the test suit multiple times. When asserting the key pair does not exist on the `global_err_dicts ` or `global_pairs`, the pair actually exists start on the second test run. 

#### Steps to reproduce the issue

1.  install pytest flaky finder with `pip install pytest-flakefinder`
2.  run pytest with flake-finder with command  `pytest -k test_dictionary.py --flake-finder`

#### Issue of the code

The reason is the code tried to assert the key pair or error read from the file doesn't exist. However, the `global_err_dicts ` and `global_pairs` only initialize once per pytest runs, so start from the second run of `test_dictionary_looping`, the assert will fail.

#### Proposed solution
Adding a variable to track how many txt files we had read. With a pytest fixture to initialize both global_err_dicts ` and `global_pairs` if the variable indicates it is a new test run.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1653,6 +1653,7 @@ alwasys->always
 alwauys->always
 alway->always
 alwyas->always
+alwys->always
 alyways->always
 amacing->amazing
 amacingly->amazingly
@@ -10851,7 +10852,7 @@ doublely->doubly
 doubletquote->doublequote
 doucment->document
 doucmentated->documented
-doucmentation->documenting
+doucmentation->documentation
 doucmented->documented
 doucmenter->documenter
 doucmenters->documenters
@@ -11569,6 +11570,7 @@ enironments->environments
 enities->entities
 enitities->entities
 enitity->entity
+enitre->entire
 enity->entity
 enivornment->environment
 enivornments->environments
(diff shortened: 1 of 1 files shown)


peternewman:
> This PR aims to fix the flaky test on **test_dictionary.py::test_dictionary_looping** so the test could pass for multiple test runs
>
> The test can pass when running only once, but fail when running the test suit multiple times. When asserting the key pair does not exist on the `global_err_dicts ` or `global_pairs`, the pair actually exists start on the second test run.

I assume you've not had this test fail on a single run, so it's not flaky in the traditional sense but it's not compatible with re-running the test multiple times in the same session?

I'm not targeting this at you, but this fix seems like quite a bodge to me, isn't there some set up and tear down functions for the test (rather than the test suite) which we could deal with this in? Flaky ( https://github.com/box/flaky ) seems to support this stuff, based on issues like ( https://github.com/box/flaky/issues/124 https://github.com/box/flaky/issues/109 https://github.com/box/flaky/issues/53 https://github.com/box/flaky/issues/39 ). Maybe pytest-flakefinder does too and we just need to add setup/tearDown or add these to it?

> should I re-formate the code with the format requirements of codespell and submit a new PR?

You can just fix it in this PR if you'd prefer.

PR #2108, opened 2021-10-15
Misspelling: Jenkins

vikivivi:
INFO: List of code search URL:
  * https://github.com/search?q=jekins&type=code
  * https://github.com/search?q=jenkin&type=code  (187,186 hits)
  * https://github.com/search?q=jenkkins&type=code
  * https://github.com/search?q=jenkns&type=code

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -17550,9 +17550,13 @@ jave->java, have,
 javsscript->javascript
 jeapardy->jeopardy
 jeffies->jiffies
+jekins->Jenkins
 jelous->jealous
 jelousy->jealousy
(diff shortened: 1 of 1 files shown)


peternewman (approved):
LGTM thanks @vikivivi .

Indeed you're correct, the typo isn't a name (unlike say tim->time), so it shouldn't be in names. We should indeed try to correctly capitalise the corrections for the future, even if it's not made use of currently (as it's easier to have them correct on entry than have to go through the whole dictionary when we add that feature).

peternewman:
Are you happy now then @robin-wayve , then we can merge this?

robin-wayve:
Merge away!

PR #2121, opened 2021-10-22
move creat to code dictionary

### LINKED ISSUE #2120: `creat` is a linux syscall you might use in C/C++
See: https://linux.die.net/man/2/creat

I'm not sure this correction has a better home: should it be in `code`? ("for words common to code and/or mathematics that might be typos")

https://github.com/codespell-project/codespell/blob/5dcde7ecd65deb81fb30761d3562f87f1e93bd3e/codespell_lib/data/dictionary.txt#L8039

robin-wayve:
`creat(2)` is a Linux syscall. Also add crate as a possible correction.

Resolves #2120

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -7,7 +7,7 @@ clas->class
 cloneable->clonable
 cmo->com
 copyable->copiable
-creat->crate, create,
+creat->create, crate,
 define'd->defined
 dof->of, doff,
 dont->don't
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -7,6 +7,7 @@ clas->class
 cloneable->clonable
 cmo->com
 copyable->copiable
+creat->crate, create,
 define'd->defined
 dof->of, doff,
 dont->don't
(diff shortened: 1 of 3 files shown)


peternewman:
> CI/CD failed because suggested word `crate` is flagged as misspell in `dictionary_rare.txt` Should we remove valid word `crate` , `crated` from `dictionary_rare.txt` ? Or should CI/CD ignore checking against `dictionary_rare.txt`?
> 
> Since`dictionary_rare.txt` are for rare (but valid) words which are likely to be errors, so we should ignore checking against `dictionary_rare.txt`

I've added the necessary exception to the tests so we can suggest `creat->crate, create` in the code dictionary.

codespell_lib/data/dictionary_code.txt
```diff
@@ -7,6 +7,7 @@ clas->class
 cloneable->clonable
 cmo->com
 copyable->copiable
+creat->crate, create,
```
peternewman, line 10: I'd probably put create first as the more likely correction.
```suggestion
creat->create, crate,
```

peternewman (approved):
LGTM, just one suggestion to improve it.

luzpaz (approved):

PR #2143, opened 2021-11-11
ony can be also on

matkoniecz:
See https://github.com/streetcomplete/StreetComplete/commit/9276a4c47a3f496c60a60295885350ae4e431146 for case found in wild

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -20600,7 +20600,7 @@ ontext->context
 onthe->on the
 ontop->on top
 ontrolled->controlled
-ony->only
+ony->only,on,
 onyl->only
 oommits->commits
 opactity->opacity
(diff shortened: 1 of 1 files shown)


matkoniecz:
CI failure has way too many lines, but it seems to be caused by

```
skipping aspell tests. Got error during import:
  No module named 'aspell'
    'error during import:\n%s' % (exp,))
-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

unrelated to this PR

codespell_lib/data/dictionary.txt
```diff
@@ -20600,7 +20600,7 @@ ontext->context
 onthe->on the
 ontop->on top
 ontrolled->controlled
-ony->only
+ony->only,on,
```
peternewman, line 20603: Nit:
```suggestion
ony->only, on,
```

peternewman:
> CI failure has way too many lines, but it seems to be caused by

Read up a bit further:
`error ony: correction 'only,on,' contains a comma *not* followed by a space`

PR #1327, opened 2019-11-08
Typos from Linux Kernel

Gelma:
This is a collection of typos fixed by hand some years ago, by me.

Real typos. Not fantasy.

PR #2144, opened 2021-11-11
brouter is name of an existing project

matkoniecz:
https://github.com/abrensch/brouter 

it also seems to not be a likely misspelling

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -4404,7 +4404,6 @@ broswed->browsed
 broswer->browser
 broswers->browsers
 broswing->browsing
-brouter->router
 brower->browser
 browers->browsers
 browing->browsing
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -2,6 +2,7 @@ amin->main
 apoint->appoint
 atend->attend
 atending->attending
+brouter->router
 cas->case, cast,
 clas->class
 cloneable->clonable


peternewman:
It came in here ( https://github.com/codespell-project/codespell/pull/1327 ), we should just move it to the code dictionary so it won't be run by default or normally on codebases.

peternewman (changes requested):
As I didn't review properly before:
https://github.com/codespell-project/codespell/pull/2144#issuecomment-966233396

matkoniecz:
Is it better described as `dictionary_rare_false_positives`?

matkoniecz:
I added it to `_code` but without real understanding. No idea what exactly differentiates words present there and not present in main dictionary.

peternewman (approved):

peternewman:
> > won't be run by default or normally on codebases.
> 
> So dictionary_code is listing typos which should be ignored in code but not elsewhere? Either way brouter as project name will occur in both code and in documentation.

Probably more broadly in "computing related" source data. So if you were running codespell over a cookery book, you could include the code dictionary with no risk.

Personally I'd include it regardless and just skip the words specific to my language/usage etc, but things like stdio, uint or movei would obviously break a lot of stuff if auto-corrected, or generate lots of false positives so they're moved off into this dictionary.

> And some typos there (protecten, `storaged->storage, stored`, `storaget->storage`) that seem generally applicable...

protecten is a Linux register which is why it will be in there:
https://github.com/torvalds/linux/928296ea5da37838d7127de4b10f47cd97401b13/master/drivers/soc/mediatek/mtk-infracfg.c#L20

I've not bothered looking at the others, but there likely to be similar. So if you're writing something in C you might hit them, so you wouldn't use the code dictionary (or you'd ignore them), but if you're writing something else, you might.

> `codespell --help` describes it as `'code' for words common to code and/or mathematics that might be typos` so I am even more confused.

How so? The code and name dictionaries are negated, which should maybe be in there, i.e. don't run them if you're checking content including that type of thing (e.g. names, or code).

> Is it better described as `dictionary_rare_false_positives`?

Yes sort of, but they are a special class of rare false positives compared to the normal rare dictionary which just contains infrequently used but valid English words and their more likely/popular corrections.

> I added it to `_code` but without real understanding. No idea what exactly differentiates words present there and not present in main dictionary.

It could probably broadly be described that running codespell with automatic fix mode on over the Linux kernel would be fine with the main dictionary, but not with code. Aside from any domain specific edge cases (e.g. acn->can breaks for me in one project as ACN is a protocol name).

matkoniecz:
> How so? The code and name dictionaries are negated, which should maybe be in there, i.e. don't run them if you're checking content including that type of thing (e.g. names, or code).

so maybe instead

> 'code' for words common to code and/or mathematics that might be typos

use

> 'code' for words from code and/or mathematics that are likely typos in other context (such as uint)

peternewman:
That certainly sounds like an improvement to me, do you want to open a PR @matkoniecz ?

matkoniecz:
#2150

PR #2147, opened 2021-11-12
"keyserver" is correct, fixes #2084

### LINKED ISSUE #2084: "keyserver" is correct
It looks like _keyserver_ is a correct technical word in cryptography. For example, the [Wikipedia _key server_ entry](https://en.wikipedia.org/wiki/Key_server_(cryptographic)) has 27 occurrences of _key server_ and 31 occurrences of _keyserver_.

On Google, [1,140,000 hits for _keyserver_](https://www.google.com/search?q=keyserver) and [537,000 hits for _key server_](https://www.google.com/search?q=%22key+server%22) - but it's true that a large part of the _keyserver_ hits are command line examples, code, URLs.

See discussion in https://github.com/numpy/numpy/pull/19911#discussion_r713107622.

matkoniecz:


peternewman (commented):
See further comment in the issue, I'm not sure this should go.

vikivivi:
`keyserver` is a very common term being in use, I have no objection removing it.

DimitriPapadopoulos:
These are not part of `dictionary.txt`, they are only part of `dictionary_code.txt`, which is not supposed to be used or at least overly trusted to check source code:
```
keyserver->key server
keyservers->key servers
```

Isn't that sufficient ?

 @matkoniecz In any case, GitHub CI tests are stuck for some reason and I cannot restart them. Rebasing and pushing your branch should restart them.

matkoniecz:
> they are only part of dictionary_code.txt, which is not supposed to be used or at least overly trusted to check source code

I that I have seen "keyserver" also in prose describing computer security.

See say https://en.wikipedia.org/wiki/Key_server_(cryptographic)#Public_versus_private_keyservers ("keyserver" is in a section name in "Key server (cryptographic)" article)

DimitriPapadopoulos (approved):

PR #1870, opened 2021-01-31
Add several spelling corrections

luzpaz:

PR #2149, opened 2021-11-12
remove "on-path attacker" as replacement for "man-in-the-middle" (the proposed term is unusual replacement for a standard term that is entirely fine)

matkoniecz:
In this case:

- while https://tools.ietf.org/id/draft-knodel-terminology-00.html cited in justification for this change claims "It is not a standard term, not as clear as its alternative “on-path attacker”, and should therefore be avoided." it seems not true to me. 
- I am likely actually from group that is trying to be "protected" so I feel relatively OK with declaring that it seems weird to me (though there could be more complicated explanation  for that replacement that would nullify this and claim that also here I am from group not supposed to comment)
 
Maybe it is preferred among people known to author of the cited text or in a highly technical academic writing, it is not popular in general, it is not standard at all in general use. I encountered it here for the first time in my life despite being interested in computer security and related topics. (though I am not an expert)
 
For example there is not even a Wikipedia redirect as of 2021-11-11 - see https://en.wikipedia.org/w/index.php?search=on-path+attacker It may be worth creating for people confused by this term and looking for an explanation, if that term is actually being used. [man-in-the-middle](https://en.wikipedia.org/w/index.php?title=Man-in-the-middle&action=history) redirect exists since 2005.

Term "man-in-the-middle" is actually clear, communicates well and is actually case of an useful metaphor.

"Terminology, Power and Oppressive Language" mentions https://www.orwellfoundation.com/the-orwell-foundation/orwell/essays-and-other-works/politics-and-the-english-language/ as part of justification but I have no idea which part of this text is supposed to encourage not using well fitting and explanatory terms such as "man-in-the-middle". 

Both are metaphors, similar in terms of being abstract ("on-path attacker" seems worse here?), no significant difference in length and so on.

Though it is nice to read someone complain about the same things that make me irritated decades later :)

> Many political words are similarly abused. The word Fascism has now no meaning except in so far as it signifies ‘something not desirable’. The words democracy, socialism, freedom, patriotic, realistic, justice, have each of them several different meanings which cannot be reconciled with one another. In the case of a word like democracy, not only is there no agreed definition, but the attempt to make one is resisted from all sides.

----------------------------------------

Also, this rule would replace "man-in-the-middle attack" by "on-path attacker attack" (not that "on-path attack" would be substantially better and explaining well)

-----------------------------------------

BTW, what kind of political problem is with term "blueish"? Or is it this one single non-political replacement there? ( and it is about https://english.stackexchange.com/questions/18824/why-are-blueish-and-bluish-both-considered-correct-spellings )

### DIFF
--- a/codespell_lib/data/dictionary_usage.txt
+++ b/codespell_lib/data/dictionary_usage.txt
@@ -1,7 +1,6 @@
 blacklist->blocklist
 blacklists->blocklists
 blueish->bluish
-man-in-the-middle->on-path attacker
 master->primary, leader, active, writer, coordinator, parent, manager, main,
 masters->primaries, leaders, actives, writers, coordinators, parents, managers,
 slave->secondary, follower, standby, replica, reader, worker, helper, subordinate, subsystem,


vikivivi:
My 2 cents only. I am not here to debate. Personally, I do not feel MITM offensive. But some may need to be diverse and inclusive in their coding and documentation and use tool like codespell.

See https://www.cio.com/article/3262704/diversity-and-inclusion-8-best-practices-for-changing-your-culture.html

Microsoft Style Guide quoted "_Microsoft technology reaches every part of the globe, so it's critical that all our communications are inclusive and diverse._"

Recommended using "gender-neutral" alternatives for common terms. 
such as "man, mankind" -> "humanity, people, humankind"

See https://docs.microsoft.com/en-us/style-guide/bias-free-communication

Google developer documentation style guide quoted for man-in-the-middle (MITM)
"_Avoid using gendered terms. Instead use terms like on-path attacker, person-in-the-middle (PITM)._"

See https://developers.google.com/style/word-list#letter-m

Codespell is just a tool for many who care about their coding and documentation standard.

The default enabled codespell dictionaries are `clear,rare`. Under normal use, codespell do not enable `usage` dictionary.
This mean codespell will not impact anyone by remaining neutral and being inclusive and diverse. Users who need to be "gender-neutral" can enable `usage` dictionary at wish.

Maybe you might want to consider withdrawing this PR.

peternewman:
> BTW, what kind of political problem is with term "blueish"? Or is it this one single non-political replacement there? ( and it is about https://english.stackexchange.com/questions/18824/why-are-blueish-and-bluish-both-considered-correct-spellings )

That was my error, I suggested the wrong place to @luzpaz in their PR (just looking at the dictionary name not it's description probably). I guess it probably wants to go in rare or informal but it's a bit of a grey area:
https://github.com/codespell-project/codespell/pull/1870#discussion_r579344511

matkoniecz:
> My 2 cents only. I am not here to debate. > Personally, I do not feel MITM offensive. But some may need to be diverse and inclusive in their coding and documentation and use tool like codespell.

I am also not going to debate/discuss/reask in case of this PR getting rejected.

But to me it seems case where removing it from replacement list is improving it. And definitely I would not want endorsing/encouraging/recommending this specific change.

But obviously it is not my project, it is just suggestion and noone is obligated to follow suggestions (and some suggestions are bad).

peternewman:
> But to me it seems case where removing it from replacement list is improving it.

Can you try and clarify why you think removing it improves things over just people who don't want to use it skipping it? I think there's probably a lower bar for entry for things like the usage "dictionary" compared to say the spelling ones, or it's likely to need a strong reason to remove stuff. If you point out a normal dictionary entry that corrects to another typo, it's a no-brainer to drop it, but like these and the informal ones, they're more of a style guide or a linter, you can choose to ignore some or all of the rules if you'd like.

Things like fireman (rather than firefighter) or master/slave rather than active/standby, may well have all been historically well known terms, but hopefully the reasoning for moving away from them is somewhat stronger and clearer?

>     * I am likely actually from group that is trying to be "protected" so I feel relatively OK with declaring that it seems weird to me (though there could be more complicated explanation  for that replacement that would nullify this and claim that also here I am from group not supposed to comment)

I think this is another case of trying to be gender neutral, apologies if I'm mistaken, but I assume from your name (and a quick Wikipedia check), you're a man, so I don't think you're the main target protected group for this particular replacement.

We could add person-in-the-middle too if you'd like, which would be just as clear (if not as well known)?

matkoniecz:
> Can you try and clarify why you think removing it improves things over just people who don't want to use it skipping it?

Mostly because it is requesting changes that I consider unreasonable, for political reasons which I do not entirely support and based on false claims. And goes into attempt to change language rather than to point out where I am diverging from standard use of a language.

https://tools.ietf.org/id/draft-knodel-terminology-00.html cited in justification for this change claims "It is not a standard term" what is clearly untrue.

---------------------------------------------------------------

In general I am not the greatest fan of attempting to stop using all words with any kind of problem or associated with any kind of problem.

It is probably caused by the fact that I am not a native English speaker and I am not entirely comfortable with English grammar, English vocabulary and may miss some problems. Adding on top "this word/phrase was declared racist/oppressive within last months" is not a welcome surprise and I would prefer if that trend would not accelerate. At the very least I do not want to support pressure to declare entirely fine terms as wrong and to be avoided.

There is no direct risk of that, but I worry that entire Polish language may be declared oppressive as it is grammatically gendered (word `rock` is grammatically male, word `cloud` female and word `child` is grammatically neuter). Maybe it is over the top worry but given that replacements that I would not expect 10 years ago are happening I am not really sure.

This makes "any gendered word is oppressive" trend worrying to me.

---------------------------------------------------------

And obviously that is political. Some would prefer ban more terms (`trigger warning`, `she`, `he`, `addict`, `victim` [1]) some would prefer to avoid altogether different classes for different reasons.

Which terms actually should be avoided highly depends on situation and is highly political issue.

[1] From "Oppressive Language List" that really does not seem to be parody/joke

DimitriPapadopoulos:
Note that `usage_dictionary.txt` is not a default dictionary and is "_for replacing phrasing with recommended terms_". Of course this does raise questions:
* Recommended by whom?

So yes, I understand your analysis. I believe (but may be wrong because I was not part of the process) that the initial intent of this dictionary was to apply the kind of rules you may not like. I guess there was enough demand for that at some point.

Perhaps the way out is not discussing individual cases, but providing general rules. We should probably clarify which of the sources cited in https://github.com/codespell-project/codespell/pull/2149#issuecomment-967791725 this dictionary applies, and stick to that. Those who agree may use that dictionary, others just skip it. Note that this is not a default dictionary.

DimitriPapadopoulos:
In the meantime, additional fixes have been added, that should at least avoid automatic changes.

PR #2150, opened 2021-11-12
better explain what code disctionary contains

matkoniecz:
https://github.com/codespell-project/codespell/pull/2144#issuecomment-966919948

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -58,7 +58,7 @@ _builtin_dictionaries = (
         True, True, supported_languages_en, supported_languages_en),
     ('usage', 'for replacing phrasing with recommended terms', '_usage',
         None, None, None, None),
-    ('code', 'for words common to code and/or mathematics that might be typos', '_code',  # noqa: E501
+    ('code', 'for words from code and/or mathematics that are likely typos in other context (such as uint)', '_code',  # noqa: E501
         None, None, None, None,),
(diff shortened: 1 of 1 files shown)


codespell_lib/_codespell.py
```diff
@@ -58,7 +58,7 @@
         True, True, supported_languages_en, supported_languages_en),
     ('usage', 'for replacing phrasing with recommended terms', '_usage',
         None, None, None, None),
-    ('code', 'for words common to code and/or mathematics that might be typos', '_code',  # noqa: E501
+    ('code', 'for words from code and/or mathematics that are likely typos in other context (such as uint)', '_code',  # noqa: E501
```
peternewman, line 61: s/context/contexts/

Thanks for this, I'm not sure if the example helps, as it's not necessarily a very strong example, and opens us up to the challenge of our typos appearing in our codebase if we wanted to check with the code dictionary.
```suggestion
    ('code', 'for words from code and/or mathematics that are likely to be typos in other contexts', '_code',  # noqa: E501
```

peternewman (changes requested):
I was going to ask you to add disctionary but I see it's already in the dictionary. :laughing:

matkoniecz:
> I was going to ask you to add disctionary but I see it's already in the dictionary. 😆

Well, let's say that there is a reason why I am interesting in typo detection

peternewman (approved):
LGTM thanks!

PR #2151, opened 2021-11-12
Also add mistmatch->mismatch

sylvestre:
We already have mistmatches->mismatches


vikivivi:
It would be great, if you can add these as well.

* https://github.com/search?q=mistmatched&type=code
* https://github.com/search?q=mistmatching&type=code

```text
mistmatched->mismatched
mistmatching->mismatching
```

vikivivi (approved):
LGTM

peternewman (approved):

PR #2158, opened 2021-11-16
oce->one

### LINKED ISSUE #2156: oce->one?
It appears that oce is not a word in English

https://en.wiktionary.org/wiki/oce

https://www.google.com/search?hl=en&q=%22oce%22

https://duckduckgo.com/?t=ffab&q=%22oce%22&ia=web

matkoniecz:
happened in my own code (near codespell-detected misspelling)

It appears that oce is not a word in English

https://en.wiktionary.org/wiki/oce

https://www.google.com/search?hl=en&q=%22oce%22

https://duckduckgo.com/?t=ffab&q=%22oce%22&ia=web

fixes #2156

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -20463,7 +20463,7 @@ occurrance->occurrence
 occurrances->occurrences
 occurrencs->occurrences
 occurrs->occurs
-oce->one
+oce->once, one, ore,
 oclock->o'clock
 ocntext->context
 ocorrence->occurrence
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -20463,6 +20463,7 @@ occurrance->occurrence
 occurrances->occurrences
 occurrencs->occurrences
 occurrs->occurs
+oce->one
```
peternewman, line 20466: It should definitely suggest `once`, probably first and possibly only (@matkoniecz that may also be the correction you should actually be using too). Otherwise it should suggest `ore` too, which is at least the same side of the keyboard.

peternewman (changes requested):

codespell_lib/data/dictionary.txt
matkoniecz: In my case it was definitely `one` (I do not remember exact commit or project, but it was definitely referring to number).

peternewman (approved):

PR #2159, opened 2021-11-16
Handle bad globs passed to if --skip/-S

### LINKED ISSUE #2145: bad character range in a file skip glob
> Also, as result `codespell --skip "example file with actual [weird signs in it and b-a]"` goes down with
> 
> ```
> Traceback (most recent call last):
>   File "/home/mateusz/.local/bin/codespell", line 8, in <module>
>     sys.exit(_script_main())
>   File "/home/mateusz/.local/lib/python3.8/site-packages/codespell_lib/_codespell.py", line 748, in _script_main
>     return main(*sys.argv[1:])
>   File "/home/mateusz/.local/lib/python3.8/site-packages/codespell_lib/_codespell.py", line 876, in main
>     if glob_match.match(root):  # skip (absolute) directories
>   File "/home/mateusz/.local/lib/python3.8/site-packages/codespell_lib/_codespell.py", line 108, in match
>     if fnmatch.fnmatch(filename, p):
>   File "/usr/lib/python3.8/fnmatch.py", line 36, in fnmatch
>     return fnmatchcase(name, pat)
>   File "/usr/lib/python3.8/fnmatch.py", line 70, in fnmatchcase
>     match = _compile_pattern(pat)
>   File "/usr/lib/python3.8/fnmatch.py", line 46, in _compile_pattern
>     return re.compile(res).match
>   File "/usr/lib/python3.8/re.py", line 252, in compile
>     return _compile(pattern, flags)
>   File "/usr/lib/python3.8/re.py", line 304, in _compile
>     p = sre_compile.compile(pattern, flags)
>   File "/usr/lib/python3.8/sre_compile.py", line 764, in compile
>     p = sre_parse.parse(p, flags)
>   File "/usr/lib/python3.8/sre_parse.py", line 948, in parse
>     p = _parse_sub(source, state, flags & SRE_FLAG_VERBOSE, 0)
>   File "/usr/lib/python3.8/sre_parse.py", line 443, in _parse_sub
>     itemsappend(_parse(source, state, verbose, nested + 1,
>   File "/usr/lib/python3.8/sre_parse.py", line 834, in _parse
>     p = _parse_sub(source, state, sub_verbose, nested + 1)
>   File "/usr/lib/python3.8/sre_parse.py", line 443, in _parse_sub
>     itemsappend(_parse(source, state, verbose, nested + 1,
>   File "/usr/lib/python3.8/sre_parse.py", line 598, in _parse
>     raise source.error(msg, len(this) + 1 + len(that))
> re.error: bad character range b-a at position 56
> ```

_Originally posted by @matkoniecz in https://github.com/codespell-project/codespell/issues/2142#issue-1050634961

While that glob obviously won't work, we probably shouldn't crash quite so unhelpfully and instead catch the error...

DimitriPapadopoulos:
Fixes #2145.

The downside is that we catch the `re.error` exception when `glob_match` is actually used. We cannot tell which exact glob fails because `glob_match` is initialized with a **list** of globs:
https://github.com/codespell-project/codespell/blob/0ee668811694e2f6a0df4612ffc4e9d4e447b04d/codespell_lib/_codespell.py#L864

The alternative would be to run `fnmatch.fnmatch()` on each item in `options.skip`, before creating a global `GlobMatch` for all items. We will suppose that the global `GlobMatch` instance cannot fail if none of the simple `fnmatch.fnmatch()` calls fails.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -898,7 +898,15 @@ def main(*args):
 
     file_opener = FileOpener(options.hard_encoding_detection,
                              options.quiet_level)
+
     glob_match = GlobMatch(options.skip)
+    try:
+        glob_match.match("/random/path")  # does not need a real path
+    except re.error:
+        print("ERROR: --skip/-S has been fed an invalid glob, "
+              "try escaping special characters",
+              file=sys.stderr)
+        return EX_USAGE
 
     bad_count = 0
     for filename in options.files:
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -117,8 +117,28 @@ def test_basic(tmpdir, capsys):
     assert cs.main(d) == 0
 
     # empty directory
-    os.mkdir(op.join(d, 'test'))
+    os.mkdir(op.join(d, 'empty'))
     assert cs.main(d) == 0
+
+
+def test_bad_glob(tmpdir, capsys):
+    # disregard invalid globs, properly handle escaped globs
+    g = op.join(tmpdir, 'glob')
+    os.mkdir(g)
+    fname = op.join(g, '[b-a].txt')
+    with open(fname, 'a') as f:
+        f.write('abandonned\n')
+    assert cs.main(g) == 1
+    # bad glob is invalid
+    code, _, stderr = cs.main('--skip', '[b-a].txt',
+                              g, std=True)
+    if sys.hexversion < 0x030A05F0:  # Python < 3.10.5 raises re.error
+        assert code == EX_USAGE, 'invalid glob'
+        assert 'invalid glob' in stderr
+    else:  # Python >= 3.10.5 does not match
+        assert code == 1
+    # properly escaped glob is valid, and matches glob-like file name
+    assert cs.main('--skip', '[[]b-a[]].txt', g) == 0
 
 
 @pytest.mark.skipif(


DimitriPapadopoulos:
Because `fnmatch` uses the `re` module under the hood, it may raise cryptic exceptions such as `sre_constants.error`. These exceptions are subclasses of `re.error` and should be caught using this superclass of `re`-related exception.

The documentation is unclear, it should explicitly mention the type of exceptions that might be raised. I've opened a Python bug:
https://bugs.python.org/issue45815

codespell_lib/_codespell.py
```diff
@@ -862,7 +862,14 @@ def main(*args):
 
     file_opener = FileOpener(options.hard_encoding_detection,
                              options.quiet_level)
+
     glob_match = GlobMatch(options.skip)
+    try:
+        glob_match.match("/random/path")  # does not need a real path
+    except re.error:
+        print("ERROR: --skip/-S has been fed an invalid glob",
```
peternewman, line 870: I don't know if it's worth extending it to something like:
```suggestion
        print("ERROR: --skip/-S has been fed an invalid glob, try escaping special characters",
```

On the assumption a lot of people will just be doing `--skip foo.txt,bar.txt` etc and their filename just happens to have square brackets in it.

peternewman (approved):
Just one possible tweak.

codespell_lib/_codespell.py
```diff
@@ -862,7 +862,14 @@ def main(*args):
 
     file_opener = FileOpener(options.hard_encoding_detection,
                              options.quiet_level)
+
     glob_match = GlobMatch(options.skip)
+    try:
+        glob_match.match("/random/path")  # does not need a real path
+    except re.error:
+        print("ERROR: --skip/-S has been fed an invalid glob, try escaping special characters",
```
peternewman, line 870: Can you fix the lint error.

codespell_lib/tests/test_basic.py
```diff
@@ -99,6 +99,10 @@ def test_basic(tmpdir, capsys):
     assert 'cannot find dictionary' in stderr
     os.remove(fname)
 
+    code, _, stderr = cs.main('--skip', 'file with [weird signs b-a].txt',
+                              f.name, std=True)
+    assert code == EX_USAGE, 'invalid glob'
```
peternewman, line 104: It would be nice to have the counter test that it passes when the escaping has been added!

peternewman (changes requested):
One obvious thing to fix and one that I didn't think of before sorry.

codespell_lib/tests/test_basic.py
peternewman: Just this please @DimitriPapadopoulos .

DimitriPapadopoulos: I'm afraid I don't understand the request here. Which counter test are you referring to? Perhaps `count=True` in `MainWrapper.main()`?
https://github.com/codespell-project/codespell/blob/bb7d185a71acbe651608bf3527c1ac10cebe4ddf/codespell_lib/tests/test_basic.py#L25-L28

peternewman: Sorry, I mean anti/counter, i.e. can you successfully skip a file by escaping that character class's brackets.

DimitriPapadopoulos: Mmmh.... I doubt casual users are able to escape the brackets. I had to read [`glob(7)`](https://man7.org/linux/man-pages/man7/glob.7.html) attentively:
> #### Character classes
>  An expression "`[...]`" where the first character after the leading '`[`' is not an '`!`' matches a single character, namely any of the characters enclosed by the brackets.  The string enclosed by the brackets cannot be empty; therefore '`]`' can be allowed between the brackets, provided that it is the first character.  (Thus, "`[][!]`" matches the three characters '`[`', '`]`', and '_!_'.)

to understand how to escape the brackets:
```python
>>> import fnmatch
>>> 
>>> filename = '[b-a].txt'
>>> 
>>> fnmatch.fnmatch(filename, '[[]b-a[]].txt')
True
>>> 
>>> fnmatch.fnmatch(filename, '[b-a].txt')
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/fnmatch.py", line 36, in fnmatch
    return fnmatchcase(name, pat)
  File "/usr/lib/python3.8/fnmatch.py", line 70, in fnmatchcase
    match = _compile_pattern(pat)
  File "/usr/lib/python3.8/fnmatch.py", line 46, in _compile_pattern
    return re.compile(res).match
  File "/usr/lib/python3.8/re.py", line 252, in compile
    return _compile(pattern, flags)
  File "/usr/lib/python3.8/re.py", line 304, in _compile
    p = sre_compile.compile(pattern, flags)
  File "/usr/lib/python3.8/sre_compile.py", line 764, in compile
    p = sre_parse.parse(p, flags)
  File "/usr/lib/python3.8/sre_parse.py", line 948, in parse
    p = _parse_sub(source, state, flags & SRE_FLAG_VERBOSE, 0)
  File "/usr/lib/python3.8/sre_parse.py", line 443, in _parse_sub
    itemsappend(_parse(source, state, verbose, nested + 1,
  File "/usr/lib/python3.8/sre_parse.py", line 834, in _parse
    p = _parse_sub(source, state, sub_verbose, nested + 1)
  File "/usr/lib/python3.8/sre_parse.py", line 443, in _parse_sub
    itemsappend(_parse(source, state, verbose, nested + 1,
  File "/usr/lib/python3.8/sre_parse.py", line 598, in _parse
    raise source.error(msg, len(this) + 1 + len(that))
re.error: bad character range b-a at position 5
>>> 
```

DimitriPapadopoulos: Test with escaped brackets works now.

DimitriPapadopoulos: @peternewman I have applied the requested change.

DimitriPapadopoulos:
@peternewman I think all fixes have been applied. Can you have a look?

DimitriPapadopoulos:
Also, Python has been fixed, future versions (>= 3.10.5) should not crash:
https://github.com/python/cpython/issues/89973

larsoner (approved):

PR #2160, opened 2021-11-16
`print("ERROR: ...")` should print to `sys.stderr`

DimitriPapadopoulos:


peternewman (approved):
LGTM, it matches grep's behaviour and makes us internally consistent.

PR #2172, opened 2021-11-20

luzpaz:


matkoniecz:
natch is a slang, but still a word

https://en.wiktionary.org/wiki/natch https://www.merriam-webster.com/dictionary/natch

PR #2175, opened 2021-11-24
Add explicete*

sebweb3r:


codespell_lib/data/dictionary.txt
```diff
@@ -13254,6 +13254,9 @@ explecitily->explicitly
 explecitly->explicitly
 explenation->explanation
 explicat->explicate
+explicete->explicit
+explicetely->explicitly
+explicetly->explicitly
 explicilt->explicit
 explicilty->explicitly
 explicite->explicit, explicitly,
```
peternewman, line 13262: It might be good to be consistent, does explicite often actually mean explicitly, in which case explicete probably would too, or drop it from both.

sebweb3r: People mix up adjectives and adverbs quite often. Shall I add both variants to all suggestions?

luzpaz: > Shall I add both variants to all suggestions?  

Yes, that makes sense

PR #2176, opened 2021-11-24
Another "sensitive" misspelling

DimitriPapadopoulos:
See https://github.com/zarr-developers/zarr-python/pull/880.

https://github.com/search?q=case-sensitive&type=code


peternewman (commented):
Given there appear to be seven typos for sensitive, but only three are in as case-sensitive, is it worth pre-emptively adding the ones now?

DimitriPapadopoulos:
Added other variations, consistently for `sensitive`/`insensitive` and `case-sensitive`/`case-insensitive`.

luzpaz (approved):

PR #2179, opened 2021-11-27
Misspelling: variations on standari[sz]e

DimitriPapadopoulos:
See https://github.com/Blosc/c-blosc/pull/326.


peternewman (changes requested):
Thanks, can you add the ise variants too please.

Also standardi[sz]ation versions.

peternewman (approved):

PR #1107, opened 2019-05-07
Removed 'compliancy' from dictionary

luzpaz:
It's a legit [word](https://www.merriam-webster.com/thesaurus/compliancy)
Once merged, please close #1062

Issue #1062, opened 2019-04-06
'compliancy' is a valid word


luzpaz:
https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary.txt#L3043

https://www.merriam-webster.com/thesaurus/compliancy

peternewman:
I'm not sure this needs an issue @luzpaz, if it's a valid word, it just needs fixing.

I'm beginning to wonder if we should validate the alleged typos in the Codespell dictionary against a word list of valid spellings during CI, as things like this seem to keep creeping in when people source typos from existing codebases (where an infrequently used but valid word may be a typo for another, also valid, word).

luzpaz:
> I'm beginning to wonder if we should validate the alleged typos in the Codespell dictionary against a word list of valid spellings during CI, as things like this seem to keep creeping in when people source typos from existing codebases

Yes, you're on to something here.

PR #2184, opened 2021-11-30
Move 'compliancy' into dictionary_rare.

cfi-gb:
This valid word has been removed in the past in #1107 based on #1062 but it was added again since then, ~~just removing it again.~~  moving into the rare dictionary.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -6410,7 +6410,6 @@ complette->complete
 complettly->completely
 complety->completely
 compliace->compliance
-compliancy->compliance
 complianse->compliance
 complied-in->compiled-in
 complience->compliance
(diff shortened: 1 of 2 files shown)


peternewman (changes requested):
Can I suggest moving to rare, because it is:
https://books.google.com/ngrams/graph?content=compliancy%2Ccompliance&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Ccompliancy%3B%2Cc0%3B.t1%3B%2Ccompliance%3B%2Cc0

Also it will stop people adding it back in each time!

cfi-gb:
> Can I suggest moving to rare, because it is: https://books.google.com/ngrams/graph?content=compliancy%2Ccompliance&year_start=1800&year_end=2019&corpus=26&smoothing=3&direct_url=t1%3B%2Ccompliancy%3B%2Cc0%3B.t1%3B%2Ccompliance%3B%2Cc0
> 
> Also it will stop people adding it back in each time!

Sounds good. I have updated the PR accordingly.

peternewman (approved):

PR #2186, opened 2021-12-01
Add various spelling correction variants for "upgrade".

cfi-gb:
Seems for various existing entries some of the (e|es|ed|ing) variants are currently missing. Hope i was able to catch all of them correctly.

Newly added `upgarad(e|es|ed|ing)` seems to be not that common but these are still happening so might worth to add:

https://grep.app/search?q=Upgarade


peternewman (approved):

peternewman:
Thanks @cfi-gb . I don't know if you fancy covering downgrade etc too?

cfi-gb:
> Thanks @cfi-gb . I don't know if you fancy covering downgrade etc too?

@peternewman Sure :+1: . Took a wile but here we go: #2205

PR #2187, opened 2021-12-02
Move `hist` from rare to code

### LINKED ISSUE #2185: Misidentifying a matplotlib function as spelling error
Add `plt.hist` anywhere in your code and you'll get the following correction suggestion:

```
hist ==> heist, his
```

The fact that `codespell -w` would overwrite actual code seems highly problematic to me (this is a bad example because `-w` won't do anything in this particular case since there are two possibilities).

To reproduce:

```py
$ mkdir foo; cd foo; echo "import matplotlib.pyplot as plt;plt.hist([0,1,1]);" >> foo.py; codespell .
./foo.py:1: hist ==> heist, his
```

DimitriPapadopoulos:
The rationale is that while `hist` is a [valid exclamation](https://www.lexico.com/definition/hist), it is foremost the widely used [`matplotlib.pyplot.hist`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.hist.html) function, used by almost all scientific Python software that require visualization.

Fixes #2185.

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -18,6 +18,7 @@ errorstring->error string
 exitst->exits, exists,
 files'->file's
 gae->game, Gael, gale,
+hist->heist, his,
 iam->I am, aim,
 iff->if
 ith->with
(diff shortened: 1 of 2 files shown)


peternewman (approved):

peternewman:
Thanks @DimitriPapadopoulos

PR #2195, opened 2021-12-11
Fix uncaught exception on empty files

DimitriPapadopoulos:
Fixes #2194.

PR #2196, opened 2021-12-11
Avoid bailing out with uncaught `PermissionError`

DimitriPapadopoulos:


DimitriPapadopoulos:
I need to rebase this one on #2195 to fix the new exception raised in tests.

I suggest we focus on #2195 first, then I can rebase this one.

luzpaz:
Tests failing

DimitriPapadopoulos:
Yes, I can try that.

DimitriPapadopoulos:
@larsoner You never know:
* `is_text_file` calls `open(filename, mode='rb')` and the part you have removed calls `file_opener.open(filename)`. While they should behave 100% the same with respect to access rights, I have no proof of that.
* One could run `chmod` on spell checked files, concurrently to `codespell`. This is hard to test of course, and should never happen in practice. But I'd rather be on the safe side.

larsoner:
If those are the reasons I'd rather keep things simple and assume YAGNI rather than as more stuff to our code for such deep corner cases. If we try to be so complete everywhere our codebase, maintainability/interpretability will go down for very little (in practice, maybe no actual) gain

PR #2218, opened 2021-12-31

luzpaz:


vikivivi:
Misspell `al` is valid. Example `et al`

luzpaz:
Maybe I should move it to the rare dictionary then?

vikivivi:
I am not sure moving to `rare` will able to get those integration checks to pass.

PR #2227, opened 2022-01-05
Add many tests for `--skip`

DimitriPapadopoulos:
These tests shows what needs to be fixed in the current implementation and will help avoid regressions while fixing.

Will help test and review #2058.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -15820,6 +15820,10 @@ higlights->highlights
 higly->highly
 higth->height
 higway->highway
+hijkack->hijack
+hijkacked->hijacked
+hijkacking->hijacking
+hijkacks->hijacks
 hilight->highlight
 hilighted->highlighted
 hilighting->highlighting


peternewman:
> I'm puzzled by this test:
> 
> https://github.com/codespell-project/codespell/blob/bdcf7a62a2152be3adc4fea27dfb0780bd1efa18/codespell_lib/tests/test_basic.py#L308
> 
> Is `*ignoredir/bad*` really supposed to match `tempdir/ignoredir/subdir/bad.txt`?

No, the old/current code actually generates:
`tempdir/ignoredir/bad.txt`

But the fact it passes shows how bad the current skip matching is!

codespell_lib/tests/test_basic.py
```diff
@@ -285,7 +285,19 @@ def test_encoding(tmpdir, capsys):
...
+    """Test ignoring of files and directories.
+
+    Test the following file hierarchy:
+
+        tmpdir
+        ├── good.txt
+        ├── bad.txt
+        ├── ignoredir
+        │   └── subdir
+        │       └── bad.txt
```
peternewman, line 297: So the current is this:
```suggestion
        │   └── bad.txt
```

Do we need another level to test some more edge cases?

DimitriPapadopoulos: I'd rather put `bad.txt` under `subdir` (which I was certain it was :smiley:) to get that extra level.

I believe that would catch all possible errors.

DimitriPapadopoulos: This time I've **really** changed the file hierarchy to:
```
        tmpdir
        ├── good.txt
        ├── bad.txt
        ├── ignoredir
        │   └── subdir
        │       └── bad.txt
        └── bad.js
```
I think we have enough tests to proceed with the changes without regressions.

DimitriPapadopoulos: The current skip matching isn't that bad. The 3 main issues are:
* handle `--skip=ignoredir/` with a trailing `/`
* skip relative paths when codespell is called on a relative path, for example handle  `--skip=my/path` in addition to `--skip=./my/path` when codespell is called on the current directory
* skip absolute paths when codespell is called on a relative path, for example handle `--skip=/absolute/path` when codespell is called on a relative path (probably the least important issue)

peternewman: Do we want another bad.txt in ignoredir too?

peternewman: > The current skip matching isn't that bad.

Yeah I think it was actually just your tofix labelling. 😄 

> The 3 main issues are:
> 
> * handle `--skip=ignoredir/` with a trailing `/`
> * skip relative paths when codespell is called on a relative path, for example handle  `--skip=my/path` in addition to `--skip=./my/path` when codespell is called on the current directory
> * skip absolute paths when codespell is called on a relative path, for example handle `--skip=/absolute/path` when codespell is called on a relative path (probably the least important issue)

Sounds good.

DimitriPapadopoulos: We could add a `bad.txt` in the 1st level subdirectory. However, we know when we have succeeded ignoring `ignnoredir` because `ignnoredir/subdir/bad.txt` will be ignored too. Furthermore, if ignoring a file works both for `bad.txt` in the top directory and `ignoredir/subdir/bad.txt` in the 2nd level subdirectory, I think testing in the 1st level subdirectory too is far-fetched. We could do that of course, but then why not test a 3rd level, and a 4th level, and...

peternewman: Although we can't currently do `ignoredir/subdir` test can we? To prove it doesn't hit/miss the one within ignoredir?

peternewman: We probably also want another file in the subdirectories, maybe with another typo, or unique typos per file or something, so we can confirm we aren't just ignoring everything in ignoredir/subdir when we are just trying to block bad.txt within there.

codespell_lib/tests/test_basic.py
```diff
@@ -297,15 +309,67 @@ def test_ignore(tmpdir, capsys):
...
+    ignoredir = op.join(d, 'ignoredir')
+    os.mkdir(ignoredir)
+    subdir = op.join(ignoredir, 'subdir')
     os.mkdir(subdir)
     with open(op.join(subdir, 'bad.txt'), 'w') as f:
         f.write('abandonned')
     assert cs.main(d) == 2
     assert cs.main('--skip=bad*', d) == 0
-    assert cs.main('--skip=*ignoredir*', d) == 1
+    assert cs.main('--skip=*ad*', d) == 0
```
peternewman, line 320: Sort of the same with this, if it's actually gone wrong and just matched `*`, we'd be none the wiser.

Perhaps bad and badder 😆 .

DimitriPapadopoulos:
I have mistakenly erased this branch. Need to rewrite it, taking into account your remarks. Sorry.

peternewman:
> I have mistakenly erased this branch. Need to rewrite it, taking into account your remarks. Sorry.

Oops.

We should probably also have some files and **folders** like:
```
bad
1bad
bad1
```

So we know it's not matched the files/folders we didn't intend to match.

PR #2234, opened 2022-01-12
Mentioned .editorconfig

polluks:


peternewman (commented):
That works, although you could also add it to https://github.com/codespell-project/codespell/blob/618dda35977c58ecc7e8f4b57f6591315c2781df/codespell_lib/_codespell.py#L403 and we'd support it natively!

peternewman:
@polluks is there some reason not to add it to our config so it's just supported natively in Codespell?

PR #2236, opened 2022-01-13
Add various Underflow / Overflow spelling corrections.

cfi-gb:
The `oberflow` typo doesn't happen that much:

https://grep.app/search?q=oberflow

But i guess the typo might happen because of the `b` next to the `v` on keyboards so still could make sense to add it.

Other `s`, `ed` and `ing` variants got taken / extended their existing variants, hope i got everything correctly.

The "rare" placement was done based on a suggestion in https://github.com/codespell-project/codespell/pull/2236#discussion_r788300427

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -21050,6 +21050,10 @@ obejct->object
 obeject->object
 obejection->objection
 obejects->objects
+oberflow->overflow
+oberflowed->overflowed
+oberflowing->overflowing
+oberflows->overflows
 obervation->observation
 obervations->observations
 oberve->observe
@@ -21207,7 +21211,6 @@ oder->order, odor,
 odly->oddly
 ody->body
 oen->one
-oerflow->overflow
 ofcource->of course
 offcers->officers
 offcial->official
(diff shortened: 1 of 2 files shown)


luzpaz (approved):

codespell_lib/data/dictionary.txt
```diff
@@ -21154,6 +21158,9 @@ odly->oddly
 ody->body
 oen->one
 oerflow->overflow
+oerflowed->overflowed
```
peternewman, line 21161: This was old English, and is used in a few poems so perhaps rare? Maybe for all of those versions?

codespell_lib/data/dictionary.txt
```diff
@@ -21780,11 +21790,20 @@ overlaping->overlapping
...
+overlflows->overflows
 overlfow->overflow
+overlfowed->overflowed
+overlfowing->overflowing
+overlfows->overflows
 overlodaded->overloaded
 overloded->overloaded
 overlodes->overloads
 overlow->overflow
+overlowed->overflowed
```
peternewman, line 21804: Possibly overloaded at a stretch.

codespell_lib/data/dictionary.txt
cfi-gb: I'm fine with that, pushed a new version of the PR with this suggestion applied.

codespell_lib/data/dictionary.txt
cfi-gb: Pushed a new version of the PR with this suggestion applied.

peternewman (approved):

PR #2246, opened 2022-01-17
add tenatious->tenacious

sed-i:


codespell_lib/data/dictionary.txt
```diff
@@ -30481,6 +30481,7 @@ tenacle->tentacle
 tenacles->tentacles
 tenanet->tenant
 tenanets->tenants
+tenatious->tenacious
```
luzpaz, line 30484: Please also add `tenatiously->tenaciously`

sed-i: Done

PR #2248, opened 2022-01-19
Add more quest variants

peternewman:
Generated from request typos and some real world

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -24339,7 +24339,6 @@ querries->queries
 queryies->queries
 queryinterace->queryinterface
 querys->queries
-ques->quest
 queset->quest
 quesets->quests
 quesiton->question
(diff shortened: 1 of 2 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -24289,8 +24289,16 @@ pytnon->python
 pytohn->python
 pyton->python
 pytyon->python
+qest->quest
+qests->quests
+qeuest->quest
```
luzpaz, line 24294: please add `quests` as another answer

peternewman: I'd better fix the request one too!

peternewman: Actually sorry why @luzpaz , I don't really see why this suggests quests rather than the singular. I've got the plural versions of those typos in already.

luzpaz: You're right. My mistake.

luzpaz (approved):

PR #2257, opened 2022-01-27
lod=level of detail

### LINKED ISSUE #2256: LOD (Level Of Detail) should not be treated as an error

matkoniecz:
fixes #2256

I agree that it is not the most common acronym, but I prefer codespell to not have false positives (unless user opts-in)

### DIFF
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -91,6 +91,7 @@ leaded->led, lead,
 leas->least, lease,
 lightening->lightning, lighting,
 loafing->loading
+lod->load
 loos->loose, lose,
 loosing->losing
 lousily->loosely
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -18823,7 +18823,6 @@ locatins->locations
 loccked->locked
 locgical->logical
 lockingf->locking
-lod->load
 loding->loading
 loev->love
 logarithmical->logarithmically


peternewman:
> I agree that it is not the most common acronym, but I prefer codespell to not have false positives (unless user opts-in)

Note that code is the one not included by default:
https://github.com/codespell-project/codespell/blob/618dda35977c58ecc7e8f4b57f6591315c2781df/codespell_lib/_codespell.py#L68

peternewman (approved):
LGTM, up to @matkoniecz if he wants to move it to code instead...

matkoniecz:
> Note that code is the one not included by default:

I know. I put it into rare as I have seen it discussed in computer graphics/computer games outside of what is considered as programming.

peternewman:
> > Note that code is the one not included by default:
> 
> I know. I put it into rare as I have seen it discussed in computer graphics/computer games outside of what is considered as programming.

Okay, that does mean the correction will still be offered by default in all cases (unless people explicitly turn off the rare dictionary).

matkoniecz:
I think I opted out (or will) exactly because it may contain false positives.

codespell_lib/data/dictionary_rare.txt
```diff
@@ -91,6 +91,7 @@ leaded->led, lead,
 leas->least, lease,
 lightening->lightning, lighting,
 loafing->loading
+lod->load
```
peternewman, line 94: Actually a bit late (having just merged), but suggesting loud would also be good, as well as stopping any auto-fixing...

PR #2259, opened 2022-01-27
Add spelling corrections for template and variants.

cfi-gb:
See e.g.:

https://grep.app/search?q=templace&words=true

As there is https://github.com/ryanve/templace i wasn't absolutely sure about the `templace->template` and have put it into the code dictionary for now.


peternewman (approved):

peternewman:
> As there is https://github.com/ryanve/templace i wasn't absolutely sure about the `templace->template` and have put it into the code dictionary for now.

Perfect @cfi-gb I think that's the safest option!

PR #2286, opened 2022-02-23
Move ba->by, be, as it's a name

peternewman:
As it's a name:
https://en.wikipedia.org/wiki/Ba_(given_name)

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3777,7 +3777,6 @@ axpressed->expressed
 aysnc->async
 ayway->anyway, away,
 ayways->always
-ba->by, be,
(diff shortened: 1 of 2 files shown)


luzpaz:
TBH, i think it's useless to have `ba` because I'm constantly needing to whitelist it.

peternewman:
> TBH, i think it's useless to have `ba` because I'm constantly needing to whitelist it.

Well it will be disabled by default now its in names. But it is also a typo.

peternewman:
Do you want to approve then @luzpaz ?

luzpaz (approved):

luzpaz:
Sorry about that. Approved.

PR #2290, opened 2022-02-24
Fix #2055: Add support for PEP 518

### LINKED ISSUE #2055: PEP 518 Support pyproject.toml for configuration
It would be nice if this project supported reading configuration from `pyproject.toml` files.

INI files read by `configparser` don't have a well defined syntax, and PEP 518 is supported by a lot of other Python linting tools.

---
https://www.python.org/dev/peps/pep-0518/#tool-table

Freed-Wu:
Fix #2055

### DIFF
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -16,10 +16,10 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - 3.7
-          - 3.8
-          - 3.9
-          - 3.10
+          - '3.7'
+          - '3.8'
+          - '3.9'
+          - '3.10'
     name: Python ${{ matrix.python-version }} test
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -2,6 +2,9 @@
 # For general usage in your repo, see the example in codespell.yml
 # https://github.com/codespell-project/codespell
 name: codespell Private Actions
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
+  cancel-in-progress: true
 on: [push, pull_request]
 jobs:
   test:
@@ -10,12 +13,13 @@ jobs:
     # Make sure we're using the latest aspell dictionary
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false
       matrix:
         python-version:
-          - 3.6
           - 3.7
           - 3.8
           - 3.9
+          - 3.10
     name: Python ${{ matrix.python-version }} test
     steps:
       - uses: actions/checkout@v3
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -392,7 +392,8 @@ def parse_options(args):
                         help='print LINES of surrounding context')
     parser.add_argument('--config', type=str,
                         help='path to config file.')
-
+    parser.add_argument('--toml', type=str,
+                        help='path to a pyproject.toml file.')
     parser.add_argument('files', nargs='*',
                         help='files or directories to check')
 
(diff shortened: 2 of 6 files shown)


Freed-Wu:
I import toml as a new dependence. In fact, it looks like it can be an optional dependence, because we can:

```python
# Read toml before other config files.
if os.path.isfile(os.path.realpath(tomlfile)):
    import toml
    data = toml.load(tomlfile).get('tool', {})
    config.read_dict(data)
```

I am not sure if it is a good idea.

DimitriPapadopoulos:
To avoid the current error in tests, you need to add `toml` to dependencies in `setup.py`, or make it optional as you suggested.

Freed-Wu:
```
Added lines #L411 - L413 were not covered by tests
```

Need change test file to cover the if-clause.

setup.py
```diff
@@ -64,5 +64,6 @@
               "dev": ["check-manifest", "flake8", "pytest", "pytest-cov",
                       "pytest-dependency"],
               "hard-encoding-detection": ["chardet"],
+              "toml": ["toml"],
```
cclauss, line 67: Given https://docs.python.org/3.11/library/tomllib.html why not use https://pypi.org/project/tomli instead of https://pypi.org/project/toml ?

Freed-Wu: Okay. :+1:

Freed-Wu:
Have rebased.

cclauss (approved):

larsoner:
Sorry for the slow response @Freed-Wu , I just enabled GH actions tests to actually run. Hopefully that helps :)

codespell_lib/_codespell.py
```diff
@@ -401,9 +401,17 @@ def parse_options(args):
...
     if options.config:
         cfg_files.append(options.config)
     config = configparser.ConfigParser()
+
+    # Read toml before other config files.
+    if os.path.isfile(os.path.realpath(tomlfile)):
+        import tomli
+        with open(tomlfile, 'rb') as f:
+            data = tomli.load(f).get('tool', {})
+        config.read_dict(data)
```
larsoner, line 414: To actually test this, you can add some new test using a `tmp_path` fixture and add a `pyproject.toml` file in that dir, and run `codespell` from that temporary directory. If it's not clear, let me know and I can push a commit

Freed-Wu: I have read <https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html?highlight=tmp_path> but I still don't know how to realize it. Can you push a commit? Thanks.

larsoner:
Okay I added a test. I also bumped the min Python version to 3.7 since 3.6 has been EOL for 9 months.

@cclauss has approved already so I'll mark for merge-when-green, thanks in advance @Freed-Wu !

larsoner:
... I started to add a `print('WARNING: ` when pyproject.toml is found but tomli is not installed, in case there are people using `codespell` who do have such a toml file (but no `tomli` in their codespell CI run) -- which I think will be fairly common/likely.

Then I realized that just checking for `pyproject.toml` in the `cwd` is a bit fragile. So I changed it so that users need to specify `--toml pyproject.toml` or whatever filename they want to load from. This allows for absolute paths to other directories for example if they want. It's more explicit which seems safer. And now we can just `raise ImportError` if `tomli` is not found, because if a user passes `--toml` they *should* get an error if `tomli` is not present (they are unambiguously trying to do something not allowed).

peternewman (commented):
@larsoner while I can see the argument for not requiring it, it does seem like one of these things where you'd hope it will just work, and it's a bit of a shame if you have to explicitly specify it.

Can we not hijack whatever searching configparser does for setup.cfg and .codespellrc so it behaves the same as them (and maybe throws a warning if the file is present but the module unavailable or something?

.github/workflows/codespell-private.yml
```diff
@@ -2,6 +2,9 @@
 # For general usage in your repo, see the example in codespell.yml
 # https://github.com/codespell-project/codespell
 name: codespell Private Actions
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}-${{ github.event.ref }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
larsoner: fail-fast cancels jobs within an action if one of them fails

this cancels jobs if another commit is pushed

larsoner:
> If the current working directory contains a pyproject.toml and that file contains [codespell] configuration, can I just run codespell, or am I required to run codespell --toml? If the latter, will I be required to also add --toml to codespell commands run by Poetry, Flit, pre-commit, tox, etc.?

Currently you'll be required to add the `--toml` arg. But if poetry etc. guarantee you're in the dir containing `pyproject.toml` when running `codespell` this is a good reason to autodetect and use it. I can make a quick PR for it

cclauss:
My sense is that Poetry, Flit, Hatch, and pdm ___require___ a `pyproject.toml` file.

.github/workflows/codespell-private.yml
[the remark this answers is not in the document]
larsoner: I think that our tests are fast enough already and as we add others (different OSes etc) they can potentially show different problems. But we reenable it in if you want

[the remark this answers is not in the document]
peternewman: Presumably it won't start testing the next commit until the last one is done?

PR #2294, opened 2022-02-26
Add hearbeat->heartbeat, heart beat,

peternewman:
Add variants

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -15866,6 +15866,9 @@ headquaters->headquarters
 headrom->headroom, bedroom,
 heaer->header
 healthercare->healthcare
+hearbeat->heartbeat, heart beat,
+hearbeating->heartbeating, heart beating,
+hearbeats->heartbeats, heart beats,
 heared->heard, header,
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -15866,6 +15866,9 @@ headquaters->headquarters
 headrom->headroom, bedroom,
 heaer->header
 healthercare->healthcare
+hearbeat->heartbeat, hear beat, heart beat,
+hearbeating->heartbeating, hear beating, heart beating,
+hearbeats->heartbeats, hear beats, heart beats,
```
luzpaz, line 15871: I think the `hear beat{s|ing}` is a stretch but if you're adamant about it, sure...I'm OK with it. Though, I would move all the those variants to the end (3rd option) because I'd bet the higher probability would be the user meant `heart ...`

peternewman: I'll admit they're rare, I did find one example on grep.app for hear beats, but admittedly the rest were just typos for heartbeat.

It's partly to avoid causing errors if we auto-correct with only one suggestion when others could also be valid.

PR #2302, opened 2022-03-02
Add spelling corrections for validate and variants.

cfi-gb:
See e.g.:
- https://github.com/search?q=Invalidtion&type=code
- https://github.com/search?q=Validtion&type=code

plus the suggestions given by @vikivivi in:

https://github.com/codespell-project/codespell/pull/2302#issuecomment-1058738530

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -18172,6 +18172,11 @@ invalde->invalid
 invalidatiopn->invalidation
 invalide->invalid
 invalidiate->invalidate
+invalidte->invalidate
+invalidted->invalidated
+invalidtes->invalidates
+invalidting->invalidating
+invalidtion->invalidation
 invalud->invalid
 invarient->invariant
 invarients->invariants
@@ -33103,6 +33108,11 @@ validaterelase->validaterelease
 valide->valid
 valididty->validity
 validing->validating
+validte->validate
+validted->validated
+validtes->validates
+validting->validating
+validtion->validation
 valied->valid
 valies->values
 valif->valid


luzpaz (approved):

vikivivi:
Shall we add these misspellings as well?
```text
invalidted->invalidated
invalidting->invalidating
validted->invalidated
validting->invalidating
```
* https://github.com/search?q=invalidted&type=code
* https://github.com/search?q=invalidting&type=code
* https://github.com/search?q=validted&type=code
* https://github.com/search?q=validting&type=code

luzpaz:
ping @cfi-gb

cfi-gb:
I have seen this question but don't have any opinion on that and thus kept silent as it also wasn't directed to me. It would need to be answered by some one else.

vikivivi:
> I have seen this question but don't have any opinion on that and thus kept silent as it also wasn't directed to me. It would need to be answered by some one else.

@cfi-gb Shall we add these misspellings as well?
```text
invalidted->invalidated
invalidting->invalidating
validted->invalidated
validting->invalidating
```

cfi-gb:
So some one else would need to say yes/no then and if there is a yes i can apply the changes. :+1:

luzpaz:
Green light from me

cfi-gb:
Thanks a lot for @vikivivi for the suggestion and @luzpaz for the green light / review. The additional words / corrections have been included in a newly pushed version of this MR and are ready for review.

luzpaz (approved):

peternewman (approved):

peternewman:
> So some one else would need to say yes/no then and if there is a yes i can apply the changes. +1

Thanks @cfi-gb and feel free to be bold in future! Especially if it's just a variant of a typo you've found it's likely if you don't add them now, then someone else will in due course...

PR #2307, opened 2022-03-04
Add connectec->connected

peternewman:
I'm a bit torn here, as about 50% of the typos via grep.app are for when it's a SCALA (?) execution context connection variable, but the rest are this intended correction.

We could move to code, but then most of the typos are in code comments...


vikivivi (approved):
LGTM

luzpaz (approved):

luzpaz:
lets try it out and see what happens. We can modify if needed :wink:

PR #2309, opened 2022-03-11

luzpaz:


vikivivi:
@luzpaz suggested word `tansverse` is a misspell. Shall we suggest `transverse`?
```text
tranverse->tansverse, traverse,

tranverse->transverse, traverse,
```

luzpaz:
@vikivivi fixed in 806878b

PR #2313, opened 2022-03-17
Remove anomalous duplicates

luzpaz:
For some reason tests are missing these and `make sort-dictionaries` is malfunctioning as well.


peternewman:
I assume you're just assuming that because it's in the wrong place?

It's not a c, but a https://www.fileformat.info/info/unicode/char/0441/index.htm !

PR #2318, opened 2022-03-25
Add spelling corrections for (un)able.

cfi-gb:
See e.g.:
- https://grep.app/search?q=unablet&words=true
- https://github.com/greenbone/openvas-scanner/pull/1069

Note that `Ablet` seems to be a fish according to https://en.wiktionary.org/wiki/ablet so i have put it into the `dictionary_rare.txt`, hope this is correct like this.


luzpaz (approved):

peternewman (approved):
LGTM thanks.

That's very diligent on ablet, I don't normally bother checking unless the CI complains.

Issue #2273, opened 2022-02-10
Wrong replacement of "complies" (v. comply) with "compiles" (v. compile)


akien-mga:
Seen while testing https://github.com/codespell-project/codespell/commit/e47c98c2e6e93286c062120bacafc79b4970ec7b on the https://github.com/godotengine/godot codebase:

```diff
diff --git a/doc/classes/EditorPaths.xml b/doc/classes/EditorPaths.xml
index c4d4c92afe..afd55d15f3 100644
--- a/doc/classes/EditorPaths.xml
+++ b/doc/classes/EditorPaths.xml
@@ -6,7 +6,7 @@
	<description>
		This editor-only singleton returns OS-specific paths to various data folders and files. It can be used in editor plugins to ensure files are saved in the correct location on each operating system.
		[b]Note:[/b] This singleton is not accessible in exported projects. Attempting to access it in an exported project will result in a script error as the singleton won't be declared. To prevent script errors in exported projects, use [method Engine.has_singleton] to check whether the singleton is available before using it.
-		[b]Note:[/b] Godot complies with the [url=https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html]XDG Base Directory Specification[/url] on [i]all[/i] platforms. You can override environment variables following the specification to change the editor and project data paths.
+		[b]Note:[/b] Godot compiles with the [url=https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html]XDG Base Directory Specification[/url] on [i]all[/i] platforms. You can override environment variables following the specification to change the editor and project data paths.
	</description>
	<tutorials>
		<link title="File paths in Godot projects">https://docs.godotengine.org/en/latest/tutorials/io/data_paths.html</link>
```

While it's true that "compiles" is probably more common than "complies" in codebases, the verb "comply" is still commonly used when talking about specifications. Maybe this hit should be downgraded to a hint of potential typo instead of an autofix one?

luzpaz:
Yes, I wish we had a dictionary that has these types of terms but isn't utilized by default when running codespell.

luzpaz:
@peternewman I've encountered the false positive more than actual hits. But I still think this is valid. But it doesn't belong in code dictionary. Thoughts?

peternewman:
So this is the same issue as #2281 .

I'd prefer not to put them in the main code dictionary, as currently all the "typos" in there are invalid dictionary words, so we can validate that the typo doesn't exist in the dictionary at least.

[...]

PR #2324, opened 2022-03-31
False positive complies->compiles

polluks:
form of "comply"

### DIFF
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -33,7 +33,6 @@ compliancy->compliance
 complied->compiled
 complier->compiler
 compliers->compilers
-complies->compiles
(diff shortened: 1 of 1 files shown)


peternewman:
Hi @polluks ,

Thanks for this, please see the discussion in #2273 , essentially the whole rare dictionary is potential false positives, you can always turn it off if you'd prefer.

polluks:
I see.

sed-i:
Today suddenly started complaining about  `./LICENSE:128: complies ==> compiles`.
How do I disable the rares dictionary? Afaict it's not documented.

peternewman:
> Now need to run with `codespell --builtin clear`.

Just ignoring that one word via an ignore list or ignore file would be the non-nuclear option here.

PR #2331, opened 2022-04-06
Add usesd->used correction.

cfi-gb:
See e.g.:

https://grep.app/search?q=usesd&words=true

Most are valid variables like `useSD` so i have put it into the code dictionary. But some of them are actual typos like e.g.:

> the file whose last modification time is usesd as the

or:

> Usesd for hardware stack checking

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -65,6 +65,6 @@ udate->update, date,
 udates->updates, dates,
 uint->unit
 uis->is, use,
-usesd->used
+usesd->used, uses,
 were'->we're
 whome->whom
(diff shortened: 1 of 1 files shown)


luzpaz:
Can also be `uses` where the `d` is accidental

codespell_lib/data/dictionary_code.txt
```diff
@@ -65,5 +65,6 @@ udate->update, date,
 udates->updates, dates,
 uint->unit
 uis->is, use,
+usesd->used
```
cfi-gb, line 68:
```suggestion
usesd->used, uses,
```

As suggested by @luzpaz in https://github.com/codespell-project/codespell/pull/2331#issuecomment-1092362357

PR #2345, opened 2022-05-08

luzpaz:


codespell_lib/data/dictionary.txt
```diff
@@ -8128,7 +8129,7 @@ correnspondents->correspondents
 corrensponding->corresponding
 corrensponds->corresponds
 corrent->correct, current,
-corrently->correctly
+corrently->correctly, currently,
```
peternewman, line 8132: I think swap these round, your suggestion seems far more likely!

peternewman (changes requested):

codespell_lib/data/dictionary.txt
luzpaz: I initially thought this as well.. but then I looked at line 8130 (above) and decided to copy the order based on that.

peternewman: Yeah actually grep.app seems to agree. Perhaps correct(ly) is just used more commonly.

peternewman (approved):

PR #2346, opened 2022-05-10
Add spelling correction for in-flight

david-marchand:
Signed-off-by: David Marchand <david.marchand@redhat.com>


luzpaz (approved):

peternewman:
Sorry but I think we need to revert this, see for example:
https://www.oed.com/view/Entry/95509

It's also an IBM tool:
https://www.ibm.com/docs/en/scbn?topic=inflight-what-is

For me it's also tripping up some camel case InFlight/inFlight, but I don't know if that's due to my particular regex config.

PR #2360, opened 2022-06-11
typos found in iperf

DimitriPapadopoulos:
See https://github.com/esnet/iperf/pull/1350.

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -64,6 +64,8 @@ shouldnot->should not, shouldn't,
 sinc->sync, sink, since,
 sincs->syncs, sinks, since,
 snd->send, and, sound,
+sockt->socket
+sockts->sockets
 spawnve->spawn
 stdio->studio
 stdios->studios


vikivivi (approved):
LGTM

luzpaz:
i think `sockt` would give **way** too many false positives. Also why is the PR summary have 'relavent' in it?

DimitriPapadopoulos:
`relavant` because it was initially missing,  I guess it disappeared after a rebase.

I agree `sockt` might raise many alerts in existing code, but would they be false positives. I mean, why use `sockt` instead of `socket` in actual code?

skangas:
> I agree `sockt` might raise many alerts in existing code, but would they be false positives? I mean, why use `sockt` instead of `socket` in actual code?

I took a look at [DuckDuckGo](https://html.duckduckgo.com/html/?q=%22sockt%22&norw=1) and [GitHub search](https://github.com/search?q=sockt&type=code), and AFAICT the instances of "sockt" that were found should indeed have been spelled "socket". If we want to be extremely careful, this could of course go to the code dictionary instead. But I'd personally only do that later, if it turns out that users are very unhappy about this change for some reason. So I think this should go in.

Just my two cents.

skangas (approved):

luzpaz:
Alright, lets see what happens.

luzpaz (approved):

luzpaz (changes requested):
2nd thought, i think it should go in the code dict. 
I did a github code search and `sockt` is popular in C code. I'd rather stay on the safe side.

DimitriPapadopoulos:
The question is whether it is a library function, the name of which cannot be changed, or a local variable name which would actually qualify as a typo.

That said, `socket` is a POSIX function name, which probably explains why programmers use alternative spellings.

luzpaz (approved):

PR #2364, opened 2022-06-12
Apply pyupgrade to project

jdufresne:
Update patterns to modern Python syntax and features. Drops unnecessary
legacy syntax that was required for Python 2 (removed in
9c3db1a5c5e02739eedba9e9915acd51736300b3).

Details on pyupgrade can be found at:
https://github.com/asottile/pyupgrade


codespell_lib/tests/test_dictionary.py
```diff
@@ -41,9 +39,9 @@
 
 def test_dictionaries_exist():
     """Test consistency of dictionaries."""
-    doc_fnames = set(op.basename(f[0]) for f in _fnames_in_aspell)
```
ikamensh, line 44: This change kinda makes it less readable - have to look into the expression to see if it's a dict or set.

DimitriPapadopoulos: I would revert these two lines in this merge request. Let's discuss them in #2085.
Thinking about it, `list()`/`set()` are easy to understand because they are English words, while the `[]`/`{}` "_syntax needs to be understood_". While both are acceptable, `[]`/`{}` reminds me of the arcane syntax of Perl. It reminds me that years ago, I would struggle to understand my own commented Perl scripts a few months after having written them, because of the syntax.

codespell_lib/tests/test_basic.py
```diff
@@ -174,7 +172,7 @@ def test_interactivity(tmpdir, capsys):
         with FakeStdin('0\n'):  # blank input -> nothing
             assert cs.main('-w', '-i', '3', f.name) == 0
         assert cs.main(f.name) == 0
-        with open(f.name, 'r') as f_read:
```
DimitriPapadopoulos, line 177: Not sure about this one either. What's wrong with the explicit `r'`? It echoes the 'w' below.

DimitriPapadopoulos:
I would limit this merge request to:
* remove `# -*- coding: utf-8 -*-`
* *remove `(object)`
* change `..._regex_def = u"...` to `..._regex_def = "...`

codespell_lib/tests/test_basic.py
jdufresne: `'r'` is the default, so is unnecessary:

https://docs.python.org/3/library/functions.html#open

It is conventional within the larger Python community to omit it. The goal of this PR is to update the code to follow wider community conventions.

jdufresne:
> I would limit this merge request to ...

I'm going to leave this as-is unless changes are requested by a maintainer. I think using a tool, such as `pyupgrade` to get all changes at once simplifies the PR and history. If any of the PRs you listed lands early, I'll rebase and re-run the tool.

larsoner:
> I'm going to leave this as-is unless changes are requested by a maintainer. I think using a tool, such as pyupgrade to get all changes at once simplifies the PR and history. If any of the PRs you listed lands early, I'll rebase and re-run the tool.

I lean toward this as well, and I think we can revert the `list`/`set` stuff elsewhere if we want. I don't have a strong preference for either syntax personally, so let's decide in #2085

larsoner:
Thanks @jdufresne

PR #2372, opened 2022-06-22
Added memove->memmove to code dict

luzpaz:


codespell_lib/data/dictionary_code.txt
```diff
@@ -30,6 +30,7 @@ ith->with
 jupyter->Jupiter
 keyserver->key server
 lateset->latest
+memove->memmove
```
peternewman, line 33: I don't see memove as a genuine word in code, only memmove, so this can just go in the normal dict.

Again, the code dictionary is only required when the "typo" is a genuine word for coding.

peternewman (changes requested):

luzpaz:
Closing in favor of #2377

PR #2376, opened 2022-06-24
Upgrade GitHub actions across workflows

jdufresne:


luzpaz:
What does this upgrade do?

jdufresne:
Here are the changes: https://github.com/actions/checkout/blob/main/CHANGELOG.md

Most importantly, it keeps the codespell project using the latest compatible version so to allow security fixes should they arise.

The action's README recommends using v3. https://github.com/actions/checkout

luzpaz (approved):

peternewman (approved):
LGTM thanks @jdufresne .

Feel free to do the same for some of our other repos too:
https://github.com/codespell-project

Unless there's a Dependabot or similar to fix them automagically?

PR #2378, opened 2022-06-25
Replace codecs.open with open

jdufresne:
Since Python 3, using codecs.open() is unnecessary as the builtin
supports an encoding argument. Use the simpler and more conventional
pattern.

The default mode for open() is 'r', so redundant arguments have been
removed.

The argument newline='' was added to FileOpener to preserve line endings
when modifying a file.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -17,7 +17,6 @@ Copyright (C) 2011  ProFUSION embedded systems
 """
 
 import argparse
-import codecs
 import configparser
 import fnmatch
 import os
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Thank you for the merge request. It does more than advertised, as it removes the `'r'` mode. I would leave that specific change for a different merge request - such as #2364.

jdufresne:
@DimitriPapadopoulos Rebased :slightly_smiling_face:

DimitriPapadopoulos:
In short, this looks good. I would just explain the`'rb'`/`'wb'` → `'r'`/`'w'` + `newline=''` transition in the commit message.

jdufresne:
> In short, this looks good. I would just explain the`'rb'`/`'wb'` → `'r'`/`'w'` + `newline=''` transition in the commit message.

Done. Let me know if you would like any additional tweaks to the wording.

DimitriPapadopoulos (approved):

PR #2381, opened 2022-06-26
Move tests out of package and into top-level directory

jdufresne:
This follows the "Tests outside application code" pattern documented in
the pytest guides. See:
https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code

> This has the following benefits:
>
> - Your tests can run against an installed version after executing pip
> install ..
>
> - Your tests can run against the local copy with an editable install
> after executing pip install --editable ..
>
> - If you don’t use an editable install and are relying on the fact
> that Python by default puts the current directory in sys.path to
> import your package, you can execute python -m pytest to execute the
> tests against the local copy directly, without using pip.

This keeps tests separate from actual library code. End users don't need
the tests mixed in with the installed packages. Instead, tests are only
interesting to library developers and packagers. The tests remain
available, just moved.


DimitriPapadopoulos:
Moving to `pyproject.toml` and changes in tests require resolving conflicts, probably complex ones.

Also, it looks like this change did not convince: it's one set of conventions against another. Why change and not stick with the initial conventions?

PR #2382, opened 2022-06-26
Resolve CoverageWarning when running tests

jdufresne:
Appeared as:

    .../site-packages/coverage/inorout.py:472: CoverageWarning: --include is ignored because --source is set (include-ignored)
      self.warn("--include is ignored because --source is set", slug="include-ignored")


larsoner:
Thanks @jdufresne

PR #2397, opened 2022-07-09
Try and improve the code dictionary explanation

peternewman:


luzpaz (approved):

peternewman:
Do you think that removes some of your confusion @luzpaz or does it need extending still?

PR #2400, opened 2022-07-11
Implement inline ignores

### LINKED ISSUE #1212: Support inline ignores
I'd like to use codespell as part of an automated check workflow.  For that I need to be able to ignore certain cases on a per instance bases.

Here's an example.  In the code base there is a command called 'Trim All'.  The comment contains a statement like:
// Track how many Trim Alls have been issued.

codespell finds Alls as a possible error.  

I don't want to add 'Alls' to the ignore words list as it could be a real error in other cases.  But I don't want this message generated on every run either.
The ignore word list only supports single words So I can't do 'Trim Alls' as an ignore.

Some sort of  in line keystring that allows the next occurrence of a specific word error would be useful.  Most code formatters, pretty-printers, and static analysis tools have some sort of inline // disable-whatever  option that allows you to skip the trigger on a per instance basis.

kaste:
Fixes #1212

Following @rob-miller's suggestion implement inline ignore hints.

E.g.

```
crate  // codespell:ignore crate
abandonned abondon abilty  # codespell:ignore abondon,abilty
abandonned abondon abilty  # codespell:ignore  # to ignore all
```

### DIFF
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -54,7 +54,7 @@ jobs:
       # tomli should not be required for the next two steps (and make sure it's not)
       - run: pip uninstall -yq tomli
         if: ${{ matrix.no-toml == 'no-toml' }}
-      - run: codespell --check-filenames --skip="./.git/*,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
+      - run: codespell --check-filenames --skip="./.git/*,./.pytest_cache/*,./junit-results.xml,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
       # this file has an error
       - run: "! codespell codespell_lib/tests/test_basic.py"
 
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -54,7 +54,7 @@ jobs:
       # tomli should not be required for the next two steps (and make sure it's not)
       - run: pip uninstall -yq tomli
         if: ${{ matrix.no-toml == 'no-toml' }}
-      - run: codespell --check-filenames --skip="./.git/*,./.pytest_cache/*,./junit-results.xml,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
+      - run: codespell --check-filenames --skip="./.git/*,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
       # this file has an error
       - run: "! codespell codespell_lib/tests/test_basic.py"
 
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -54,7 +54,7 @@ jobs:
       # tomli should not be required for the next two steps (and make sure it's not)
       - run: pip uninstall -yq tomli
         if: ${{ matrix.no-toml == 'no-toml' }}
-      - run: codespell --check-filenames --skip="./.git/*,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
+      - run: codespell --check-filenames --skip="./.git/*,./.pytest_cache/*,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
       # this file has an error
       - run: "! codespell codespell_lib/tests/test_basic.py"
 
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -54,7 +54,7 @@ jobs:
       # tomli should not be required for the next two steps (and make sure it's not)
       - run: pip uninstall -yq tomli
         if: ${{ matrix.no-toml == 'no-toml' }}
-      - run: codespell --check-filenames --skip="./.git/*,./.pytest_cache/*,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
+      - run: codespell --check-filenames --skip="./.git/*,./.pytest_cache/*,./junit-results.xml,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
       # this file has an error
       - run: "! codespell codespell_lib/tests/test_basic.py"
 
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -56,7 +56,7 @@ uri_regex_def = (
 # alternative misspellings and fixes.
 alt_chars = (("'", "’"),)  # noqa: RUF001
 inline_ignore_regex = re.compile(
-    r"([^\w\s]) codespell:ignore(\s(?P<words>[\w,]*))?(\s+\1|$)"
+    r"[^\w\s]\s?codespell:ignore\b(\s+(?P<words>[\w,]*))?"
 )
 USAGE = """
 \t%prog [OPTIONS] [file1 file2 ... fileN]
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -55,6 +55,8 @@ uri_regex_def = (
 # Pass all misspellings through this translation table to generate
 # alternative misspellings and fixes.
 alt_chars = (("'", "’"),)  # noqa: RUF001
+inline_ignore_regex = re.compile(
+    r"[^\w\s] codespell:ignore(?:(?:\s|$)([\w,]*))")
 USAGE = """
 \t%prog [OPTIONS] [file1 file2 ... fileN]
 """
@@ -974,6 +976,15 @@ def parse_file(
         if line.rstrip() in exclude_lines:
             continue
 
+        extra_words_to_ignore = set()
+        match = inline_ignore_regex.search(line)
+        if match:
+            extra_words_to_ignore = set(
+                filter(None, (match.group(1) or "").split(","))
+            )
+            if not extra_words_to_ignore:
+                continue
+
         fixed_words = set()
         asked_for = set()
 
@@ -998,7 +1009,7 @@ def parse_file(
             if word in ignore_words_cased:
                 continue
             lword = word.lower()
-            if lword in misspellings:
+            if lword in misspellings and lword not in extra_words_to_ignore:
                 # Sometimes we find a 'misspelling' which is actually a valid word
                 # preceded by a string escape sequence.  Ignore such cases as
                 # they're usually false alarms; see issue #17 among others.
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -55,9 +55,7 @@ uri_regex_def = (
 # Pass all misspellings through this translation table to generate
 # alternative misspellings and fixes.
 alt_chars = (("'", "’"),)  # noqa: RUF001
-inline_ignore_regex = re.compile(
-    r"[^\w\s]\s?codespell:ignore\b(\s+(?P<words>[\w,]*))?"
-)
+inline_ignore_regex = re.compile(r"[^\w\s]\s?codespell:ignore\b(\s+(?P<words>[\w,]*))?")
 USAGE = """
 \t%prog [OPTIONS] [file1 file2 ... fileN]
 """
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -56,7 +56,8 @@ uri_regex_def = (
 # alternative misspellings and fixes.
 alt_chars = (("'", "’"),)  # noqa: RUF001
 inline_ignore_regex = re.compile(
-    r"[^\w\s] codespell:ignore(?:(?:\s|$)([\w,]*))")
+    r"([^\w\s]) codespell:ignore(\s(?P<words>[\w,]*))?(\s+\1|$)"
+)
 USAGE = """
 \t%prog [OPTIONS] [file1 file2 ... fileN]
 """
@@ -980,7 +981,7 @@ def parse_file(
         match = inline_ignore_regex.search(line)
         if match:
             extra_words_to_ignore = set(
-                filter(None, (match.group(1) or "").split(","))
+                filter(None, (match.group("words") or "").split(","))
             )
             if not extra_words_to_ignore:
                 continue
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -391,43 +391,58 @@ def test_ignore_word_list(
     assert cs.main("-Labandonned,someword", "-Labilty", tmp_path) == 1
 
 
-@pytest.mark.parametrize('content, expected_error_count', [
-    # recommended form
-    ('abandonned abondon abilty  # codespell:ignore abondon', 2),
-    ('abandonned abondon abilty  // codespell:ignore abondon,abilty', 1),
-    ('abandonned abondon abilty  /* codespell:ignore abandonned,abondon,abilty', 0),  # noqa: E501
-    # ignore unused ignore
-    ('abandonned abondon abilty  # codespell:ignore nomenklatur', 3),
-    # wildcard form
-    ('abandonned abondon abilty  # codespell:ignore ', 0),
-    ('abandonned abondon abilty  # codespell:ignore', 0),
-    ('abandonned abondon abilty  # codespell:ignore\n', 0),
-    ('abandonned abondon abilty  # codespell:ignore\r\n', 0),
-    ('abandonned abondon abilty  # codespell:ignore  # noqa: E501\n', 0),
-    ('abandonned abondon abilty  # codespell:ignore # noqa: E501\n', 0),
-    ('abandonned abondon abilty  # codespell:ignore# noqa: E501\n', 0),
-    ('abandonned abondon abilty  # codespell:ignore, noqa: E501\n', 0),
-    ('abandonned abondon abilty  #codespell:ignore\n', 0),
-    # ignore these for safety
-    ('abandonned abondon abilty  # codespell:ignorenoqa: E501\n', 3),
-    ('abandonned abondon abilty  codespell:ignore\n', 3),
-    ('abandonned abondon abilty codespell:ignore\n', 3),
-    # showcase different comment markers
-    ("abandonned abondon abilty ' codespell:ignore\n", 0),
-    ('abandonned abondon abilty " codespell:ignore\n', 0),
-    ('abandonned abondon abilty ;; codespell:ignore\n', 0),
-    ('abandonned abondon abilty /* codespell:ignore */\n', 0),
-    # prose examples
-    ('You could also use line based igore ( codespell:ignore ) to igore ', 0),
-    ('You could also use line based igore (codespell:ignore) to igore ', 0),
-    ('You could also use line based igore (codespell:ignore igore) to igore ', 0),  # noqa: E501
-    ('You could also use line based igore (codespell:ignore igare) to igore ', 2),  # noqa: E501
-])
+@pytest.mark.parametrize(
+    'content, expected_error_count',
+    [
+        # recommended form
+        ('abandonned abondon abilty  # codespell:ignore abondon', 2),
+        ('abandonned abondon abilty  // codespell:ignore abondon,abilty', 1),
+        (
+            'abandonned abondon abilty  /* codespell:ignore abandonned,abondon,abilty',
+            0,
+        ),  # noqa: E501
+        # ignore unused ignore
+        ('abandonned abondon abilty  # codespell:ignore nomenklatur', 3),
+        # wildcard form
+        ('abandonned abondon abilty  # codespell:ignore ', 0),
+        ('abandonned abondon abilty  # codespell:ignore', 0),
+        ('abandonned abondon abilty  # codespell:ignore\n', 0),
+        ('abandonned abondon abilty  # codespell:ignore\r\n', 0),
+        ('abandonned abondon abilty  # codespell:ignore  # noqa: E501\n', 0),
+        ('abandonned abondon abilty  # codespell:ignore # noqa: E501\n', 0),
+        ('abandonned abondon abilty  # codespell:ignore# noqa: E501\n', 0),
+        ('abandonned abondon abilty  # codespell:ignore, noqa: E501\n', 0),
+        ('abandonned abondon abilty  #codespell:ignore\n', 0),
+        # ignore these for safety
+        ('abandonned abondon abilty  # codespell:ignorenoqa: E501\n', 3),
+        ('abandonned abondon abilty  codespell:ignore\n', 3),
+        ('abandonned abondon abilty codespell:ignore\n', 3),
+        # showcase different comment markers
+        ("abandonned abondon abilty ' codespell:ignore\n", 0),
+        ('abandonned abondon abilty " codespell:ignore\n', 0),
+        ('abandonned abondon abilty ;; codespell:ignore\n', 0),
+        ('abandonned abondon abilty /* codespell:ignore */\n', 0),
+        # prose examples
+        (
+            'You could also use line based igore ( codespell:ignore ) to igore ',
+            0,
+        ),
+        ('You could also use line based igore (codespell:ignore) to igore ', 0),
+        (
+            'You could also use line based igore (codespell:ignore igore) to igore ',
+            0,
+        ),  # noqa: E501
+        (
+            'You could also use line based igore (codespell:ignore igare) to igore ',
+            2,
+        ),  # noqa: E501
+    ],
+)
 def test_inline_ignores(
     tmpdir: pytest.TempPathFactory,
     capsys: pytest.CaptureFixture[str],
     content: str,
-    expected_error_count: int
+    expected_error_count: int,
 ) -> None:
     d = str(tmpdir)
     with open(op.join(d, 'bad.txt'), 'w') as f:
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -396,6 +396,8 @@ def test_ignore_word_list(
     ('abandonned abondon abilty  # codespell:ignore abondon', 2),
     ('abandonned abondon abilty  // codespell:ignore abondon,abilty', 1),
     ('abandonned abondon abilty  /* codespell:ignore abandonned,abondon,abilty', 0),  # noqa: E501
+    # ignore unused ignore
+    ('abandonned abondon abilty  # codespell:ignore nomenklatur', 3),
     # wildcard form
     ('abandonned abondon abilty  # codespell:ignore ', 0),
     ('abandonned abondon abilty  # codespell:ignore', 0),
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -391,6 +391,42 @@ def test_ignore_word_list(
     assert cs.main("-Labandonned,someword", "-Labilty", tmp_path) == 1
 
 
+@pytest.mark.parametrize('content, expected_error_count', [
+    # recommended form
+    ('abandonned abondon abilty  # codespell:ignore abondon', 2),
+    ('abandonned abondon abilty  // codespell:ignore abondon,abilty', 1),
+    ('abandonned abondon abilty  /* codespell:ignore abandonned,abondon,abilty', 0),  # noqa: E501
+    # wildcard form
+    ('abandonned abondon abilty  # codespell:ignore ', 0),
+    ('abandonned abondon abilty  # codespell:ignore', 0),
+    ('abandonned abondon abilty  # codespell:ignore\n', 0),
+    ('abandonned abondon abilty  # codespell:ignore  # noqa: E501\n', 0),
+    ('abandonned abondon abilty  # codespell:ignore # noqa: E501\n', 0),
+    # ignore these for safety
+    ('abandonned abondon abilty  # codespell:ignore# noqa: E501\n', 3),
+    ('abandonned abondon abilty  # codespell:ignore, noqa: E501\n', 3),
+    ('abandonned abondon abilty  # codespell:ignore/noqa: E501\n', 3),
+    ('abandonned abondon abilty  # codespell:ignorenoqa: E501\n', 3),
+    ('abandonned abondon abilty  #codespell:ignore\n', 3),
+    ('abandonned abondon abilty  codespell:ignore\n', 3),
+    ('abandonned abondon abilty codespell:ignore\n', 3),
+    # showcase different comment markers
+    ("abandonned abondon abilty ' codespell:ignore\n", 0),
+    ('abandonned abondon abilty " codespell:ignore\n', 0),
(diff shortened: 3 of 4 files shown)


codespell_lib/_codespell.py
```diff
@@ -642,6 +644,15 @@ def parse_file(filename, colors, summary, misspellings, exclude_lines,
         if line in exclude_lines:
             continue
 
+        extra_words_to_ignore = set()
```
peternewman, line 647: Maybe a more useful variable name like local_words_to_ignore or ignore_words_line or something.

codespell_lib/tests/test_basic.py
```diff
@@ -233,6 +233,37 @@ def test_ignore_word_list(tmpdir, capsys):
...
 
+@pytest.mark.parametrize('content, expected_error_count', [
+    # recommended form
+    ('abandonned abondon abilty  # codespell:ignore abondon', 2),
+    ('abandonned abondon abilty  // codespell:ignore abondon,abilty', 1),
+    ('abandonned abondon abilty  /* codespell:ignore abandonned,abondon,abilty', 0),  # noqa: E501
+    # wildcard form
+    ('abandonned abondon abilty  # codespell:ignore ', 0),
+    ('abandonned abondon abilty  # codespell:ignore', 0),
+    ('abandonned abondon abilty  # codespell:ignore\n', 0),
```
peternewman, line 244:
```suggestion
    ('abandonned abondon abilty  # codespell:ignore\n', 0),
    ('abandonned abondon abilty  # codespell:ignore\r\n', 0),
```

codespell_lib/_codespell.py
```diff
@@ -32,6 +32,8 @@
 # endpoint.  Emails are more restrictive, so the endpoint remains flexible.
 uri_regex_def = (u"(\\b(?:https?|[ts]?ftp|file|git|smb)://[^\\s]+(?=$|\\s)|"
                  u"\\b[\\w.%+-]+@[\\w.-]+\\b)")
+inline_ignore_regex = re.compile(
+    r"[^\w\s] codespell:ignore(?:(?:\s|$)([\w,]*))")
```
peternewman, line 36: Won't this also match, and hence ignore the typos in this:
`You could also use the line based ignor ( codespell:ignore ) to ignor the whole line.`

Which I'd call a false positive!

codespell_lib/tests/test_basic.py
```diff
@@ -233,6 +233,37 @@ def test_ignore_word_list(tmpdir, capsys):
     assert cs.main('-Labandonned,someword', '-Labilty', d) == 1
 
 
+@pytest.mark.parametrize('content, expected_error_count', [
```
peternewman, line 236: You haven't tested where the typo doesn't exist in the line at all, but is still being erroneously ignored!

peternewman (dismissed):
Some of the tests are failing here due to files in ./.pytest_cache so you'll need to add that in our private GitHub workflow.

A few comments too.

Personally I'm still not entirely convinced given the same issue as other ignores, there really needs to be a way to escape it.

codespell_lib/_codespell.py
kaste: Questionable what the example sentence you give here should mean.  False positives are always a problem but then it looks like you constructed the example to produce the false positive.  In other words your sentence is probably not something a user would type.   (In fact, you probably can only write down ` codespell:ignore ` in technical documentation *about* codespell itself.)  

For prose, "inline-disabling" is probably not what you want anyway because there are no comments in prose.  A user would read the technical ignore statement in any case.  (There is also no syntax highlighting in the editor which would gray out our ignore statements.)    

Usually, such ignore statements are the last thing on a line except you have other *comments* on the same line,  so maybe we can change the behavior and only let `... | codespell:ignore | ...` match t.i. on the left an right we find the same character, thus the ` ( ... ) ` you gave would not match.   I'll push an update here.

Actually a simple `... (codespell:ignore) ...` would be very nice syntax for prose and I wonder if a simpler `r"[^\w\s]\s?codespell:ignore\b(\s+(?P<words>[\w,]*))?"` would *just* work or "be enough" in practice.

codespell_lib/tests/test_basic.py
kaste: Done.

codespell_lib/_codespell.py
kaste: I don't like your suggestions.  Postpone.  What it holds is "additional (extra) words to ignore on this line".

kaste:
I just pushed a more relaxed version as - after a week -  I found this more and more nice for prose.  I push this so then I don't have a dirty workdir on my computer, and we can easily see which tests pass now, and with git we don't loose the previous version anyway.  

Since the verbatim `codespell:ignore` is already seldom to find in the wild a lax version can be just good enough.  

I get the feeling that you don't want to decide on this one; and honestly I just took the most liked issue to give it a try and I did not loose a deep thought as it's not my idea I implemented but ... I think I like it and it is a good simple implementation idea.  It is naive but likable.

codespell_lib/_codespell.py
```diff
@@ -32,6 +32,8 @@
 # endpoint.  Emails are more restrictive, so the endpoint remains flexible.
 uri_regex_def = (u"(\\b(?:https?|[ts]?ftp|file|git|smb)://[^\\s]+(?=$|\\s)|"
                  u"\\b[\\w.%+-]+@[\\w.-]+\\b)")
+inline_ignore_regex = re.compile(
+    r"[^\w\s]\s?codespell:ignore\b(\s+(?P<words>[\w,]*))?")
```
akx, line 36: To make this compatible with https://github.com/codespell-project/codespell/issues/1212#issuecomment-1181813266 (i.e. https://cspell.org/configuration/document-settings/), how about

```suggestion
    r"[^\w\s]\s?(?:codespell|cspell|spell-checker):ignore\b(\s+(?P<words>[\w,]*))?", flags=re.I)
```

codespell_lib/_codespell.py
```diff
@@ -642,6 +644,15 @@ def parse_file(filename, colors, summary, misspellings, exclude_lines,
         if line in exclude_lines:
             continue
 
+        extra_words_to_ignore = set()
+        match = inline_ignore_regex.search(line)
+        if match:
+            extra_words_to_ignore = set(
+                filter(None, (match.group("words") or "").split(","))
+            )
```
akx, line 652: Could use a set comprehension:
```suggestion
            extra_words_to_ignore = {w for w in (match.group("words") or "").split(",") if w}
```

akx (changes requested):

yarikoptic:
is there a way to add more than one :heart: to the description of this PR? ;-)

victor-kostyuk:
What's holding up this PR? Without inline ignores, codespell is much less useful.

kaste:
The problem is having false positives with this approach as I understand the maintainer(s).  I don't know if we could avoid them.  I do think they are not very likely and then not very harmful.

DimitriPapadopoulos (approved):

victor-kostyuk:
@kaste : could you accept @akx 's suggested changes? 

@peternewman : any issues you consider serious enough to block this PR?

DimitriPapadopoulos:
Need to rebase to resolve conflicts too.

kaste:
I do not implement cspells ignore functionality, https://cspell.org/configuration/document-settings/#ignore  I also replied over there https://github.com/codespell-project/codespell/issues/1212#issuecomment-1193002440 that I do *not*.  This is only ignore-word-on-this-line.  cspell doesn't have this functionality.  They have ignore-for-this-file.  

Having multiple prefixes, spellings does not mitigate false positives either, on contraire.

victor-kostyuk:
> I do not implement cspells ignore functionality, [cspell.org/configuration/document-settings/#ignore](https://cspell.org/configuration/document-settings/#ignore) I also replied over there [#1212 (comment)](https://github.com/codespell-project/codespell/issues/1212#issuecomment-1193002440) that I do _not_. This is only ignore-word-on-this-line. cspell doesn't have this functionality. They have ignore-for-this-file.
> 
> Having multiple prefixes, spellings does not mitigate false positives either, on contraire.

cspell has [ignore line](https://cspell.org/configuration/document-settings/#disable-checking): `// cspell:disable-line`. Adding it is unlikely to produce false positives, for the same reason that this is unlikely to produce false positives. Also, @akx second comment was not about cspell.

kaste:
He?  I don't implement the `:disable-line` directive but the `:ignore` one.  `cspell:ignore` has a different meaning than the `codespell:ignore` I tried here.

neiljp:
I was just about to open an issue until finally finding #1212, and then this PR!
Thanks for looking at this @kaste :+1:

This is a great exploratory PR; perhaps slimming it down could help find consensus for merging?
- Would looking up an appropriate comment character based on the file type help with some of the false-positive concerns, since the regex could then be more specific? Initially one could only enable this option if users provide a mapping from extensions to characters
- I suspect many would be initially satisfied with such comments on end of line only
- While excluding specific words on a line would be great, simply marking a line to be ignored is a great feature on its own; the word-specific feature could be added later

I think there's plenty of scope for extending this later:
- the words on a line feature currently here
- customizing the tool prefix used, eg. to allow `codespell: `, and/or `cspell:`
- supporting auto file type detection (optionally, to avoid a required magic dependency?)
- the line-below features in cspell? (comment on different line to 'code')
- perhaps a similar 'block below' (paragraph, function, curly-brace block?)
- turning the checker on/off on lines in a file, like isort offers
- excluding a file this way, as a source-equivalent version of the existing option
- excluding words in a file this way
- possibly support aliases for documents marked up for other tools? (not the tool prefix, but also actions)

codespell_lib/_codespell.py
DimitriPapadopoulos: `cspell` looks like a bad idea for the reasons explained in https://github.com/codespell-project/codespell/issues/1212#issuecomment-1193002440. I would also defer `spell-checker` to a later time when in document settings consistent with _cspell_ are implemented.

codespell_lib/_codespell.py
DimitriPapadopoulos: Perhaps `words_to_ignore_in_line`?

DimitriPapadopoulos (commented):
I could approve when it passes tests.

Perhaps change `extra_words_to_ignore` to `words_to_ignore_in_line`, but that's a suggestion, and you don't have to apply it if you don't agree.

DimitriPapadopoulos:
> * Would looking up an appropriate comment character based on the file type help with some of the false-positive concerns, since the regex could then be more specific? Initially one could only enable this option if users provide a mapping from extensions to characters

We currently don't have a “file type”, adding one would probably have the inverse effect of slimming down.

> * I suspect many would be initially satisfied with such comments on end of line only

Yes, probably.

> * While excluding specific words on a line would be great, simply marking a line to be ignored is a great feature on its own; the word-specific feature could be added later

Yes, in the case ignore might be changed to `disable-line` for consistency with _cspell_.

kaste:
At least I'm green 😄 

image

finally.

benner:
is this stalled?

12rambau:
@kaste and @DimitriPapadopoulos, I just wanted to mention that this PR is perfectly solving the problem I raised in the real life example I shared https://github.com/codespell-project/codespell/issues/1212#issuecomment-1654457104. 
It seems I don't really understand the connection between this project and `cspell` and why there was so much fuss on aligning with it but let's face it, @kaste proposal here is a one-life chance to solve an issue that was open in 2019. Iterating on it for extra stuff (like the multi-line ignore mentioned in my real-life example but also by @neiljp) would be great but this simple and elegant statement will already solve hundreds of problems.

Now that all is green in the tests, any chance to see it moving forward in the near future and published in a release ?
I could continue relying on a non-published branch but it prevent me from releasing on conda-forge without tricky workarounds which is ... sad. 

Anything we can do from a non-maintainer side (apart from saying "it's great" and sending emojis 😄) ?

DimitriPapadopoulos:
Could you perhaps resolve conflicts one last time? Sorry about that.

kaste:
Yeah done. And yes, I don't refactor that `parse_file` function which was 200 lines long before this PR and now has two lines more and the linter says: wow, that's probably too long.   No, lgtm!

DimitriPapadopoulos:
If tests don't pass, don't refactor, just bump the complexity metric as needed. You're right: at some point we might need to refactor, but not here.

kaste:
I don't say that you should refactor `parse_file`.  You installed that metric.  `parse_file` is just line by line what needs to be done.  There is no modularity or polymorphism here, it is more like a script.

codespell_lib/_codespell.py
```diff
@@ -974,6 +975,15 @@ def parse_file(
...
+            if not extra_words_to_ignore:
+                continue
```
peternewman, line 985: Isn't this a no-op currently, and a potential bear trap if some additional line-based processing is added later after this statement, as it then won't be actioned if there aren't extra_words_to_ignore?

codespell_lib/tests/test_basic.py
```diff
@@ -391,6 +391,65 @@ def test_ignore_word_list(
...
+        ("abandonned abondon abilty  # codespell:ignore  # noqa: E501\n", 0),
+        ("abandonned abondon abilty  # codespell:ignore # noqa: E501\n", 0),
+        ("abandonned abondon abilty  # codespell:ignore# noqa: E501\n", 0),
+        ("abandonned abondon abilty  # codespell:ignore, noqa: E501\n", 0),
+        ("abandonned abondon abilty  #codespell:ignore\n", 0),
+        # ignore these for safety
+        ("abandonned abondon abilty  # codespell:ignorenoqa: E501\n", 3),
+        ("abandonned abondon abilty  codespell:ignore\n", 3),
+        ("abandonned abondon abilty codespell:ignore\n", 3),
+        # showcase different comment markers
```
peternewman, line 420: Just testing another edge case explicitly
```suggestion
        # ignore these as they aren't valid
        ("abandonned abondon abilty  # codespell:igore\n", 3),
        # showcase different comment markers
```

peternewman (dismissed):
Okay I'm entirely won over on the general operation!

I agree with @DimitriPapadopoulos though, I think we need to change the trigger word given it differs to what cspell uses (and they got there first). Plus they also have both a similar name to us and a generic spell-checker option too:
https://cspell.org/configuration/document-settings/

I already use codespell and lintian on some CI, so I think being able to run multiple tools with the same annotations (or at least not conflicting ones) is definitely worth a slightly longer trigger word.

I think `ignore-line` would build on what they have already without conflicting with it (and allow a future `ignore-next-line`).

I think it would also be beneficial to have a way to turn it off entirely (so just a simple command line argument which defaults to true) probably to disable all future inline annotation. Just for our docs if nothing else. Unless our regex options can work to filter the magic word out, I've not tested, but just disabling may be easier.

As @DimitriPapadopoulos it needs some documentation too.

A few other minor comments too.

codespell_lib/_codespell.py
kaste: That's for sure not a no-op.  It shouldn't.  Try deleting the line and tests should fail.  If at that point `extra_words_to_ignore` is empty, we have the semantics to ignore the line.  That is the wildcard ignore.  

Otherwise `extra_words_to_ignore` holds additional (extra) words which are only used for this line.  

Just the lines above we have the same or similar construct: `if line in exclude_lines: continue`. So there is also an early return.

kaste:
> I think ignore-line would build on what they have already without conflicting with it (and allow a future ignore-next-line).

Is that that you propose: e.g. `codespell:ignore-line frends` in favor of `codespell:ignore frinds`?  Because they have `ignore` but not `ignore-line`.  But in this form we don't use their namespace?  And when we have `ignore` we can still have `ignore-next-line`, there is no contradiction, in style or semantics.  

We could of course support `cspell:disable-line` and its "next" variant.  But that's not what I did here, and you sure can have both.

codespell_lib/tests/test_basic.py
```diff
@@ -391,6 +391,67 @@ def test_ignore_word_list(
...
+        ("abandonned abondon abilty  # codespell:ignore # noqa: E501\n", 0),
+        ("abandonned abondon abilty  # codespell:ignore# noqa: E501\n", 0),
+        ("abandonned abondon abilty  # codespell:ignore, noqa: E501\n", 0),
+        ("abandonned abondon abilty  #codespell:ignore\n", 0),
+        # ignore these for safety
+        ("abandonned abondon abilty  # codespell:ignorenoqa: E501\n", 3),
+        ("abandonned abondon abilty  codespell:ignore\n", 3),
+        ("abandonned abondon abilty codespell:ignore\n", 3),
+        # ignore these as they aren't valid
+        ("abandonned abondon abilty  # codespell:igore\n", 3),
```
kaste, line 421:
```suggestion
        ("abandonned abondon abilty  # codespell:igore\n", 4),
```

.github/workflows/codespell-private.yml
```diff
@@ -54,7 +54,7 @@ jobs:
       # tomli should not be required for the next two steps (and make sure it's not)
       - run: pip uninstall -yq tomli
         if: ${{ matrix.no-toml == 'no-toml' }}
-      - run: codespell --check-filenames --skip="./.git/*,*.pyc,./codespell_lib/tests/test_basic.py,./codespell_lib/data/*,./example/code.c,./build/lib/codespell_lib/tests/test_basic.py,./build/lib/codespell_lib/data/*,README.rst,*.egg-info/*,pyproject-codespell.precommit-toml,./.mypy_cache"
```
DimitriPapadopoulos, line 57: Please remove this change. It would be OK in a dedicated merge request, but is unrelated to inline ignores.

pyproject.toml
```diff
@@ -171,4 +171,4 @@ allow-magic-value-types = ["bytes", "int", "str",]
 max-args = 13
 max-branches = 49
 max-returns = 11
-max-statements = 113
+max-statements = 200
```
DimitriPapadopoulos, line 174: I would rather bump to the exact value needed for this change. It might help identify refactoring opportunities in future pull requests - but of course not here in a tight loop and without performance testing in CI.

DimitriPapadopoulos (changes requested):

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
> I think `ignore-line` would build on what they have already without conflicting with it (and allow a future `ignore-next-line`).

Why `ignore-line`? Linter directives I know of apply to the current line. For example, in Python, `noqa` applies to the current line. Perhaps there's a misunderstanding here. The proposed `ignore` option suggests the words after the directive `ignore:` are ignored in the current line, just like the errors after the `noqa:` directive are suppressed.

.github/workflows/codespell-private.yml
DimitriPapadopoulos: Ouch! My wrong! Removing `./junit-results.xml` results in a truckload of errors. Not sure why, as this seems orthogonal to your change.

Any way, I am sorry about it, and I suggest you put `./junit-results.xml` back in the skip list.

DimitriPapadopoulos:
> As @DimitriPapadopoulos it needs some documentation too.

Indeed, would it be possible to add a quick explanation under [Ignoring words](https://github.com/codespell-project/codespell?tab=readme-ov-file#ignoring-words).

DimitriPapadopoulos:
> I think it would also be beneficial to have a way to turn it off entirely (so just a simple command line argument which defaults to true) probably to disable all future inline annotation. Just for our docs if nothing else. Unless our regex options can work to filter the magic word out, I've not tested, but just disabling may be easier.

Not certain it's required at this point, this pull request has been waiting in queue for much too long. Nevertheless, the [Flake8 documentation](https://flake8.pycqa.org/en/latest/user/violations.html#in-line-ignoring-errors) states:
> If we ever want to disable **Flake8** respecting `# noqa` comments, we can refer to [**`flake8 --disable-noqa`**](https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-disable-noqa).

Could you perhaps add a `--disable-ignore` option? You'd have to bump the complexity metrics accordingly.

.github/workflows/codespell-private.yml
kaste: I explain in the commit ([b8183ee](https://github.com/codespell-project/codespell/pull/2400/commits/b8183ee98fcc55e74a8cd5b3b1fa6b4cba628fa1)) why I need to skip this file.  🤔

kaste:
>	 Could you perhaps add a --disable-ignore option? You'd have to bump the complexity metrics accordingly.

Who? Me? No.  But good joke on bumping the complexity metrics.  That's why we enable such rules in CI.  👍

peternewman:
> > I think `ignore-line` would build on what they have already without conflicting with it (and allow a future `ignore-next-line`).
> 
> Why `ignore-line`? Linter directives I know of apply to the current line. For example, in Python, `noqa` applies to the current line. Perhaps there's a misunderstanding here. The proposed `ignore` option suggests the words after the directive `ignore:` are ignored in the current line, just like the errors after the `noqa:` directive are suppressed in the current line.

CSpell define a chunk of directives which apply to the whole file:
https://cspell.org/configuration/document-settings/#in-document-settings

And then some that apply to a line (either the current or next one):
https://cspell.org/configuration/document-settings/#enable--disable-checking-sections-of-code

Note how it explicitly says ( https://cspell.org/configuration/document-settings/#ignore ):
> Note: words defined with ignore will be ignored for the entire file.

So I think it's counter-productive and unnecessarily confusing to define our ignore to apply to a line when there's applies to a whole file and they effectively define a "syntax" of `-line` meaning something affects just one line.

I'm fairly confident people have previously asked for the option to do other config with Codespell within the file (e.g. skipping it) so I'd imagine such clamouring will only increase once this is in.

codespell_lib/_codespell.py
peternewman: Apologies @kaste I was missing the `if match` statement a few lines up, failing to read it over the PR comment above this one.

kaste:
TL;DR  The only better pragma is the universal understood and available `sic!`.


Hm,  when bike-shedding you can't just take the one argument.  Yours is also weak
as it constantly compares to a different tool/community.  Now `cspell` comes
from the javascript ecosystem, and here they just follow the ill historical
development of `eslint` where it is `eslint-disable-line` and all the other 
complicated stuff.  Why is their decision wrong?  Well, you seldom disable
whole files *inline*, more often you then have "skip-globs" outside of these files.  
(Most often you're not allowed to write to these files so that's why they don't adhere
to the rules.)  But you regularly need to some exceptions on a case-per-case 
basis, for that you have inline-pragmas.  They belong to the place where the 
rule is violated.  Here `codespell` was always wrong with the global exclusion-
ignore-files.  

Ironically you double this error where `ruff` tells you that a function is too
complicated and as an answer bump the global value, although `ruff` has inline
ignores (of course!) and you actually just need to ignore that specific function.

What are actually the *strong* arguments for "codespell:ignore"?

- It was the natural idea of someone, back then in the issue, that did something
  in me so I did the implementation.

- It is shorter.

- It reads like the imperative form: "Codespell, ignore frind!".  That comes just 
  natural.  

All that is important esp. in prose, e.g. markdown files!  We're really looking
more for a "sic! frind" pragma than a "codepspell:ignore-next-line frind".  
Everyone can spot the difference.

- The `-` in "ignore-line" is not language agnostic.  In some texts "ignoreLine"
  or "IgnoreLine" would be look more pleasant.

- The idea is out for 2 yars, and people use the hack from the original
  issue already.  (codespell:ignore yars)

One comment:  It is okay to be a bike-shedder and micro-manager, but one should
know onself.  (sic!)

DimitriPapadopoulos:
> Who? Me? No. But good joke on bumping the complexity metrics. That's why we enable such rules in CI. 👍

I may add the option later on, and I will certainly add some documentation.

About disabling the complexity warning, you're right, but I would like to keep track of where we stand. I would like to reduce the complexity at some point, but am afraid of refactoring the tight loop that operates on lines without a performance test to make sure refactoring does not reduce speed significantly.

DimitriPapadopoulos:
Apart from the missing documentation which we can add later on, LGTM.

DimitriPapadopoulos:
@larsoner Would you mind having a look? This pull request has been sitting in the waiting queue since July 2022, I would recommend we merge despite some issues we can fix later (such as missing documentation).

peternewman:
> TL;DR The only better pragma is the universal understood and available `sic!`.

Pedantically in prose I'd say sic specifically doesn't include a correction, but I get your point.

If you'd used practically any other word, I wouldn't really have cared, I'm just trying to consider future-proofing and where else we might go. Note how in comparison Clang says this in their docs ( https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas ):

> Clang supports GCC’s pragma for compatibility with existing source code, so #pragma GCC diagnostic and #pragma clang diagnostic are synonyms for Clang.

It's just the fact you've effectively redefined what ignore means within inline comments (from what cspell picked).

> Yours is also weak as it constantly compares to a different tool/community.

I'm trying to enable compatibility (or at least not direct clashes) between tools/communities. As different tools will have different features and benefits (for example spellintian checks for double words, although does fall down on nested if statements).

> What are actually the _strong_ arguments for "codespell:ignore"?
> 
>     * It was the natural idea of someone, back then in the issue, that did something
>       in me so I did the implementation.
> 
>     * It is shorter.

But sic suits that too.

>     * It reads like the imperative form: "Codespell, ignore frind!".  That comes just
>       natural.

Again sic would work for that.

>     * The idea is out for 2 yars, and people use the hack from the original
>       issue already.  (codespell:ignore yars)

But they can easily change with a find-replace. Once this is in a release version its baked in forever or a lot harder to change.

kaste:
To be fair:  I don't think I'm getting your point.  I just don't, maybe.

`cspell:ignore finds` and `codespell:ignore fends` don't mean the same thing, d'accord? But 
they're also literally not the same thing, verbatim.  The name of the action *is* the
same, but not the namespace/product.  

These are also unlikely to be placed nearby, or even on the same line.  As one means:  
ignore some words in this file, the other one means treat (something on) this very line 
differently.

Is it that, that the users don't get that right?  Or is it that we want in the future
what they already have and then we need to name it e.g. `codespell:ignore-in-this-file`
or `codespell:additional-words`?

Obviously, it is unlikely cspell will ever implement `codespell:ignore` verbatim. (Although 
they also claim the generic `spellchecker` namespace.)  If they ever get the same functionality
is predicting the future.  In the last 2 years since the issue or then this PR is open, they at 
least did not!  (They have an open issue, just googled it, at 
https://github.com/streetsidesoftware/cspell/issues/2799 though)

But if two things/communities should not clash, they must actually cooperate.  I remember that 
e.g. mypy and pyright diverged and they had to actually put the effort in it and talk to get 
agreement on the `type: ignore[rule]` form.


> But they can easily change with a find-replace. Once this is in a release version its baked in forever or a lot harder to change.

No.  It is now a find and replace as it would be then.  We don't have to take the burden of 
getting this right.  If it sucks then we can change this. 


> sic!

There was actually never a competition for the nicest, best looking pragma as you expressed concerns
about false positives very early.  And a verbatim `(sic!)` is then nothing to start with.  

Now `sic!` typically means _intentional misspelling_.  But since codespell is not a reader but a tool
`ignore` also just means that the tool is wrong.  Other than that "sic" is the nicest, 

```py
    # sic! /frinds, aggree/
```

although might be too subtle as one might not expect a tool reading it.  (Because our tools are still stupid
and can't read, lol.)  


I would be open to make a PR that implements `cspell:ignore` and `cspell: disable-line`.  Has that
been requested?

DimitriPapadopoulos:
What's wrong with using the `codespell` namespace and specifically `codespell:ignore` for now? Then we can also accept the `spellchecker` namespace in the future, with `spellchecker:whatever` directives we can agree on.

peternewman:
> `cspell:ignore finds` and `codespell:ignore fends` don't mean the same thing, d'accord? But they're also literally not the same thing, verbatim. The name of the action _is_ the same, but not the namespace/product.

Yes, I agree with that, although it would certainly be confusing if you wished to use both tools and had different flags for the same behaviour on the same line.

> Obviously, it is unlikely cspell will ever implement `codespell:ignore` verbatim.

Yes, I'd agree with that.

> (Although they also claim the generic `spellchecker` namespace.)

It's primarily that. As I see it, the main value in all these tools (or at least those with curated dictionaries like codespell) is the dictionaries. Then it's the different features the different tools offer to make use of them. cspell seem to have a good range of curated dictionaries:
https://cspell.org/docs/dictionaries/

> But if two things/communities should not clash, they must actually cooperate.

I don't think they have to, I suspect clang just followed the gcc docs to implement their compatible options.

> I remember that e.g. mypy and pyright diverged and they had to actually put the effort in it and talk to get agreement on the `type: ignore[rule]` form.

And this is sort of the crux of my argument, why are we starting off by diverging for the sake of the wording of the option, wouldn't it be much more sensible to start either aligned (or at least different), rather than redefining red as blue.

> No. It is now a find and replace as it would be then. We don't have to take the burden of getting this right. If it sucks then we can change this.

Maybe it's just a personal thing, but assuming there hasn't been an actual fork as such, then at the moment a few people are running a non-standard build with some particular functionality. Once it's in a released version then we're sort of stuck with it forever. Unless we add options to run both versions if we change, then we'd be forcing people to change their whole codebase when we do a presumably major release, what about if they've got multiple branches (e.g. beta and dev, etc etc).

Something like this:
https://en.wikipedia.org/wiki/Dagen_H

Compared to this:
https://en.wikipedia.org/wiki/PhONEday

One is an order of magnitude easier to manage for the end user.

> I would be open to make a PR that implements `cspell:ignore` and `cspell: disable-line`. Has that been requested?

As I'd imagine it, in no particular order we might do the following, none of which feels like a huge leap to me:
- Support more of their inline options (as you mention disable-line and ignore we already have CLI behaviours for, I guess their ignore might be a bit more challenging to code if it's not right a the top of the file)
- Maybe add some additional functionality they support which we don't currently have via the CLI (like disable-next-line), although arguably because they don't make sense in the CLI context
- Support a wider range of namespaces (either just spellcheck or even mirror cspell behaviour too)

My point is that the multi-namespace thing is significantly easier to code, and to support and for users to understand if we're doing something akin to PhONEday than Dagen H, and in fact it would be easier than PhONEday, because to take the analogy, you could then dial EITHER of the two options. If we got our in-line options to be feature complete with them, we get an audience for free, as they can just run codespell on their existing files without needing to do anything.

> What's wrong with using the `codespell` namespace and specifically `codespell:ignore` for now? Then we can also accept the `spellchecker` namespace in the future, with `spellchecker:whatever` directives we can agree on.

Just simply because we then have to code and the user has to understand that `codespell:ignore`=`spellchecker:whatever` that `codespell:foo`=`spellchecker:ignore` and that `codespell:ignore`!=`spellchecker:ignore`. I personally don't really understand why we want to add that complexity.

12rambau:
ok then I'll ask the question, what is the point of using both `cspell` and `codespell` in the same project? And who is going to use it? 

I am a **Python** developer, I came to this repository because it's a **Python** implementation of something really helpful and because it's supported by the `pre-commit` tool. It's seemed developed toward the **Python** community as illustrated by the support of the different existing config files like `pyproject.toml`. So IMHO you try to align with a tool that I (as a **Python** user) don't know and will never use and by doing so you want to impose wording that is not aligned with all the other tools that I know and use along with `codespell`, namely: `black`, `ruff`, `mypy`, `isort`, `flake8`. 

I see that you are very aware of `cspell` because you are part of the spell-checker community but we are not. I think you should do as `mypy` and `pyright`: align when the users start complaining and find common ground by discussing. It's currently the reverse we are complaining because this PR is stuck for consideration outside of the realm of our usage.

kaste:
So that's it then.  That is especially sad as we/the community only came up with this idea, and there 
is no better one as of now.  I esp. came here from a restrictive regex implementing the feature to a 
very loose one that would fit aesthetically in prose.  (We can see this above that I'm approaching 
the problem more from the universal `sic!` designator.)

I do amit that from all the options `cspell` here has I (sadly) only find `cspell:ignore` (yeah) 
interesting, as it can be placed anywhere in the file and then holds additional words in the scope 
of that file. 🤨 

I just make two comments.  The first one has the essential tip and trick for anyone searching for the
feature.  

 > Maybe it's just a personal thing, but assuming there hasn't been an actual fork as such, then at the moment a few people are running a non-standard build with some particular functionality.

There is no fork, the recommended usage (hack?) currently is `--ignore-regex=".*codespell:ignore.*"`.
You can use variants of this obviously to be more restrictive.  The approach here differs from
the PR as I always thought that this should accept concrete misspellings to ignore.  The above regex
can only ignore the whole line. 

 > Just simply because we then have to code and the user has to understand that codespell:ignore=spellchecker:whatever that codespell:foo=spellchecker:ignore and that codespell:ignore!=spellchecker:ignore. I personally don't really understand why we want to add that complexity.

Hm, to be fair, we usually document that we have `codespell:foo` and `codespell:fox` -- there must be _something_
that we own, something that mirrors our perspective to the world -- and what they do.  And then say: 
"We also read and support cspell:whatever. *See their documentation on what that means.*"

That's not in itself complex as it is how "translation" works in general.  Our perspective of the world 
especially includes that we don't begin with dictionaries that tell us which words are *allowed* like 
all the other tools.  Instead we start with common typing errors; we don't state that the writer doesn't 
know how to spell a word, we just know that typing on a keyboard is very error-prone.

DimitriPapadopoulos:
@peternewman I think we should merge this. This pull request has been stuck for more than a year. Whenever changes have been asked, @kaste has been prompt to take them into account and apply changes or provide answers, but the issue subsequently remained stuck on "_change requested_" for months. I think the pull request is good enough as it is: converging on finalising a pull request in finite time is more important than getting it perfect. I suggest we use `codespell:ignore` for now, as `codespell:ignore: words` aligns perfectly with the way linters work. Also, I am not convinced that file scope as opposed to line scope is a good choice for `cspell:ignore`, it should arguably have been `cspell:file-ignore` for file scope and `cspell:ignore` for line scope. Any way, we can work with CSpell on common directives later on.

kaste:
Ridiculous... the ghosting starts again.

12rambau:
I think that's the first time I see something so sad on GitHub, @kaste proposed a solution to a community problem, sticked to it for more than a year, answered ALL your requests and now because of a single maintainer stubborness, he closed it. 
Just want to mentioned that this PR has 15 hearts on the first message. I consider it to be huge for a PR in general and for a codespell tool in particular.
@kaste please don't delete your branch I'll continue using, it works exactly as it should be.

DimitriPapadopoulos:
I'll have a look eventually.

DimitriPapadopoulos:
@kaste Thank you for your contribution and your patience.

Some documentation would be welcome, but I don't dare asking for more.

kaste:
Haha, it's out.  🚢 ⛵

PR #2401, opened 2022-07-15
'analyses' is the plural of 'analysis'

quyykk:
As the title states, 'analyses' is the correct plural of 'analysis' in American English. As such, it shouldn't get flagged.

Unfortunately, this does mean that it won't flag usage of 'to analyze' in the third person, but I'm not sure how much can be done about that.

### DIFF
--- a/codespell_lib/data/dictionary_en-GB_to_en-US.txt
+++ b/codespell_lib/data/dictionary_en-GB_to_en-US.txt
@@ -9,7 +9,6 @@ analyse->analyze
 analysed->analyzed
 analyser->analyzer
 analysers->analyzers
-analyses->analyzes
 analysing->analyzing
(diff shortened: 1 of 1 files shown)


vikivivi:
Agreed with @quyykk This is Inline with varcon as well.
https://github.com/en-wl/wordlist/blob/f6b071244ce65b0a95ae052043042e85eed72815/varcon/varcon.txt#L922

For en-US_to_en-GB: (analyzes->analyses)
A C: analyzes / B Cv: analyses

For en-GB_to_en-US:  (No conversion required for `analyses`)
A B: analyses | -- pl: analysis

luzpaz:
The issue is we should have a blacklist of words like this so they don't get re-added at a later date

PR #2405, opened 2022-07-17
More typos

DimitriPapadopoulos:
See https://github.com/Blosc/bloscpack/pull/117.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -9594,6 +9594,7 @@ delgation->delegation
 delgations->delegations
 delgator->delegator
 delgators->delegators
+deliberatey->deliberately
 deliberatly->deliberately
 deliberite->deliberate
 deliberitely->deliberately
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -18962,6 +18964,7 @@ laods->loads
 Laotion->Laotian, lotion,
 Laotions->Laotians, lotions,
 laout->layout
+larer->larger
```
luzpaz, line 18967: Please also add `layer`  
`larer->larger, layer,`

DimitriPapadopoulos: `Y` is not near `R` on (Qwerty) keyboards. What are the odds of typing `larer` instead of `layer`?

luzpaz: Please add both. TIA

DimitriPapadopoulos: Done.

PR #2416, opened 2022-07-27
Add vulnreabilit(y|ies)->vulnerabilit(y|ies) correction.

cfi-gb:
See e.g.:

https://grep.app/search?q=vulnreability


vikivivi (approved):
LGTM

PR #2421, opened 2022-07-29
Add entries in to rare dictionary

luzpaz:
Split off from #2266


codespell_lib/data/dictionary_rare.txt
```diff
@@ -190,6 +193,7 @@ wee->we
 whats->what's
 whet->when, what, wet,
 whiling->while
+whit->with, white, wit,
```
vikivivi, line 196: Suggested word `wit` for `whit->with, white, wit,` has a misspell entry `wit->with` in the rare dictionary. This causes the checks to fail.

peternewman: No, I think it's a valid suggestion, it's just like some others where we need to deal with these circular edge cases, probably by updating our check config...

peternewman: @luzpaz do you want to do a new PR with the other suggestions just leaving this one separately as one for us to address later?

Or even with @vikivivi 's shorter suggested list. But I'd like to get whit->wit in ultimately as a suggestion too.

DimitriPapadopoulos: @luzpaz Perhaps teh suggestion to remove `whit->with, white, wit,` from this merge request and create a specific one for that case is preferable, isn't it?

Issue #1193, opened 2019-07-19
Possible incorrect values and disable multiple choice


brendanator:
Below is a list of words that appear on both the left and right hand sides of the dictionary. I assume that most of these are errors.

I spotted this as I'm using `minimise` (English spelling) and it complained even though it is a correct spelling on the right hand side. Is there an option to disable suggestions when there is more than one choice?

```
$ rg --pcre2 '^([a-z]*)->.*\b\1\b' data/dictionary.txt
1931:backword->backward, backword,
1932:backwords->backwards, backwords,
2445:calculatable->calculatable, calculable,
2951:circularly->circular, circularly,
3252:commend->commend, comment, command,
3253:commends->commends, comments, commands,
4245:copyable->copyable, copiable,
4456:crate->create, crate,
4457:crated->created, crated,
4484:cretaing->cretaing
4488:cristal->crystal, cristal,
4686:dealign->dealing, dealign,
4947:degrate->degrate, degrade,
4948:degrates->degrates, degrades,
7664:followings->followings, following,
7702:forewarded->forewarded, forwarded,
7732:formule->formula, formulas, formule,
7780:fount->fount, found,
8080:gird->grid, gird,
8081:girds->grids, girds,
8506:hove->hove, have, hover, love,
8910:inactivate->inactivate, deactivate,
9410:inly->only, inly,
9984:itteration->itteration
10168:leaded->led, lead, leaded,
10262:lightening->lightening, lightning, lighting,
10326:liszt->list, liszt,
10357:loafing->loading, loafing,
10409:lousily->lousily, loosely,
10613:mater->matter, master, mother, mater,
10731:memorise->memorise, memorize,
10881:minimise->minimise, minimize,
11097:mould->mold, mould, module,
11619:ocorrence->ocorrence
11620:ocorrences->ocorrences
12161:panting->panting, painting,
13399:prosses->process, processes, possess, prosses,
13499:purportive->supportive, purportive,
13628:readd->readd, re-add, read,
13629:readded->readded, read,
13660:ream->ream, stream,
14678:retuned->retuned, returned,
14718:revered->revered, reversed,
14879:sate->state, sate,
15391:sinc->sinc, synch, sync, sink, since,
15394:sincs->sincs, syncs, sinks, since,
15397:singe->singe, single,
15398:singed->signed, singled, singed,
15867:sting->string, sting,
15869:stings->strings, stings,
16286:suppressable->suppressable, suppressible,
[...]

PR #2422, opened 2022-08-01
Change "circularly" to "circulary".

vinc17fr:
In `codespell_lib/data/dictionary_rare.txt`, one has: `circularly->circular`. However, the word "circularly" is not rare, and a mistake for "circular" is improbable (difference of 2 letters). I suspect that it is an error coming from commit d978da6e3143d2fee553e836a0b0ee46a68bd900 for the fix of issue #1193, where one previously had
```
circularly->circular, circularly,
```
while this should have been
```
circulary->circular, circularly,
```
as the adjective "circulary" exists (1913 Webster), but is rare and obsolete (the adjective "circular" is now used instead).

### DIFF
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -26,7 +26,7 @@ cant->can't
 chack->check, chalk, cheque,
 chancel->cancel
 chancels->cancels
-circularly->circular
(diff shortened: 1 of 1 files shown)


vikivivi (approved):
LGTM

luzpaz (approved):

PR #2457, opened 2022-08-17
'which' -> 'that' because the clause is essential to the meaning

mdeweerd:
A random page explaining when to use "which" or "that": https://prowritingaid.com/art/438/-Which--or--That-%3A-Know-When-to-Use-Each.aspx .


peternewman (approved):
Thank for correcting my grammar! :smile:

PR #2459, opened 2022-08-18
Add pre-commit configuration

mdeweerd:
This proposes a pre-commit configuration.

Running pre-commit with this setup will modify (format, optimize) source code and also report a number of notifications preventign commit if you install pre-commit on the repository.

You can run pre-commit on all files without committing using `pre-commit run -a` .


setup.cfg
```diff
@@ -1,3 +1,43 @@
 [flake8]
 max-line-length = 88
 extend-ignore = E203
+exclude = build, ci-helpers
+ignore =
```
larsoner, line 5: this seems to have caused a bunch of new inline errors, probably worth reverting

setup.cfg
```diff
@@ -1,3 +1,43 @@
...
+
+[tool:pytest]
```
larsoner, line 7: let's put this stuff in `pyproject.toml` instead, ideally we'd get rid of `setup.cfg`

larsoner (commented):
Can you iterate until CIs are green? I think ideally we wouldn't need to change any existing code with the given pre-commit configuration. We should make any changes to existing flake/style stuff in a separate PR probably.

setup.cfg
mdeweerd: I just tried to move the codespell configuration to pyproject.toml, but it only understands the configuration in setup.cfg.

```
[tool.codespell]
builtin=clear,rare,informal,usage,code,names
ignore-words-list=uint
skip=./.*,codespell_lib/data/*,codespell_lib/tests/*

```

setup.cfg
mdeweerd: Apparently the "ignore" should not be overriden, and the exclude is no longer needed so I'll revert that.

setup.cfg
mdeweerd: mypy provides guidance about pyproject.toml, but this pyproject.toml configuration was not picked up by codespell either:

```
[tool.codespell]
builtin=["clear","rare","informal","usage","code","names"]
ignore-words-list="uint"
skip=["./.*","codespell_lib/data/*","codespell_lib/tests/*"]
```

larsoner:
> I did not correct the following notification when running pre-commit when it runs mypy as it would require fixing the source itself:

Feel free to make this change. We need to get CIs green

setup.cfg
mdeweerd: The additional dependency 'tomli' is required:
```yaml
  - repo: https://github.com/codespell-project/codespell
    rev: v2.2.2
    hooks:
      - id: codespell
        additional_dependencies:
          - tomli
```

codespell now complains:
```
ERROR: Unknown builtin dictionary:  'code'
```

With the following toml configuration:
```
[tool.codespell]
builtin = ["clear","rare","informal","usage","code","names"]
ignore-words-list = ["uint"]
skip=[ "./.*","codespell_lib/data/*","codespell_lib/tests/*"]
```

It did not complain about that with the setup.cfg configuration:
```
[codespell]
builtin=clear,rare,informal,usage,code,names
ignore-words-list=uint
skip=./.*,codespell_lib/data/*,codespell_lib/tests/*
```

mdeweerd:
> > I did not correct the following notification when running pre-commit when it runs mypy as it would require fixing the source itself:
> 
> Feel free to make this change. We need to get CIs green

That is not what is turning the tests red.
https://github.com/codespell-project/codespell/actions/runs/3642917335/jobs/6150567655

My guess is that during the testcase the setup.cfg is read and this is modifying the behavior.

So either I move the codespell configuration to another toml file, or the tests would need to reference a forced configuration file.
I'll go with a separate codespell configuration for the moment to see what that gives and then - at a later time - the tests could be updated.

setup.cfg
mdeweerd: Actually, codespell does not support the lists in pyproject.toml while other tools do ..
This is understood:
```
[tool.codespell]
#builtin = ["clear","rare","informal","usage","code","names"]
builtin = "clear,rare,informal,usage,code,names"
#ignore-words-list = ["uint"]
ignore-words-list = "uint"
#skip=[ "./.*","codespell_lib/data/*","codespell_lib/tests/*"]
skip="./.*,codespell_lib/data/*,codespell_lib/tests/*"
```

mdeweerd:
How can I fix this?
```
missing from sdist:
  pyproject-codespell.precommit-toml
make: *** [Makefile:48: check-manifest] Error 1
Error: Process completed with exit code 2.
```

setup.cfg
larsoner: > Actually, codespell does not support the lists in pyproject.toml while other tools do ..

Feel free to improve the support here if you want. But it'll require new test cases and debugging, so maybe better in a separate PR

I pushed a commit to fix `make check-manifest`

mdeweerd:
@larsoner Thanks, I overlooked that MANIFEST.in somehow.. .

I ignore some more errors for mypy in the pre-commit run to avoid detecting more/less than the CI flow.

I also applied the pyupgrade changes so that it is all done.

pyproject.toml
```diff
@@ -90,5 +98,45 @@ pretty = true
 show_error_codes = true
 strict = true
 
+[tool.pylint]
+reports=false
+py-version=3.6
```
larsoner, line 103: Shouldn't this be at least 3.8 since that's our min now (or maybe it's 3.7, can't remember...)?

pyproject.toml
```diff
@@ -82,6 +84,12 @@ codespell_lib = [
 [tool.check-manifest]
 ignore = ["codespell_lib/_version.py"]
 
+# TODO: reintegrate codespell configuration after updating test cases
+#[tool.codespell]
+#builtin = ["clear","rare","informal","usage","code","names"]
+#ignore-words-list = ["uint"]
+#skip=[ "./.*","codespell_lib/data/*","codespell_lib/tests/*"]
```
larsoner, line 91: Is this for this PR, or a follow-up, or is this comment outdated and can be removed, or ...?

larsoner (commented):
Just a couple of minor things then I think it's good to go in!

pyproject.toml
mdeweerd: I left it in pyproject.toml as a reminder tat IMHO the configuration should be ultimately there in that format.

Currently there are 2 issues:
1. The format is not recognized by codespell (see the other toml file);
2. The configuration impacts the test - which means that the tests are subject to the project codespell configuration itself.  So ideally the tests should not rely on the codespell configuration in the project's setup.cfg, pyproject.toml, ... .

pyproject.toml
mdeweerd: The CI flow is using 3.7.  I started this a long time ago, 
3.7 is not end fo life yet - I do not know what the policy is for codespell https://devguide.python.org/versions/ .

larsoner: Let's at least bump to 3.7 then

mdeweerd:
I managed to get rid of the setup.cfg and move the flake8 configuration to pyproject.toml (and I moved more configurations from pre-commit-config to pyproject.toml).

pyproject.toml
mdeweerd: Bumped.

larsoner (approved):

larsoner:
Thanks @mdeweerd !

mdeweerd:
Good, I improved the pre-commit configurat a bit just now though.

larsoner:
Feel free to open another PR. Bonus points if you want to add list-of-str support to our TOML as well!

PR #2462, opened 2022-08-18
MAINT: 2.2.1

larsoner:


larsoner:
I have never actually heard the word "complier" so I left that, but I see "complies" and "complied" plenty so I removed those

PR #2464, opened 2022-08-18
Remove legacy python 2 `object` subclassing

kkirsche:
This removes unnecessary subclassing of object as this library supports Python 3.6+

This stuff is a holdover from:
https://docs.python.org/2/reference/datamodel.html#new-style-and-classic-classes

Python 3 doesn't have any of this, so there is no benefit to doing this

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -230,7 +230,7 @@ class NewlineHelpFormatter(argparse.HelpFormatter):
 
     def _split_lines(self, text, width):
         parts = text.split('\n')
-        out = list()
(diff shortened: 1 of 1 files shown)


peternewman (approved):
Seems fairly sensible from the docs, but hasn't had much impact on me personally...

luzpaz:
@peternewman bump

DimitriPapadopoulos:
While the request does not fix any bug, it is not without impact: sticking to Python 3 documentation will ease future maintenance.

Duplicate of #2085.

jdufresne (approved):

PR #2466, opened 2022-08-18
Update README.rst

### LINKED ISSUE #2451: Issue with --ignore-words?
image

I feel like things aren't working as they should

image

andyholmes:
Update the README to elaborate on case-sensitivity in ignored words
and dictionaries.

close #2451

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -55,19 +55,19 @@ Some noteworthy flags:
 
     codespell -w, --write-changes
 
-The ``-w`` flag will actually implement the changes recommended by codespell. Running without the ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.
+The ``-w`` flag will actually implement the changes recommended by codespell. Running without the ``-w`` flag is the same as doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.
 
 .. code-block:: sh
 
     codespell -I FILE, --ignore-words=FILE
 
-The ``-I`` flag can be used for a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said list of allowed words. **Important note:** The list passed to ``-I`` is case-sensitive based on how it is listed in the codespell dictionaries.
+The ``-I`` flag can be used for a list of certain words to allow that are in the codespell dictionaries. The format of the file is one word per line. Invoke using: ``codespell -I path/to/file.txt`` to execute codespell referencing said list of allowed words. See `Ignoring Words`_ for more details.
 
 .. code-block:: sh
 
     codespell -L word1,word2,word3,word4
 
-The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it.  **Important note:** The list passed to ``-L`` is case-sensitive based on how it is listed in the codespell dictionaries.
+The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it.  See `Ignoring Words`_ for more details.
 
 .. code-block:: sh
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -41,7 +41,11 @@ Some noteworthy flags::
 
     codespell -w, --write-changes
 
-The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
+The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.
+
+When ignoring false positives, note that allowed words are *case-sensitive*, but dictionary matches are *case-insensitive*. For example, the dictionary entry ``wrod`` will match ``Wrod``, but to allow it you must pass ``wrod``.
```
peternewman, line 46: How about this tweak? Do you think this helps, or doesn't hinder the clarity?
```suggestion
When ignoring false positives, note that spelling errors are *case-insensitive* but words to ignore are *case-sensitive*. For example, the dictionary entry ``wrod`` will also match the typo ``Wrod``, but to ignore it you must pass ``wrod``.
```

README.rst
```diff
@@ -41,7 +41,11 @@ Some noteworthy flags::
 
     codespell -w, --write-changes
 
-The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.::
+The ``-w`` flag will actually implement the changes recommended by codespell. Not running with ``-w`` flag is the same as with doing a dry run. It is recommended to run this with the ``-i`` or ``--interactive`` flag.
```
peternewman, line 44: Currently we've generally got
--foo
Explanation of foo.

You've added some comments above the flag too. I would suggest for consistency/clarity either we should move this below too, or move ALL the explanations above the flags. Afterwards certainly seems to match the man pages I've looked at.

peternewman (changes requested):
Some potential enhancements.

README.rst
andyholmes: This is much better!

README.rst
andyholmes: I'd agree it's a bit out of place. I went for above, just because it applies to both `-I` and `-L`.

I guess the obvious options are to either duplicate it for both, or maybe just move it below the `-L` entry. In that case the example `wrod` could be changed to `wrod1`, which might be a little more obvious?

peternewman: Maybe for the readme particularly, we should rewrite it as:

> **Ignoring Words**
> When ignoring false positives, note that spelling errors are *case-insensitive* but words to ignore are *case-sensitive*. For example, the dictionary entry ``wrod`` will also match the typo ``Wrod``, but to ignore it you must pass ``wrod``.
> 
> The words to be ignore can be passed in two ways:
> -I
> A file with a word per line to ignore.
> -L
> A comma separated list of words to ignore on the command line.

Or something like that? I think there's a similar case with something else further down.

andyholmes: Not a bad idea, and leaves it open to add more help if necessary. I sort of back-pedaled on the earlier changes, then added a section a bit farther down with internal links.

Let me know what you think :)

prestoncarman:
As I often have trouble understanding the ignore list and explaining it to new users, I hope this change can get merged soon to help clarify. @peternewman and @andyholmes

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
@andyholmes and @larsoner could you have a look before merging?

andyholmes:
> @andyholmes [...[ could you have a look before merging?

LGTM

PR #2468, opened 2022-08-19
move `ro` to code dictionary

robin-wayve:
As in `mount -o ro` for "read-only".


peternewman (approved):
Sounds very sensible.

Issue #2470, opened 2022-08-20
Different suggested fix wrt capitalization


dforsi:
I'm getting two different suggested fixes for what seems to be the same misspelling with codespell 2.1.0:
```
$ codespell lesstif.txt
lesstif.txt:1: Lesstiff ==> Lesstif
lesstif.txt:2: LessTiff ==> lesstif
```

See attachment [lesstif.txt](https://github.com/codespell-project/codespell/files/9386660/lesstif.txt)

BTW, according to [Wikipedia](https://en.wikipedia.org/wiki/LessTif) the correct capitalization is LessTiff

peternewman:
So there's two things driving this, firstly do you want to correct the dictionary correction so it's in the correct case:
https://github.com/codespell-project/codespell/blob/101498daa5ebd2108518c4f7ff7c6d0d2520c9ec/codespell_lib/data/dictionary.txt#L20034

Secondly the algorithm that tries to predict what case the correction should be in:
https://github.com/codespell-project/codespell/blob/14af437206c82597d98d844e610741bb7664c902/codespell_lib/_codespell.py#L496-L503

Obviously this can't currently deal with camel case, where your first line it capitalised, it offers that, otherwise it gets confused and gives up.

I think we've generally gone for the typo being in lower case, but actually if we stored this typo in camel case and then made sure it was `lower`ed throughout the codebase, and added a `elif word == word.lower():` check here, I think that would sort out this behaviour (unless I've missed something or forgotten some reasoning around storing the typos in lower case)...

vikivivi:
@peternewman @luzpaz 
Please find my patches for handling capitalization issue.
1. Do not change suggested words in dictionary to lower case during loading.
2. Capitalization decision is decided in fix_case().
3. Hope this handle proper nouns, camelCase, capitalise word.

```diff
diff --git a/codespell_lib/_codespell.py b/codespell_lib/_codespell.py
index 1ed70e89..f0607c1e 100644
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -454,10 +454,10 @@ def build_dict(filename, misspellings, ignore_words):
     with codecs.open(filename, mode='r', encoding='utf-8') as f:
         for line in f:
             [key, data] = line.split('->')
-            # TODO for now, convert both to lower. Someday we can maybe add
-            # support for fixing caps.
+            # Convert key to lower case.
+            # Do not modify data to lower case. Leave it as per dictionary.
[...]

PR #2471, opened 2022-08-20
Capitalize LessTif

dforsi:
This PR changes the replacement word from lesstif to LessTif as in https://en.wikipedia.org/wiki/LessTif

As discussed in PR #2470.


peternewman (approved):
Thanks.

peternewman:
We'll need to try and do the other stuff in there reasonably soon so we don't have unintended consequences here...

PR #2486, opened 2022-08-31
Add spelling corrections for install and variants.

cfi-gb:
See e.g.:
- https://grep.app/search?q=instlal
- https://github.com/search?q=instlal&type=code


luzpaz:
Thanks!

PR #2489, opened 2022-09-02
Allow `Bloc`

tenshiAMD:
Vist the link for reference - https://www.merriam-webster.com/dictionary/bloc


peternewman:
Hi @tenshiAMD

These are already in the rare dictionary:
https://github.com/codespell-project/codespell/blob/3a1715444cf97288d6c83d927e1c7898d4b87c2a/codespell_lib/_codespell.py#L55

So you need to either ignore the lines or the words across your whole codebase.

All the other words in there are in the same scenario, but for most people when they write bloc they actually mean block, hence it's in rare.

PR #2495, opened 2022-09-08
move `warmup` to code dictionary

robin-wayve:
I think `warmup->warm up` is valid for prose, but it seems common in code without being strictly mispelled: https://github.com/search?q=warmup&type=code

The [Java Microbenchmark Harness](https://github.com/openjdk/jmh), for example, has a `@Warmup` annotation: https://javadoc.io/doc/org.openjdk.jmh/jmh-core/latest/org/openjdk/jmh/annotations/Warmup.html


peternewman (approved):
Seems sensible thanks Robin.

PR #2496, opened 2022-09-10
Use `grep -E` instead of `egrep`

a1346054:
New version of GNU egrep now emits a deprecation message to stderr on use. This PR fixes that.


peternewman (approved):
Agreed that these two lines are identical thanks.

PR #2515, opened 2022-10-05
DOC: Better docs

larsoner:
@peternewman this should address some of your comments at least

Feel free to open a PR where, if `pyproject.toml` is detected in `cwd`, it tries to use it. But I don't think it should raise an error or print anything (because doing either could mess up existing workflows).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,7 +107,7 @@ This is equivalent to running::
 
     codespell --quiet-level 3 --count --skip "*.po,*.ts,./src/3rdParty,./src/Test"
 
-Now codespell also support ``pyproject.toml``::
+Now codespell also support ``pyproject.toml`` via the ``--toml pyproject.toml`` argument::
 
     [tool.codespell]
     skip = '*.po,*.ts,./src/3rdParty,./src/Test'
--- a/README.rst
+++ b/README.rst
@@ -120,7 +120,7 @@ This is equivalent to running::
 Any options specified in the command line will *override* options from the
 config files.
 
-.. _tomli:: https://pypi.org/project/tomli/
+.. _tomli: https://pypi.org/project/tomli/
 
 Dictionary format
 -----------------
--- a/README.rst
+++ b/README.rst
@@ -103,19 +103,24 @@ be specified in this file (without the preceding dashes), for example::
     count =
     quiet-level = 3
 
-This is equivalent to running::
-
-    codespell --quiet-level 3 --count --skip "*.po,*.ts,./src/3rdParty,./src/Test"
-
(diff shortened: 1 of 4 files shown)


larsoner:
> Feel free to open a PR where ...

Never mind @peternewman , I ended up just implementing it here. Feel free to have a look and mark for merge if you're happy!

README.rst
```diff
@@ -103,19 +103,24 @@ be specified in this file (without the preceding dashes), for example::
     count =
     quiet-level = 3
 
-This is equivalent to running::
-
-    codespell --quiet-level 3 --count --skip "*.po,*.ts,./src/3rdParty,./src/Test"
```
peternewman, line 108: Did you intend to remove this?

README.rst
```diff
@@ -103,19 +103,24 @@ be specified in this file (without the preceding dashes), for example::
...
+(or a path can be specified via ``--toml <filename>``) file, and the
+``[tool.codespell]`` entry will be used as long as the tomli_ package
+is installed, for example::
 
     [tool.codespell]
     skip = '*.po,*.ts,./src/3rdParty,./src/Test'
     count = ''
     quiet-level = 3
 
+This is equivalent to running::
```
peternewman, line 116: Or maybe this should be:
```suggestion
These are both equivalent to running::
```

codespell_lib/_codespell.py
```diff
@@ -407,14 +407,22 @@ def parse_options(args):
...
-            raise ImportError(
-                f'tomli is required to read pyproject.toml but could not be '
-                f'imported, got: {exc}') from None
-        with open(options.toml, 'rb') as f:
+            if raise_error:
+                raise ImportError(
+                    f'tomli is required to read pyproject.toml but could not '
+                    f'be imported, got: {exc}') from None
+            else:
+                continue
```
peternewman, line 424: Do we not have a verbose log option where we could write something at that level?

peternewman (approved):
Just a minor documentation related query.

peternewman:
Off the back of #2290

codespell_lib/_codespell.py
larsoner: We could, but I'm not sure we need to. Someone expecting it to "just work" who doesn't have `tomli` might be confused for a second, then they'll presumably try `--toml <whatever>` and see an error about import.

README.rst
larsoner: Yeah with the "these both" suggestion I think it's clearer why 👍

peternewman (approved):
LGTM

PR #2516, opened 2022-10-05
Migrate metadata from setup.py to setup.cfg

cclauss:
This pull request was created using https://pypi.org/project/setuptools-py2cfg plus manual modifications.

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use
> We recommend users expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the setup.py minimal with only the dynamic parts (or even omit it completely if applicable).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,56 @@
+[metadata]
+name = codespell
+version = attr: codespell_lib.__version__
+maintainer = Lucas De Marchi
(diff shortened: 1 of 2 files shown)


cclauss:
Closing in favor of #2517

PR #2517, opened 2022-10-05
Migrate metadata from setup.py and setup.cfg to pyproject.toml

cclauss:
Migrate to `pyproject.toml` using [`ini2toml`](https://pypi.org/project/ini2toml) to do the file conversion and running [`validate-pyproject`](https://github.com/abravalheri/validate-pyproject) to validate the results.
* Currently `flake8` is yet not compatible with `pyproject.toml` so move its config into `.flake8`


pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+]
+requires-python = ">=3.7"
+
+[project.readme]
+file = "README.rst"
+content-type = "text/x-rst"
+
+[project.urls]
+Homepage = "https://github.com/codespell-project/codespell/"
+Download = "https://github.com/codespell-project/codespell/"
```
peternewman, line 29: Should this be https://github.com/codespell-project/codespell/releases perhaps?
```suggestion
Download = "https://github.com/codespell-project/codespell/releases"
```

What do other people do?

peternewman (changes requested):
You need to fix our CI tests please which currently call setup.py to install itself.

cclauss:
> You need to fix our CI tests please which currently call setup.py to install itself.

Yes. These calls to `setup.py` are exactly what we are trying to [avoid](https://pip.pypa.io/en/stable/topics/secure-installs/#do-not-use-setuptools-directly).  This confusing little dance between `setup.py` and `make` increases the attack surface.

This pull request must remain in ___DRAFT___ status until someone can propose an effective fix for `make check-distutils`.  Unfortunately, my make-foo was not up to that task.

Makefile
```diff
@@ -48,7 +48,8 @@ check-manifest:
 	check-manifest
 
 check-distutils:
-	python setup.py check --restructuredtext --strict
+	echo "WARNING: 'make check-distutils' is disabled!!"
+	# python setup.py check --restructuredtext --strict
```
peternewman, line 52: Just adding a comment here to make sure we don't miss this, it will obviously be outdated when someone fixes it.

So in the same way as you changed the other commands, how do we run this command in the new format? If you can tell me on the command line, I should be able to make it work in the Makefile, i.e. I think it's a Python not a Make issue.

This page explains what it did, the correct answer might be that these particular commands are no longer required, but hopefully there's a similar set of tools to validate pyproject.toml we should switch to instead? It's interesting that we weren't validating the metadata itself before!

Makefile
```diff
@@ -47,8 +47,8 @@ trim-dictionaries:
 check-manifest:
 	check-manifest
 
-check-distutils:
-	python setup.py check --restructuredtext --strict
+check-pyproject:
```
peternewman, line 50: May as well leave some backwards compatibility...
```suggestion
check-distutils: check-pyproject

check-pyproject:
```

Makefile
```diff
@@ -2,11 +2,11 @@ SORT_ARGS := -f -b
 
 DICTIONARIES := codespell_lib/data/dictionary*.txt
 
-PHONY := all check check-dictionaries sort-dictionaries trim-dictionaries check-dictionary sort-dictionary trim-dictionary check-manifest check-distutils flake8 pytest pypi clean
+PHONY := all check check-dictionaries sort-dictionaries trim-dictionaries check-dictionary sort-dictionary trim-dictionary check-manifest check-pyproject flake8 pytest pypi clean
```
peternewman, line 5:
```suggestion
PHONY := all check check-dictionaries sort-dictionaries trim-dictionaries check-dictionary sort-dictionary trim-dictionary check-manifest check-pyproject check-distutils flake8 pytest pypi clean
```

peternewman (changes requested):
Sorry @larsoner has also done this (and some other stuff) in #2523 .

I don't know which one we merge/use first...

Makefile
cclauss: This would run the code twice under two different names.

larsoner:
> I don't know which one we merge/use first...

Argh sorry I didn't see this! #2523 also has CI infrastructure to make releasing (hopefully!) trivial, so I'd rather merge that one.

PR #2518, opened 2022-10-05
Add regresison->regression

janosh:


peternewman (changes requested):
Please undo the sort commit or use `make sort-dictionaries` to sort the dictionaries as we'd expect.

peternewman (approved):
LGTM thanks.

PR #2542, opened 2022-10-19
Add relative hidden directory and basic subdir tests

peternewman:
These pass already.


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
@peternewman Looks good. Should I (or you) squash into a single commit when merging?

peternewman:
> @peternewman Looks good.

Thanks @DimitriPapadopoulos 

> Should I (or you) squash into a single commit when merging?

I generally don't mind, but I appreciate others prefer things to be neater than editing within the web UI can result in so feel free to if you want.

PR #2552, opened 2022-10-23
DOC/ENH: Clarify usage or configuration files and log about it

### LINKED ISSUE #2534: Report when we're using setup.cfg, .codespellrc, pyproject.toml etc
identified in https://github.com/bids-standard/bids-specification/pull/1322#issuecomment-1280914648

How to reproduce:

- run a basic codespell GH action (without passing `with:` arguments)
- add a `.codespellrc` configuration file to the repo root
- the config file **is** being observed BUT
- the "argument summary" in the GH actions log is not reflecting that

image

sappelhoff:
closes #2534 

is this what you had in mind @larsoner?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,6 +117,17 @@ These are both equivalent to running::
 
     codespell --quiet-level 3 --count --skip "*.po,*.ts,./src/3rdParty,./src/Test"
 
+If several config files are present, they are read in the following order:
+
+#. ``pyproject.toml``
+#. ``setup.cfg``
+#. ``.codespellrc``
+#. any additional file supplied via ``--config``
+
+If a codespell configuration is supplied in several of these files,
+the configuration from the most recently read file overwrites previously
+specified configurations.
+
 Any options specified in the command line will *override* options from the
 config files.
 
--- a/README.rst
+++ b/README.rst
@@ -119,7 +119,7 @@ These are both equivalent to running::
 
 If several config files are present, they are read in the following order:
 
-#. ``pyproject.toml``
+#. ``pyproject.toml`` (only if the ``tomli`` library is available)
 #. ``setup.cfg``
 #. ``.codespellrc``
 #. any additional file supplied via ``--config``
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -423,8 +423,21 @@ def parse_options(args):
         with open(toml_file, 'rb') as f:
             data = tomli.load(f).get('tool', {})
         config.read_dict(data)
-    config.read(cfg_files)
 
+    # Report which config files are going to be used
+    used_cfg_files = []
+    for cfg_file in cfg_files:
+        _cfg = configparser.ConfigParser()
+        _cfg.read(cfg_file)
+        if _cfg.has_section('codespell'):
+            used_cfg_files.append(cfg_file)
+    if len(used_cfg_files) > 0:
+        print('Used config files:\n')
+    for ifile, cfg_file in enumerate(used_cfg_files):
+        print('    %i : %s' % (ifile, cfg_file))
+
+    # Use config files
+    config.read(cfg_files)
     if config.has_section('codespell'):
         # Build a "fake" argv list using option name and value.
         cfg_args = []
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -424,17 +424,13 @@ def parse_options(args):
             data = tomli.load(f).get('tool', {})
         config.read_dict(data)
 
-    # Report which config files are going to be used
+    # Collect which config files are going to be used
     used_cfg_files = []
     for cfg_file in cfg_files:
         _cfg = configparser.ConfigParser()
         _cfg.read(cfg_file)
         if _cfg.has_section('codespell'):
             used_cfg_files.append(cfg_file)
-    if len(used_cfg_files) > 0:
-        print('Used config files:\n')
-    for ifile, cfg_file in enumerate(used_cfg_files):
-        print('    %i : %s' % (ifile, cfg_file))
 
     # Use config files
     config.read(cfg_files)
@@ -458,7 +454,7 @@ def parse_options(args):
     if not options.files:
         options.files.append('.')
 
-    return options, parser
+    return options, parser, used_cfg_files
 
 
 def parse_ignore_words_option(ignore_words_option):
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -117,6 +117,17 @@ These are both equivalent to running::
...
+If several config files are present, they are read in the following order:
+
+#. ``pyproject.toml``
+#. ``setup.cfg``
+#. ``.codespellrc``
+#. any additional file supplied via ``--config``
+
+If a codespell configuration is supplied in several of these files,
+the configuration from the most recently read file overwrites previously
+specified configurations.
```
DimitriPapadopoulos, line 129: This is true in practice, but nowhere documented in theory:
https://github.com/python/cpython/issues/96765

codespell_lib/_codespell.py
```diff
@@ -423,8 +423,21 @@ def parse_options(args):
...
+        _cfg.read(cfg_file)
+        if _cfg.has_section('codespell'):
+            used_cfg_files.append(cfg_file)
+    if len(used_cfg_files) > 0:
+        print('Used config files:\n')
+    for ifile, cfg_file in enumerate(used_cfg_files):
+        print('    %i : %s' % (ifile, cfg_file))
+
+    # Use config files
+    config.read(cfg_files)
```
DimitriPapadopoulos, line 440: If you want documented/explicit order of priority for these config files, you' need to call `configparser.ConfigParser.read` sequentially on the config files, then merge the resulting configurations manually.

Don't get me wrong, I have nothing against using the undocumented/implicit priorities enforced by `configparser.ConfigParser.read`. But while can rely on these undocumented/implicit priorities to obtain the same undocumented/implicit priorities in codespell, we cannot document them in codspell if they are not documented in `configparser.ConfigParser.read`.

README.rst
sappelhoff: it is documented in https://docs.python.org/3/library/configparser.html

> It is possible to read several configurations into a single [ConfigParser](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser), where the most recently added configuration has the highest priority. Any conflicting keys are taken from the more recent configuration while the previously existing keys are retained.

codespell_lib/_codespell.py
DimitriPapadopoulos: I'm convinced by https://github.com/python/cpython/issues/96765#issuecomment-1288117030. :+1:

README.rst
DimitriPapadopoulos: I'm convinced by https://github.com/python/cpython/issues/96765#issuecomment-1288117030. :+1:

README.rst
```diff
@@ -117,6 +117,17 @@ These are both equivalent to running::
 
     codespell --quiet-level 3 --count --skip "*.po,*.ts,./src/3rdParty,./src/Test"
 
+If several config files are present, they are read in the following order:
+
+#. ``pyproject.toml``
```
cclauss, line 122: Only if tomli or tomllib are present?

sappelhoff: good catch, only if `tomli` is installed (tomllib is only python>3.11 and not conditionally imported in this module as far as I can see). I'll push a commit tomorrow.

cclauss: https://twitter.com/gvanrossum/status/1583561788204806144?s=20&t=-A2jdKe7r-ZgRUtSA8JZ_g

sappelhoff: documented in https://github.com/codespell-project/codespell/pull/2552/commits/2381c2b100a0d3dfde0455334228d5256f488244 -- adding `tomllib` would be more involved and should be done in a separate, dedicated PR.

README.rst
```diff
@@ -119,7 +119,7 @@ These are both equivalent to running::
 
 If several config files are present, they are read in the following order:
 
-#. ``pyproject.toml``
+#. ``pyproject.toml`` (only if the ``tomli`` library is available)
```
cclauss, line 122: or Python >= 3.11.

sappelhoff: Is there something I am missing? This wouldn't work: Python 3.11 ships the tomllib library, not tomli. Yet codespell only has code for tomli.

For Python 3.11 to "work out of the box", someone would have to make a PR that introduces a logic like:
- if Python version >= 3.11, use tomllib
- Else check if tomli installed, if yes: use it, if no: raise (or not)

or am I wrong?

see also what I wrote in https://github.com/codespell-project/codespell/pull/2552#discussion_r1003224716

cclauss: I was going from https://github.com/codespell-project/codespell/blob/master/pyproject.toml#L38 ... now I understand and will create a PR.

larsoner:
This would work. But instead of a direct `print` in the parsing stage, it's probably cleaner to pass it as a return value here

https://github.com/codespell-project/codespell/blob/91afcbf63764db98f5973f4f929603d0e1c473e2/codespell_lib/_codespell.py#L773

that way all `print`s are done within `main` so that we could for example triage based on QuietLevels at some point if we wanted

larsoner:
... and by "pass it", I mean "pass the list of config files parsed"

sappelhoff:
Thanks, I agree that's a good idea. Implemented in https://github.com/codespell-project/codespell/pull/2552/commits/9c34d92cead0e76c47ec53a0214b4e11686c1f02

larsoner:
Marking for merge when green, thanks in advance @sappelhoff !

sappelhoff:
Thanks -- but it'll need one more approving review to be auto merged :-)

larsoner (approved):

codespell_lib/_codespell.py
```diff
@@ -776,7 +785,13 @@ def _script_main():
 
 def main(*args):
     """Contains flow control"""
-    options, parser = parse_options(args)
+    options, parser, used_cfg_files = parse_options(args)
+
+    # Report used config files
+    if len(used_cfg_files) > 0:
```
DimitriPapadopoulos, line 791: Too late for this PR, but you could have used `if used_cfg_files` instead of `if len(used_cfg_files) > 0`.

Rationale from Deepsource.io:
> Using the `len` function to check if a sequence is empty is not idiomatic and can be less performant than checking the truthiness of the object.
> 
> `len` doesn't know the context in which it is called, so if computing the length means traversing the entire sequence, it must; it doesn't know that the result is just being compared to 0. Computing the boolean value can stop after it sees the first element, regardless of how long the sequence actually is.

PR #2559, opened 2022-10-25
Do not use len(SEQUENCE) to determine if a sequence is empty

DimitriPapadopoulos:
From DeepSource.io:
[Built-in function `len` used as condition PYL-C1802](https://deepsource.io/directory/analyzers/rust/issues/PYL-C1802)

> Using the `len` function to check if a sequence is empty is not idiomatic and can be less performant than checking the truthiness of the object.
> 
> `len` doesn't know the context in which it is called, so if computing the length means traversing the entire sequence, it must; it doesn't know that the result is just being compared to 0. Computing the boolean value can stop after it sees the first element, regardless of how long the sequence actually is.

Addresses https://github.com/codespell-project/codespell/pull/2552#discussion_r1004721881.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -782,7 +782,7 @@ def main(*args):
     options, parser, used_cfg_files = parse_options(args)
 
     # Report used config files
-    if len(used_cfg_files) > 0:
+    if used_cfg_files:
         print('Used config files:')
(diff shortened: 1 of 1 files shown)


cclauss:
https://pypi.org/project/refurb (and probably other tools) can auto-discover these as `FURB115`

DimitriPapadopoulos:
No problem, in that case do not hesitate to close this PR.

PR #2562, opened 2022-10-27
Redundant wheel dependency in pyproject.toml

DimitriPapadopoulos:
The `wheel` dependency is added by the backend automatically.

Listing it explicitly in the documentation was a historical mistake and has been fixed since:
	https://github.com/pypa/setuptools/commit/f7d30a9


larsoner:
Thanks @DimitriPapadopoulos

PR #2565, opened 2022-10-29
Update subprocess usage to use modern subprocess.run()

jdufresne:
https://docs.python.org/3/library/subprocess.html#subprocess.run

> The recommended approach to invoking subprocesses is to use the run()
> function for all use cases it can handle.

### DIFF
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -47,11 +47,10 @@ cs = MainWrapper()
 def run_codespell(args=(), cwd=None):
     """Run codespell."""
     args = ('--count',) + args
-    proc = subprocess.Popen(
+    proc = subprocess.run(
         ['codespell'] + list(args), cwd=cwd,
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
From the [`subprocess`](https://docs.python.org/3/library/subprocess.html) documentation:

> ### Using the [`subprocess`](https://docs.python.org/3/library/subprocess.html#module-subprocess) Module
> 
> The recommended approach to invoking subprocesses is to use the [`run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run) function for all use cases it can handle.

and more specifically, from the [subprocess.run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) documentation:

> _New in version 3.5._
> [...]
> _Changed in version 3.7:_ [...] Added the _capture_output_ parameter.

We support Python >= 3.7 so that's OK.

DimitriPapadopoulos (approved):

PR #2566, opened 2022-10-30
Add a few corrections

int-y1:
* `effcient->efficient` `suffcient->sufficient` and similar
* `hypothenuse->hypotenuse`
* `lexigraphic->lexicographic` and similar
* permute/permutate/permutation family

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -12717,6 +12717,8 @@ efel->evil
 eferences->references
 efetivity->effectivity
 effciency->efficiency
+effcient->efficient
+effciently->efficiently
 effctive->effective
 effctively->effectively
 effeciency->efficiency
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -24950,6 +24963,8 @@ permmissions->permissions
 permormance->performance
 permssion->permission
 permssions->permissions
+permuatation->permutation
+permuatations->permutations
```
DimitriPapadopoulos, line 24967: Add `permuatate` (with `permuatated`, `permuatates`, `permuatating`)?

DimitriPapadopoulos: Indeed. Then perhaps you should forget this `permuatate` suggestion.

As for `permutate->permute`, I would recommend against it because `permutate` is correct, it is found in the dictionary.

DimitriPapadopoulos: On the other hand, there are a few hots out there:
https://github.com/search?q=permuatate&type=code

int-y1: Done. I also added `permuate->permute, permutate,` `permution->permutation`.

DimitriPapadopoulos (approved):

PR #2568, opened 2022-10-30
Catch all cases of missing pytest

### LINKED ISSUE #2567: `make check` doesn't work on vanilla Ubuntu 22.04
⇒ Should we switch from `python` and `pytest` to `python3` and `pytest-3`?

-----

1. Command `python` is not available by default on a vanilla installation. Package [`python3-minimal`](https://packages.ubuntu.com/jammy/python3-minimall) provides `python3`, and package [`python-is-python3`](https://packages.ubuntu.com/jammy/python-is-python3), which is not installed by default, provides `python`:
  ```
  $ make check-distutils
  python setup.py check --restructuredtext --strict
  make: python: No such file or directory
  make: *** [Makefile:51: check-distutils] Error 127
  $ 
  ```
  We call `python` here:
  https://github.com/codespell-project/codespell/blob/458de804cde608110fc1ae91e4c39d92380eb0e6/Makefile#L50-L51

-----

2. Command `pytest` is not available, instead package [`python3-pytest`](https://packages.ubuntu.com/jammy/python3-pytest) provides `pytest-3`:
  ```
  $ make pytest
  pytest codespell_lib
  make: pytest: No such file or directory
  make: *** [Makefile:57: pytest] Error 127
  $ 
  ```
  We call `pytest` here:
  https://github.com/codespell-project/codespell/blob/458de804cde608110fc1ae91e4c39d92380eb0e6/Makefile#L56-L57

DimitriPapadopoulos:
Use `python3`/`pytest-3` instead of `python`/`pytest`.

Fixes #2567.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,12 @@ flake8:
 	flake8
 
 pytest:
-	pytest codespell_lib
+	@if command -v pytest > /dev/null; then \
+		pytest codespell_lib; \
+	else \
+		echo "Test dependencies not present, install using 'pip install -e \".[dev]\"'"; \
+		exit 1; \
+	fi
 
 clean:
 	rm -rf codespell.1


DimitriPapadopoulos:
The `python3` command should always be available according to [PEP 394](https://peps.python.org/pep-0394/), at least on Unix-Like Systems.

cclauss:
Is `python3` available on Windows?

DimitriPapadopoulos:
I think `pytest-3` is really specific to the Debian/Ubuntu [python3-pytest](https://packages.ubuntu.com/jammy/python3-pytest) package. Speaking of which, [`README.Debian`](https://salsa.debian.org/python-team/packages/pytest/-/blob/master/debian/README.Debian) is interesting:
> Since version 3.0.0 of pytest dropped the versioned entry points py.test-X.Y,
the recommended way to execute test suites using pytest is by running
> ```
> pythonX.Y -m pytest ...
> ```
> instead.

cclauss:
### macOS:
% `pytest-3 --version`
```
zsh: command not found: pytest-3
```
➜  `pytest --version`
```
pytest 7.1.2
```

DimitriPapadopoulos:
It looks like `python3` is not available on Windows, see for example:
[Bash: Python3: command not found (Windows, discord.py)](https://stackoverflow.com/questions/40914108/bash-python3-command-not-found-windows-discord-py)

But then, do we expect `make check` to run on a vanilla Windows machine? I mean, `make check` requires GNU Make, a Bourne shell, `grep`, `sed`, etc. Chances are, a `python3` symbolic link or something similar is available in an environment that provides these.

DimitriPapadopoulos:
I have kept `python3`, but changed `pytest` into `python3 -m pytest`.

DimitriPapadopoulos:
@cclauss Because `setup.py` is deprecated, we probably need to modify `python3 setup.py check`, don't we?

larsoner:
-1 on python3/pip3 command names. I think it's more a quirk of Debian linuxes. This will be unnecessary and wrong on conda for example. 

Also `pytest ...` is used very widely so I don't think we need to change it

DimitriPapadopoulos:
Then I willl just keep the test for `pytest` in all case.

DimitriPapadopoulos:
@larsoner Now catching all cases of missing `pytest`.

PR #2575, opened 2022-10-31
Added informal words

Paradact:
Added some informal words


skangas:
Makes sense to me, but I suggest squashing the commits into one.

peternewman (approved):
LGTM thanks

PR #2560, opened 2022-10-25
Document required setuptools version

DimitriPapadopoulos:
Building from source requires recent versions of some packages:
* `setuptools>=64`
* `setuptools_scm[tomli]>=6.2`

Addresses caveats discussed in #2523.

Fixes #2532
Fixes #2547

PR #2579, opened 2022-11-02
Fix typo in README pin → pip

DimitriPapadopoulos:
Introduced by myself in  #2560 / 479c5950f4221e3af623c81385c56d09a321173f.


peternewman (approved):
Maybe one day we can have a multi-word check for "pip install" :laughing:

PR #2585, opened 2022-11-06
Add testing and document support for Python 3.11

jdufresne:
Released Oct. 24, 2022.

https://www.python.org/downloads/release/python-3110/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,11 +21,6 @@ classifiers = [
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.7",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.10",
-    "Programming Language :: Python :: 3.11",
 ]
 dependencies = []
 dynamic = ["version"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,15 @@ classifiers = [
     "Operating System :: Microsoft :: Windows",
     "Operating System :: POSIX",
     "Operating System :: Unix",
-    "Operating System :: MacOS"
+    "Operating System :: MacOS",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
 ]
 dependencies = []
 dynamic = ["version"]
--- a/.github/workflows/codespell-private.yml
+++ b/.github/workflows/codespell-private.yml
@@ -21,6 +21,7 @@ jobs:
           - '3.8'
           - '3.9'
           - '3.10'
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -17,7 +17,15 @@ classifiers = [
...
-    "Operating System :: MacOS"
+    "Operating System :: MacOS",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.7",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
```
DimitriPapadopoulos, line 28: I would rather avoid the whole list in classifiers. It is already well documented here:
https://github.com/codespell-project/codespell/blob/6d001a1b32b53284278f297cce810e3fa453eafb/pyproject.toml#L7

This seems sufficient:
```yaml
    "Programming Language :: Python",
    "Programming Language :: Python :: 3",
    "Programming Language :: Python :: 3 :: Only",
```

jdufresne: Thanks for reviewing. I have made the suggested change in the latest commit.

As a consumer of third party packages, I find these classifiers useful. To me, they communicate that the package is verified and tested with all listed version. As well, it is easy for scripts and bots to consume this metadata to track trends across the Python community (e.g. https://hugovk.github.io/drop-python/3.6/). NBD though.

DimitriPapadopoulos: It is even easier for scripts and bots to consume `requires-python`, which is the authoritative source of information.

DimitriPapadopoulos (approved):

pyproject.toml
DimitriPapadopoulos: Ah, I get it: `requires-python` tells us that codespell is supposed to work with Python >= 3.7, while the classifiers convey the message that codespell has been actually tested with a specific Python version.

jdufresne: Another difference to consider, requires-python is a version specifier while the classifiers are on/off tags. The spec for how to interpret and write version specifiers isn't nothing and while a script could do it, I wouldn't describe it as easier.

https://peps.python.org/pep-0440/

DimitriPapadopoulos: Thinking about it again, how easy it is to parse classifiers is not relevant because the code is written once, then scripts and bots parse automatically. I believe the real problem is the duplication of information, and that the single source of information should be the one actually enforced by code. Now, I understand information is not entirely duplicated in this case – there are nuances. Plus, this duplication is customary in Python.

I don't have a strong opinion on the subject, I'd happily leave the decision to others. Actually, you could put the classifiers back, and we can ask the maintainers.

As an alternative, it might be easier to split in two PRs:
* Remove all classifiers from this PR, change the title to _Add testing for Python 3.11_, and I will merge it immediately.
* Create a new PR with title _Document tested versions of Python_ or something like that, and choose your arguments wisely.

larsoner:
For stuff like this I always just try to do what NumPy does because it's a very widely used project and tends to make reasonable decisions. Looks like this PR was along those lines before removing the minor-version classifiers:

https://github.com/numpy/numpy/blob/fcafb6560e37c948a594dce36d300888148bc599/setup.py#L92-L97

So I'll go ahead and merge and we can add the others in another PR

jdufresne:
> So I'll go ahead and merge and we can add the others in another PR

Thanks. Others are included in PR #2379

PR #2587, opened 2022-11-06
Use black to format Python files

jdufresne:
Black is a popular Python code formatter. By using it, contributors don't need to think about the project's styles preferences and avoids the need for hand-formatting. Instead, let the tool handle it. This can help smooth the review process.

Updated existing tools to follow the integration docs: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html

For more details, see the project:
https://github.com/psf/black


DimitriPapadopoulos:
I do like [blake](https://github.com/psf/black), except the enforcement of double quotes over single quotes – but if it is the price to pay for consistent Python style I am all for it.

If this commit is applied, it might be interesting to instruct `git blame` to ignore this commit in a `.git-blame-ignore-revs` file. See:
* [Ignoring mass reformatting commits with git blame](https://akrabat.com/ignoring-revisions-with-git-blame/)
* [Ignoring bulk change commits with git blame](https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame)

larsoner:
I'm fine with moving to `black`, it seems to be what I see most projects doing nowadays

PR #2588, opened 2022-11-06
Add type annotations to the project and use mypy

jdufresne:
Provide type annotations and use mypy for static type checking. Type checkers help ensure that the project is using variables and functions in the code correctly. With mypy, CI will warn when those types are used incorrectly.

The mypy project and docs:
https://github.com/python/mypy
https://mypy.readthedocs.io/en/stable/index.html


larsoner (approved):

larsoner:
Thanks @jdufresne

PR #2595, opened 2022-11-16
PEP 517

### LINKED ISSUE #2592: 2.2.2: codespell script installed in wrong location
During package 2.2.2 I found that `codespell` script is not installed in bindir.
```
Provides: codespell = 2.2.2-3.fc35 python3.8dist(codespell) = 2.2.2 python3dist(codespell) = 2.2.2
Requires(rpmlib): rpmlib(CompressedFileNames) <= 3.0.4-1 rpmlib(FileDigests) <= 4.6.0-1 rpmlib(PayloadFilesHavePrefix) <= 4.0-1
Requires: /usr/bin/python3 python(abi) = 3.8
Checking for unpackaged file(s): /usr/lib/rpm/check-files /home/tkloczko/rpmbuild/BUILDROOT/codespell-2.2.2-3.fc35.x86_64
error: Installed (but unpackaged) file(s) found:
   /usr/lib/python3.8/site-packages/bin/codespell
```

DimitriPapadopoulos:
No need to provide `bin/codespell`, it will be generated by this entry point specification:
https://github.com/codespell-project/codespell/blob/1302d7974fb12ae2d723376e81e9b2f1e5c7a5bd/pyproject.toml#L50-L51

For the reviewers: not sure at this point if we need to adapt this:
https://github.com/codespell-project/codespell/blob/1302d7974fb12ae2d723376e81e9b2f1e5c7a5bd/Makefile#L15-L17

Fixes #2592.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,7 +3,6 @@ recursive-include codespell_lib *.py
 include codespell_lib/data/dictionary*.txt
 include codespell_lib/data/linux-kernel.exclude
 include COPYING
-include bin/codespell
 exclude *.yml *.yaml
 exclude .coveragerc
 exclude .git-blame-ignore-revs
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,8 @@ check-dictionary: check-dictionaries
 sort-dictionary: sort-dictionaries
 trim-dictionary: trim-dictionaries
 
-codespell.1: codespell.1.include bin/codespell Makefile
-	PYTHONPATH=. help2man ./bin/codespell --include codespell.1.include --no-info --output codespell.1
+codespell.1: codespell.1.include Makefile
+	PYTHONPATH=. help2man codespell --include codespell.1.include --no-info --output codespell.1
 	sed -i '/\.SS \"Usage/,+2d' codespell.1
 
 check-dictionaries:
--- a/bin/codespell
+++ b/bin/codespell
@@ -1,7 +1,0 @@
-#!/usr/bin/env python3
-
-import sys
-
-if __name__ == '__main__':
-    import codespell_lib
-    sys.exit(codespell_lib.main(*sys.argv[1:]))


DimitriPapadopoulos:
See:
* [Entry points specification](https://packaging.python.org/specifications/entry-points/#use-for-scripts)
* [Declaring project metadata](https://packaging.python.org/specifications/declaring-project-metadata/#entry-points)

peternewman:
> For the reviewers: not sure at this point if we need to adapt this:
> 
> https://github.com/codespell-project/codespell/blob/1302d7974fb12ae2d723376e81e9b2f1e5c7a5bd/Makefile#L15-L17

I would imagine so, because bin/codespell won't exist on a checkout, so there's no way to generate it, which will confuse make, and also no way to run the command to generate the man file, which will break that.

DimitriPapadopoulos:
Unfortunately, I have no clue how to get `help2man` to generate the doc without an explicit `bin/codespell` script, or how to temporarily build the `bin/codespell` script without a full-fledged build.

larsoner:
> I would imagine so, because bin/codespell won't exist on a checkout, so there's no way to generate it, which will confuse make

The standard workflow is to do something like `pip install -e .` or `python setup.py develop`, both of which should make `codespell` available (wherever standard binary installations put their scripts). For me, because I use system python with `--user`, I get:
```
$ which codespell
/home/larsoner/.local/bin/codespell
```
So +1 for removing `bin/codespell` from me

larsoner:
(and assuming in `make` that just running `codespell ...` will hit the correct codespell, i.e., the dev has set up their testing env properly with an editable installation)

DimitriPapadopoulos:
@larsoner Does this mean that I keep the `codespell.1` target in the `Makefile` and accept the fact that it will fail if the developer has not set up the environment properly? If so, is it OK to explicitly test the existence of `bin/codespell` in the target rule, and print a more user-friendly error message?

larsoner:
I would get rid of any reference to `./bin/codespell` and just use `codespell` directly

DimitriPapadopoulos:
We do not run `codespell`, we run `help2man`:
 ```
help2man ./bin/codespell
```
Do you mean I should change to the following?
```
help2man `which codespell`
```

**Comment by larsoner** [comment]
Did you try it directly? For me this is appears to be enough
```
$ help2man codespell
.\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.49.2.
.TH CODESPELL "1" "November 2022" "codespell 2.2.3.dev23+ga050986b" "User Commands"
.SH NAME
codespell \- manual page for codespell 2.2.3.dev23+ga050986b
.SH DESCRIPTION
usage: codespell [\-h] [\-\-version] [\-d] [\-c] [\-w] [\-D DICTIONARY]
.IP
...
```
but yes just use whatever works...

**Comment by DimitriPapadopoulos** [comment]
@larsoner Works for me, thank you for the tip about `help2man`:
```
$ make codespell.1
PYTHONPATH=. help2man codespell --include codespell.1.include --no-info --output codespell.1
sed -i '/\.SS \"Usage/,+2d' codespell.1
$ 
```

jdufresne (approved):

larsoner:
Thanks @DimitriPapadopoulos !

PR #2596, opened 2022-11-16
Remove extra point

DimitriPapadopoulos:


peternewman (changes requested):
We seem to be inconsistent throughout the file with:
`word. ::`
`word.::`
`word::`

It would be good to make them match...

DimitriPapadopoulos:
Indeed, I have noticed the discrepancy. However, I think we do need these differences, at least in some cases:
* [word. ::](url) → word.
* [word::](url) → word:

Only this one is clearly wrong, and I think I have fixed its only occurrence:
* [word.::](url) → word.:

PR #2600, opened 2022-11-17
Move sudo-like command name "doas" to code dictionary

skangas:
Please consider not correcting "doas", as that is the command name used for the OpenBSD `sudo` replacement.

Source: https://man.openbsd.org/doas

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -14,6 +14,7 @@ copyable->copiable
 creat->create, crate,
 define'd->defined
 deque->dequeue
+doas->does, do as,
 dof->of, doff,
 dont->don't
 dosclose->disclose
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -12156,7 +12156,6 @@ doalog->dialog
 doamin->domain, dopamine,
 doamine->dopamine, domain,
 doamins->domains
-doas->does
 doasn't->doesn't
 doble->double
 dobled->doubled


peternewman (changes requested):
Please can you move this to the code dictionary rather than just removing it, as it's relevant in e.g. a readme. Can you also add "do as" as a suggestion.

skangas:
For what it's worth, I have several places where we use "doas" in source code too, for example:

https://github.com/emacs-mirror/emacs/blob/f320663239a14aceee01868c465a4461e3a69954/lisp/eshell/em-tramp.el#L133

But I'm happy to make the change you suggest, if you prefer. I can add the word to our project local whitelist.

peternewman:
> For what it's worth, I have several places where we use "doas" in source code too

Yes, it will skip those too:
https://github.com/codespell-project/codespell/blob/8d0d82bcc07aaef5211127683aba7de67b923aac/codespell_lib/_codespell.py#L62

skangas:
Thanks, I misunderstood the purpose of the code dictionary. I have pushed an updated patch.

peternewman:
> Thanks, I misunderstood the purpose of the code dictionary.

Yeah I think everyone does, we've tried to tweak the wording a few times without success.

> I have pushed an updated patch.

Great

codespell_lib/data/dictionary_code.txt
```diff
@@ -14,6 +14,7 @@ copyable->copiable
 creat->create, crate,
 define'd->defined
 deque->dequeue
+doas->does
```
peternewman, line 17:
```suggestion
doas->does, do as,
```

peternewman (changes requested):
Add another suggestion

codespell_lib/data/dictionary_code.txt
skangas: Pushed a new commit with that change, thanks. Sorry I missed it in your previous comment.

peternewman: No worries, I couldn't do a proper code suggestion on the deleted line.

peternewman (approved):
LGTM thanks.

PR #2601, opened 2022-11-17
Applied lower case before splitting string

bdsoha:


peternewman (commented):
Hi @bdsoha ,

Why? There's a minor speed-up presumably, but it's taking us away from, not towards supporting fixing caps which seems like the wrong way to go to me?

bdsoha:
@peternewman I was focusing on code complexity, but you are correct.

Issue #2281, opened 2022-02-17
`crate` is super common in Rust contexts


mrjones2014:
Currently the word "crate" is recognized as a typo for "create".

In the context of Rust projects, you might have the word "crate" all over your README.md.

"crate" should probably be added to the dictionary.

peternewman:
Hi @mrjones2014 ,

Please see the previous discussion in #1660 #1902 .

Essentially the issue is that `crate->create` is likely to be of use for everyone apart from Rust users, who can skip that one word. Also unlike everything else we've got, it falls across two groups, as it's a genuine potential typo in day to day use, as well as being a code based typo for one language.

carlosflorencio:
Can't we check the file extension and if it's `.rs`, ignore certain words like `crate`? This would allow having skip lists per filetype.

Sorry if the proposal is too naive, I don't know how codespell works underneath.

DimitriPapadopoulos:
codespell is currently language-agnostic or said otherwise file-type-agnostic. We'd have to change that first.

See https://github.com/codespell-project/codespell/issues/17#issuecomment-1446487445.

DimitriPapadopoulos:
In any case, `crate` has been moved to `dictionary_rare.txt`, but _rare_ is still part of the default `'clear,rare'` dictionaries.

Perhaps we should move it to `dictionary_code.txt`.

ismet55555:
The `codespell: crate ==> create` is the most annoying thing when programming rust. Mix that with some of the `nvim` configurations and trying to add some exempt word to it. It sucks.

DimitriPapadopoulos:
@ismet55555 Don't hesitate to submit a PR to move `crate` to the _code_ dictionary.

ismet55555:
I mean I think it can be fixed by specifying arguments for codespell. However, nvim configuration frameworks like AstroNvim, to get Mason to install it and reference a codespell config file or pass cli arguments is not the most straightforward (at least for me). So for the time being I’m using cspell

PR #2602, opened 2022-11-17
Don't correct "crate" since it is used by Rust

skangas:
Please consider not correcting "crate"->"create", as Rust uses crates.

Source: https://crates.io/


peternewman:
I'm closing this as it's a duplicate of #2281 see some of the discussion in there for why.

PR #2605, opened 2022-11-17
Move "acount" to code dictionary

skangas:
I have code using `acount` as a variable name. Please consider moving that correction to the code dictionary. Thanks.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -665,7 +665,6 @@ acordingly->accordingly
 acordinng->according
 acorss->across
 acorting->according
-acount->account
 acounts->accounts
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
Other programmers too may use variable names starting with `a`, `the` or `my`. It's not necessarily a reason to remove such possible typos from the dictionary. Instead, you may use option `-L`.

peternewman:
Yeah as @DimitriPapadopoulos said, this exception should only be for words that aren't actually typos and are used in system functions or popular languages, or I guess popular libraries for a popular language.

Hence I'll close this now.

skangas:
Makes sense, thanks.

PR #2611, opened 2022-11-19
Add fixes from misspell-fixer's safe.1.dict

skangas:
This adds some missing corrections found in the dictionary used by [misspell-fixer](https://github.com/vlajos/misspell-fixer/). I have also added in variations, and curated the list of corrections slightly (removed dubious fixes of alternative spellings and similar).

This is part 1 of 2, covering the `safe.*.dict` files. Once both pull requests are merged, I believe that we will have all relevant fixes from that project in codespell. (The other one is #2612.)

Source: https://github.com/vlajos/misspell-fixer/blob/master/dict/safe.1.dict

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -2918,6 +2918,7 @@ artillary->artillery
 artuments->arguments
 arugment->argument
 arugments->arguments
+arument->argument
 aruments->arguments
 arund->around
 arvg->argv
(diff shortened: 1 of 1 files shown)


skangas:
Turns out that only two pull requests were needed, so this is the last one.

DimitriPapadopoulos (approved):

skangas:
There were merge conflicts, so I have pushed a rebased version.

I also see that this was marked as "approved" but not yet merged? Please let me know if there's anything more that needs fixing before this can go in. Thanks in advance.

PR #2614, opened 2022-11-19
Move "spawnve" to code dictionary

skangas:
On MS-Windows, apparently `spawnve` or `_spawnve` is a commonly used C function, so I suggest we move it to the code dictionary.

Source: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/spawnve?view=msvc-170


peternewman (approved):
LGTM

Issue #2375, opened 2022-06-23
Bug?  `ignore-words-list` seems to have no effect


joshgoebel:
Am I doing something wrong?  How can I get it to stop flagging `MSDOS` as an error?  `ignore-words-list` seems to have no effect

```
codespell
./src/keyszer/models/key.py:161: MSDOS ==> MS-DOS
```

My config:

```
[codespell]
ignore-words-list = MSDOS
skip = .venv,.git,build,keyszer.egg-info
```

peternewman:
Like so:
> ```ini
> [codespell]
> ignore-words-list = msdos
> skip = .venv,.git,build,keyszer.egg-info
> ```

Which admittedly doesn't match our docs:
https://github.com/codespell-project/codespell/#usage

Based on:
https://github.com/codespell-project/codespell/blob/978a22315238749224c5659b95f5fab145252a0f/codespell_lib/data/dictionary.txt#L20494

joshgoebel:
That's it - fixes the issue, but yeah the way i read the docs was that I should try to match case...

peternewman:
https://github.com/codespell-project/codespell/blob/1d9dee38735a173b4827b67a4f13987290a03475/codespell_lib/_codespell.py#L456-L462

I think the issue may be that MSDOS is one of the few entries which is upper case in the dictionary, hence flagging this issue.

mwestphal:
Looks like there is a case issue:

```
nNumber ==> number
```

```
ignore-words-list=nNumber
```

Not excluded

```
ignore-words-list=nnumber
```

excluded successfully.

peternewman:
> Looks like there is a case issue:

Yeah our comparison to match the dictionary from the ignore list is happening with different case requirements to what we claim in the docs.

yarikoptic:
FWIW just ran into this issue and was about to complain as well -- would be great to have it fixed, as to conform to the promise of being case sensitive. My use case:

```
❯ codespell --help | grep -A3 -e ' -L'
  -L WORDS, --ignore-words-list WORDS
                        comma separated list of words to be ignored by
                        codespell. Words are case sensitive based on how they
                        are written in the dictionary file
❯ cat spellme.txt
Ingress for Microsoft Azure Kubernetes Service (AKS)
❯ codespell -L AKS spellme.txt
spellme.txt:1: AKS ==> ASK
❯ codespell --version
2.2.2
```

so I would dislike completely avoiding fixing `aks` into `ask` ;)

[...]

PR #2615, opened 2022-11-20
Move "MSDOS" to code dictionary

skangas:
I'm getting a _ton_ of false positives for the MSDOS correction, such as:

```
-#ifdef MSDOS
-#include "msdos.h"
+#ifdef MS-DOS
+#include "ms-dos.h"
```

I suggest moving it to the code dictionary. Thanks.


skangas:
See also #2375.

peternewman (approved):
LGTM thanks.

PR #2617, opened 2022-11-20
Skip comments and blanks in dictionary files

skangas:
This allows using comments and blank lines in dictionary files. My use case is that I have a file like the following:

```
# submit upstream
sceanrios->scenarios
eamcs->emacs
sccopes->scopes

# emacs specific
brunches->branches
ecmacs->emacs
ehsell->eshell
[...]
```


skangas:
The code coverage test seems to be failing. I'll take a closer look if this feature is considered acceptable.

PR #2535, opened 2022-10-17
MAINT: Add CODEOWNERS

larsoner:
To facilitate review, we can have a CODEOWNERS file that allows maintainers to be auto-tagged for review. I've added myself here. I propose we merge this, then @luzpaz @peternewman you can open PRs to add yourself wherever you want.

PR #2618, opened 2022-11-23
Don't print config files by default

skangas:
I find the default output to be to verbose when using the `--config` flag. There is no need for me to see that configuration file printed out on every invocation. I suggest hiding that behind a new `--debug` flag.

Thanks.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -132,6 +132,7 @@ class QuietLevels:
     DISABLED_FIXES = 4
     NON_AUTOMATIC_FIXES = 8
     FIXES = 16
+    CONFIG_FILES = 32
 
 
 class GlobMatch:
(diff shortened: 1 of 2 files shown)


skangas:
Here too, code coverage seems to be complaining. I'll look into it if this feature is accepted. Thanks.

DimitriPapadopoulos:
Printing the config file is a very recent addition, see #2534 and #2535.

larsoner:
I think I'd rather add this to `-q` as 32, so the default would now be 34 (32+2)

skangas:
> I think I'd rather add this to `-q` as 32, so the default would now be 34 (32+2)

I personally don't have much of an opinion, as long as I can silence the message. If it's silent by the default, then all the better from my point of view.

If your proposal is the preferred one, I'm happy to write up a new patch along those lines.

Thanks.

skangas:
> I think I'd rather add this to `-q` as 32, so the default would now be 34 (32+2)

I've implemented this and updated the pull request. Thanks.

larsoner (approved):

larsoner:
Thanks @skangas !

PR #2620, opened 2022-11-25
Most of `MANIFEST.in` is obsolete, get rid of it

DimitriPapadopoulos:
Fom [pypa /setuptools_scm](https://github.com/pypa/setuptools_scm):

> Additionally `setuptools_scm` provides setuptools with a list of files that are managed by the SCM (i.e. it automatically adds all of the SCM-managed files to the sdist). Unwanted files must be excluded by discarding them via `MANIFEST.in`.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,12 +1,7 @@
-include codespell_lib/__init__.py
-recursive-include codespell_lib *.py
-include codespell_lib/data/dictionary*.txt
-include codespell_lib/data/linux-kernel.exclude
-include COPYING
 exclude *.yml *.yaml
 exclude .coveragerc
 exclude .git-blame-ignore-revs
-exclude example example/* snap snap/* tools tools/*
+exclude example example/* snap snap/*
 exclude Makefile
 exclude codespell.1.include
 exclude pyproject-codespell.precommit-toml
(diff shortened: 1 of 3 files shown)


jdufresne (approved):

DimitriPapadopoulos:
As far as I can see, `check-manifest` is not `setuptools_scm`-aware, it insists files such as `COPYING` are added to `MANIFEST.in`:
```
$ check-manifest --no-build-isolation
lists of files in version control and sdist do not match!
missing from VCS:
  codespell_lib/_version.py
missing from sdist:
  COPYING
suggested MANIFEST.in rules:
  include COPYING
$ 
```

larsoner:
If there are things to ignore, we should be able to add them to the `--ignore` I think, e.g.,

https://github.com/mne-tools/mne-python/blob/8f01c9002a8ec6d89824e0cc1c6415be73ab1987/Makefile#L120

DimitriPapadopoulos:
I'm not sure what `check-manifest` is supposed to check now that all files in the VCS are implicitly and automatically included. What is left to check?

DimitriPapadopoulos:
By the way, the failing tests seem unrelated to this merge request. They appear to be caused by #2626, which runs `python -m build` without adding `build` as depedency to pyproject.toml.

PR #2625, opened 2022-11-27
Switch from pytest tmpdir to tmp_path in tests

jdufresne:
The tmp_path fixture was introduced in pytest 3.9.0 (2018-10-15). It generates a Python pathlib.Path object. The docs mention the following about tmpdir:

https://docs.pytest.org/en/7.2.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

> These days, it is preferred to use tmp_path and tmp_path_factory.

This change helps make the test suite a bit more future proof for when the pytest project eventually deprecates the old fixtures.

The Path object provides several nice convenient methods that simplifies and flattens the code a bit.


DimitriPapadopoulos:
Just a suggestion: perhaps just change `tmpdir` → `tmp_path` in a first commit, then simplify/flatten the code in a different commit? Would be easier to understand, at least for me.

In any case, this needs to be rebased.

larsoner (approved):
Looks like a nice improvement for me, pushed a commit to fix the conflict, marking for merge when green!

pcaversaccio:
hey, I'm not sure whether this PR introduced a regression, but my codespell actions worked like charm around 30mins ago and now are failing with the following error:

```bash
  File "/usr/local/bin/codespell", line 8, in <module>
    sys.exit(_script_main())
  File "/usr/local/lib/python3.8/site-packages/codespell_lib/_codespell.py", line 987, in _script_main
    return main(*sys.argv[1:])
  File "/usr/local/lib/python3.8/site-packages/codespell_lib/_codespell.py", line 992, in main
    options, parser, used_cfg_files = parse_options(args)
  File "/usr/local/lib/python3.8/site-packages/codespell_lib/_codespell.py", line 568, in parse_options
    data = tomllib.load(f).get("tool", {})
UnboundLocalError: local variable 'tomllib' referenced before assignment
```

Here an example run: https://github.com/pcaversaccio/snekmate/actions/runs/4367179013/jobs/7638378970

Running locally the `codespell` check returns successfully so over the last hour or so something was introduced here (or in the GH actions build itself) that breaks it.

PR #2626, opened 2022-11-28
Remove setup.py

### LINKED ISSUE #2606: `python setup.py check` → `twine check`
Because `setup.py ...` is deprecated, we need an alternative to `setup.py check` such as `twine`.

jdufresne:
With pyproject.toml, setup.py becomes unnecessary. Future versions of setuptools (and Python packaging in general) will deprecate the file.

The command `setup.py check` was replaced by `twine check` which serves a similar role.

setuptools is still a dependency of the build, just doesn't require setup.py.

Fixes #2606

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,11 @@ SORT_ARGS := -f -b
 
 DICTIONARIES := codespell_lib/data/dictionary*.txt
 
-PHONY := all check check-dictionaries sort-dictionaries trim-dictionaries check-dictionary sort-dictionary trim-dictionary check-manifest check-distutils flake8 pytest pypi clean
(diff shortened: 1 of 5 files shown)


larsoner:
Including this was intentional, following the `Important` at the top here

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

But we can try this and see if it breaks stuff...

Makefile
```diff
@@ -47,8 +47,11 @@ trim-dictionaries:
 check-manifest:
 	check-manifest --no-build-isolation
 
-check-distutils:
-	python setup.py check --restructuredtext --strict
+check-dist:
+	$(eval TMP := $(shell mktemp -d))
+	python -m build -o $(TMP)
```
DimitriPapadopoulos, line 52: This requires [build](https://github.com/pypa/build), but build has not been added as a dependency in `pyproject.toml`.

PR #2639, opened 2022-12-12
README: Document how to use pre-commit

WilliamJamieson:
I came across codespell recently and decided to try to introduce it to my project (asdf-format/asdf#1267). While doing so I also wanted to setup codespell to work with the pre-commit, as this is used extensively by that project. However, I found that the documentation of how to use the pre-commit hook very lacking (in particular getting the hook to properly work with a `pyproject.toml` configuration). This PR attempts to better document how to set codespell up with pre-commit.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,6 +154,29 @@ config files.
 
 .. _tomli: https://pypi.org/project/tomli/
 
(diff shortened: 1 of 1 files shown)


larsoner (approved):
Thanks @WilliamJamieson !

PR #2642, opened 2022-12-16
Flush stdout when asking word fix

adrien-berchet:
In some cases the prompt for word fix does not appears (e.g. when ran in `tox >= 4`). This PR solves this issue.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -693,8 +693,7 @@ def ask_for_word_fix(
         r = ""
         fixword = fix_case(wrongword, misspelling.data)
         while not r:
-            print(f"{line}\t{wrongword} ==> {fixword} (Y/n) ", end="",
-                  flush=True)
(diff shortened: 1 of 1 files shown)


codespell_lib/_codespell.py
```diff
@@ -694,6 +694,7 @@ def ask_for_word_fix(
         fixword = fix_case(wrongword, misspelling.data)
         while not r:
             print(f"{line}\t{wrongword} ==> {fixword} (Y/n) ", end="")
+            sys.stdout.flush()
```
larsoner, line 697: More compact
```suggestion
            print(f"{line}\t{wrongword} ==> {fixword} (Y/n) ", end="", flush=True)
```

adrien-berchet: Even better indeed, thanks!
I guess we could also use it to replace the line 720?

DimitriPapadopoulos (approved):

codespell_lib/_codespell.py
larsoner: Sure, feel free

PR #2646, opened 2022-12-22
 Most of `MANIFEST.in` is obsolete, get rid of it

DimitriPapadopoulos:
`MANIFEST.in` is useful to **exclude** files, but not to **include** files any more according to [pypa /setuptools_scm](https://github.com/pypa/setuptools_scm):
> Additionally `setuptools_scm` provides setuptools with a list of files that are managed by the SCM (i.e. it automatically adds all of the SCM-managed files to the sdist). Unwanted files must be excluded by discarding them via `MANIFEST.in`.

As far as I can see, `check-manifest` is not `setuptools_scm`-aware and obsolete, it insists files such as `COPYING` are added to `MANIFEST.in`:
```
$ check-manifest --no-build-isolation
lists of files in version control and sdist do not match!
missing from VCS:
  codespell_lib/_version.py
missing from sdist:
  COPYING
suggested MANIFEST.in rules:
  include COPYING
$ 
```

Finally, add `build` to the dev dependencies to fix issue caused by #2626, which added `python -m build` without adding build as dependency to `pyproject.toml`.

Re-opens #2620 after changing branch name.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,12 +1,7 @@
-include codespell_lib/__init__.py
-recursive-include codespell_lib *.py
-include codespell_lib/data/dictionary*.txt
-include codespell_lib/data/linux-kernel.exclude
-include COPYING
 exclude *.yml *.yaml
 exclude .coveragerc
 exclude .git-blame-ignore-revs
-exclude example example/* snap snap/* tools tools/*
+exclude example example/* snap snap/*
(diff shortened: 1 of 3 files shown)


jdufresne (approved):

larsoner:
Thanks @DimitriPapadopoulos

PR #2654, opened 2022-12-24
Add some typos found in FreedroidRPG

skangas:
Just some typos found in FreedroidRPG.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -8584,7 +8584,12 @@ conveinent->convenient
 conveinience->convenience
 conveinient->convenient
 convenant->covenant
+convencion->convention
+convencional->conventional
+convencionally->conventionally
 convenction->convention, convection,
+convenctional->conventional
+convenctionally->conventionally
 conveneince->convenience
 conveniance->convenience
 conveniant->convenient
@@ -9138,6 +9143,8 @@ countain->contain
 countainer->container
 countainers->containers
 countains->contains
+countere->counter
+counteres->counters
 counterfit->counterfeit
 counterfits->counterfeits
 counterintuive->counter intuitive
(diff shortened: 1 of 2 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -34544,6 +34553,7 @@ teusday->Tuesday
 texchnically->technically
 texline->textline
 textfrme->textframe
+texturers->textures
```
DimitriPapadopoulos, line 34556: While in [FreedroidRPG](https://gitlab.com/freedroid/freedroid-src/-/blob/master/src/text.c#L27-L32) the "_OpenGL texturers_” really ought to be "_OpenGL texturers_", "_texturers_" might refer to specialised workers applying textures in rendering. Seen in some ads, for example:
* [Volunteer modellers, texturers and animators needed for ambitious FNAF project](https://blenderartists.org/t/volunteer-modellers-texturers-and-animators-needed-for-ambitious-fnaf-project/1379788)

Very few hits in Google and GitHub. Again, most refer to specialists of textures.

That's really a niche. What do you think? Keep this as is or move `texturer`/`texturers` to rare?

skangas: > While in [FreedroidRPG](https://gitlab.com/freedroid/freedroid-src/-/blob/master/src/text.c#L27-L32) the "_OpenGL texturers_” really ought to be "_OpenGL textures_", "_texturers_" might refer to specialised workers applying textures in rendering. [...] What do you think? Keep this as is or move `texturer`/`texturers` to rare?

Hmm, given that `--help` says to use,

```
                        - 'rare' for rare (but valid) words that are likely to
                          be errors
```

and I think "texturers" is _not_ an "unambiguous" error, it should indeed go to rare. I also took a look at the first 5 pages of the 160 or so hits on GitHub, and they were all false positives: "texturers" was correct in those places.

So I think it makes sense to move it to rare (or just delete it, I guess, but it did help me find that one typo in FreedroidRPG). I've moved it for now.

DimitriPapadopoulos (approved):

PR #2655, opened 2022-12-24
Harden GitHub CI configuration

skangas:
This reduces the attack surface for GitHub actions by restricting the defaults. I'm not sure how practical or important these attacks are currently, but it seems like there's no reason not to do this for the tests, linters and spell checkers that are currently used.

See here for more:
https://docs.github.com/en/actions/security-guides/
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs
https://github.com/actions/checkout#checkout-v3

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -3,10 +3,13 @@ name: black
 on:
   - push
   - pull_request
+permissions: {}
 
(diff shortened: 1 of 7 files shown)


larsoner:
Thanks @skangas

PR #1953, opened 2021-06-14
Typos from marvin.cs.uidaho.edu/misspell.html (rework of #1374): D

bl-ue:
For #1949

PR #2663, opened 2022-12-26
Typos from marvin.cs.uidaho.edu: D

skangas:
This replaces PR #1953. I took care of the comments from the code review, and also added in several variations myself.

See: http://marvin.cs.uidaho.edu/misspell.html

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -9984,6 +9984,9 @@ cyrto->crypto
 cywgin->Cygwin
 daa->data
 dabase->database
+dabree->debris
+dabue->debut
+dackery->daiquiri
 daclaration->declaration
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -10277,6 +10286,10 @@ decendentant->descendant
 decendentants->descendants
 decendents->descendents, descendants,
 decending->descending
+decern->discern
+decerned->discerned
+decerning->discerning
+decerns->discerns
```
DimitriPapadopoulos, line 10292: I would expect French speakers to use "decern" instead of "award". Just adding this for future reference: no need to change anything, because that's not a spelling issue as such.

DimitriPapadopoulos (approved):

PR #1954, opened 2021-06-14
Typos from marvin.cs.uidaho.edu/misspell.html (rework of #1374): E

bl-ue:
For #1949

PR #2664, opened 2022-12-26
Typos from marvin.cs.uidaho.edu: E

skangas:
This replaces PR #1954. I took care of the comments from the code review, and also added in several variations myself.

See: http://marvin.cs.uidaho.edu/misspell.html

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -14291,9 +14291,7 @@ emplying->employing
 emplyment->employment
 emplyments->employments
 emporer->emperor
-empress->impress
 empressed->impressed
-empresses->impresses
 empressing->impressing
 empressive->impressive
 empressively->impressively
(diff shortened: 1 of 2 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -13936,7 +13996,9 @@ entaglements->entanglements
 entended->intended
 entension->extension
 entensions->extensions
+entent->intent
```
DimitriPapadopoulos, line 13999: Also `entending->intending`?

DimitriPapadopoulos: Wait, `entend` is an obsolete form of `intend`!

In any case, either have both or remove both, or move to _rare_:
```
entended->intended
entending->intending
```

skangas: I can't find them in any modern dictionary, so I believe putting them in `dictionary.txt` should be fine. For what it's worth, I don't recall ever having seen that spelling before either.

DimitriPapadopoulos (approved):

PR #2700, opened 2023-01-04
Remove alias make targets to facilitate tab completion

jdufresne:
When running the check/sort/trim make targets, I normally type something along the lines of:

    make check-d<TAB>

And let the shell tab completion finish the line for me. As there exists multiple variations, (check-dictionary and check-dictionaries) tab completion stops at the common substring and then I must finish typing the rest of the command.

By removing the aliases, tab completion will complete the full command immediately.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,15 +2,11 @@ SORT_ARGS := -f -b
 
 DICTIONARIES := codespell_lib/data/dictionary*.txt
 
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
I have the same problem, but hadn't noticed these were actually aliases until now – I thought `sort-dictionary` was for `dictionary.txt` only and `sort-dictionaries` for all dictionary files!

I am all for it :+1:

DimitriPapadopoulos (approved):

PR #2702, opened 2023-01-04
Use type annotations directly from chardet package

jdufresne:
Starting with chardet 5.1.0, the package now includes type annotations. The typeshed stubs are obsolete and should no longer be used.

See the release notes:
https://github.com/chardet/chardet/releases/tag/5.1.0

> Add type annotations to the project and run mypy on CI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,11 +48,11 @@ toml = [
     "tomli; python_version < '3.11'"
 ]
 types = [
+    "chardet>=5.1.0",
     "mypy",
     "pytest",
     "pytest-cov",
     "pytest-dependency",
-    "types-chardet",
 ]
 
 [project.scripts]


DimitriPapadopoulos:
Is there a way to have `types-chardet` if `chardet <5.1.0`? It might be less disruptive for existing checkouts.

On the other hand, that's for type checking and I would expect the concerned public to be able to sort out related issues.

jdufresne:
> Is there a way to have types-chardet if chardet <5.1.0? It might be less disruptive for existing checkouts.

Not that I'm aware of. AFAICT, Python dependencies don't have a concept of "either this package OR that package".

> On the other hand, that's for type checking and I would expect those concerned to be able to sort out related issues.

I agree. This only affects development environments and not end users.

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @jdufresne.

PR #2708, opened 2023-01-12
move `pullrequest` to code dictionary

robin-wayve:
moving `pullrequest` et al to the code dictionary after seeing a false positive in a GitHub GraphQL query: https://docs.github.com/en/graphql/reference/objects#pullrequest

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -29528,9 +29528,6 @@ Pucini->Puccini
 Puertorrican->Puerto Rican
 Puertorricans->Puerto Ricans
 pulisher->publisher
-pullrequenst->pull requests, pull request,
-pullrequest->pull request
-pullrequests->pull requests
 puls->pulse, plus,
 pumkin->pumpkin
 punctation->punctuation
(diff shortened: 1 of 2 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -29528,9 +29528,6 @@ Pucini->Puccini
 Puertorrican->Puerto Rican
 Puertorricans->Puerto Ricans
 pulisher->publisher
-pullrequenst->pull requests, pull request,
```
DimitriPapadopoulos, line 29531: Moving  `pullrequest` is OK, but not `pullrequenst`.

DimitriPapadopoulos (changes requested):

codespell_lib/data/dictionary.txt
robin-wayve: if I misspelled `pullRequest` as `pullRequenst` in a GraphQL query, I wouldn't expect `pull Request` as the correction though 🤔 

it also looks like there isn't much evidence for `pullrequenst` ever occurring in the wild https://github.com/search?q=pullrequenst&type=code -- so maybe I could just drop it entirely?

DimitriPapadopoulos: True, let's keep it then.

If it has been added, that's probably because it has seen somewhere in the wild.

DimitriPapadopoulos: I was able to find a few "pull requenst" occurrences using Google.

DimitriPapadopoulos (approved):

PR #2715, opened 2023-01-17
Improve support of TOML config files

### LINKED ISSUE #2713: codespell toml parsing with python 3.11
Hi,

this seems not to work currently.

the pyproject.toml

```yaml
toml = [
    "tomli; python_version < '3.11'"
]
```

but without tomly nothing will be parsed, because nowhere the python 3.11 stdlib tomlib gets imported and used.

DimitriPapadopoulos:
* Import tomlib with Python >= 3.11, and ensure compatibility as suggested here:
   https://github.com/hukkin/tomli#building-a-tomlitomllib-compatibility-layer
* Import tomlib/tomli only once, instead of re-importing for each TOML config file.
* If tomli cannot be imported, raise exception only if a TOML file is explicitly specified with option --toml. Do not raise an exception if pyproject.toml exists but is not explicitly specified.

Fixes #2713.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -544,25 +544,29 @@ def parse_options(
     config = configparser.ConfigParser(interpolation=None)
 
     # Read toml before other config files.
-    toml_files_errors = []
+    toml_files = []
+    tomlib_raise_error = False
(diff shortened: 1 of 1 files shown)


larsoner:
Thanks @DimitriPapadopoulos !

PR #2721, opened 2023-01-31
Add evaluatated->evaluated

arm-in:


DimitriPapadopoulos (commented):
We usually add all verbal forms, provided the typos do appear in the wild, which is the case here:
```
evaluatate->evaluate
evaluatated->evaluated
evaluatates->evaluates
evaluatating->evaluating
```

arm-in:
Thank you for the review. Changes applied.

PR #2740, opened 2023-02-12
Typos from Emacs

skangas:
This adds some typos found in Emacs over the last month and a half. I've also added several variations.


codespell_lib/data/dictionary.txt
```diff
@@ -23257,6 +23272,10 @@ materils->materials
 materla->material
 materlas->materials
 mathamatics->mathematics
+mathc->match
```
DimitriPapadopoulos, line 23275: Could it be `math` or `maths` too?

If all occurrences in the wild are actually `match` typos, keep this as is.

luzpaz (approved):

PR #2741, opened 2023-02-14
dict: Add instersection(s)->intersection(s)

akien-mga:
Seen in Godot Engine:
```
$ rg instersection
core/math/rect2i.h
90:     // Returns the instersection between two Rect2is or an empty Rect2i if there is no intersection

core/math/rect2.h
147:    // Returns the instersection between two Rect2s or an empty Rect2 if there is no intersection
```

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -21076,7 +21076,11 @@ insterad->instead
 insterrupts->interrupts
 instersction->intersection
 instersctions->intersections
+instersection->intersection
+instersectional->intersectional
+instersectionality->intersectionality
 instersectioned->intersection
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -21076,7 +21076,9 @@ insterad->instead
 insterrupts->interrupts
 instersction->intersection
 instersctions->intersections
+instersection->intersection
```
luzpaz, line 21079: Please add `intersectional->intersectional`

akien-mga: Done.

DimitriPapadopoulos (approved):

PR #2757, opened 2023-02-23
Accept duplicate entries in INI config files

### LINKED ISSUE #2727: Multiple dictionaries (add default one) for `dictionary` config setting
we have 

```shell
❯ grep dictionary .codespellrc
dictionary = .codespell_dict
```
in https://github.com/bids-standard/bids-specification/blob/HEAD/.codespellrc but that disables default dictionaries, and I need to specify `-D-` to get them used:

```shell
❯ codespell
❯ codespell -D-
./DECISION-MAKING.md:17: Curent ==> Current
./src/common-principles.md:188: entites ==> entities
```

Unfortunately, `-D` is the different style than e.g. `--skip` which expects them comma separated and `-D` doesn't work that way:

```shell
❯ git diff
diff --git a/.codespellrc b/.codespellrc
index 28aeafee..a513a1a7 100644
--- a/.codespellrc
+++ b/.codespellrc
@@ -2,5 +2,5 @@
 skip = *.js,*.svg,*.eps,.git
 ignore-words-list = fo,te,als
 builtin = clear,rare
-dictionary = .codespell_dict
+dictionary = -,.codespell_dict
 exclude-file = src/CHANGES.md
❯ codespell
...
codespell: error: argument -D/--dictionary: expected one argument
```

so the question is how could I keep using default dictionaries + custom one while relying on specification in the config file?

DimitriPapadopoulos:
Rationale:
1. The `-D` command line option can be specified multiple times. It makes sense to be able to specify the respective config file option `dictionary` multiple times too.
2. While specifying an option multiple times might be an error, setting _strict_ to `True` catches duplicates only within a single INI file, not across multiple INI files. If we are serious about handling duplicate options, we should therefore handle that manually, distinguishing between options that can be duplicated or not.

Fixes #2727.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -541,7 +541,7 @@ def parse_options(
     cfg_files = ["setup.cfg", ".codespellrc"]
     if options.config:
         cfg_files.append(options.config)
-    config = configparser.ConfigParser(interpolation=None)
+    config = configparser.ConfigParser(interpolation=None, strict=False)
 
     # Read toml before other config files.
     toml_files = []
@@ -571,7 +571,7 @@ def parse_options(
     # Collect which config files are going to be used
     used_cfg_files = []
     for cfg_file in cfg_files:
-        _cfg = configparser.ConfigParser()
+        _cfg = configparser.ConfigParser(interpolation=None, strict=False)
         _cfg.read(cfg_file)
         if _cfg.has_section("codespell"):
             used_cfg_files.append(cfg_file)


larsoner:
Worth adding a test that fails on `main` but passes here?

DimitriPapadopoulos:
Besides, it doesn't even work because the underlying data structure is still a dict: in case of duplicate keys, the new value overwrites the old value, which makes sense but is not what we would like here. I will try a few things, such as:
```ìni
dictionary = (dictionary1.txt, dictionary2.txt)
```

yarikoptic:
an alternative could be specification via `,` but also there is https://github.com/codespell-project/codespell/pull/1980 which offers to list on separate lines - IMHO might be a better choice

DimitriPapadopoulos:
#1980 does not just concatenate lines as claimed, instead it `",".join(lines)`. This will work with option `-S`/`--skip` which expects a “_comma-separated list of files to skip_”, but not with option `-D`/`--dictionary`. Therefore, it is not an alternative, we would have to modify this option to accept a “_comma-separated list of dictionaries_” anyway.

DimitriPapadopoulos:
Replaced by #2767, a better alternative (see https://github.com/codespell-project/codespell/issues/2727#issuecomment-1445132814).

PR #2759, opened 2023-02-26
Add --skip-file option: files to skip

sharonyogev:
Just like the --skip option, but instead of inline the comma separated list of files we want to skip, put them inside a file where each line is a file / glob to skipe


spetrosi:
@DimitriPapadopoulos what's the status of this? I find it to be an essential functionality and would like it to be implemented.

PR #2760, opened 2023-02-27
More precise ignore directive for mypy

DimitriPapadopoulos:


larsoner:
Thanks @DimitriPapadopoulos !

PR #2765, opened 2023-03-03
Remove whet from dictionary_rare

yarikoptic:
"Whet your appetite" is a good and used phrase. It is incorrect to suggest incorrect fixing of it.

Ref: https://github.com/codespell-project/codespell/issues/1193


DimitriPapadopoulos:
It is in _rare_, because it is a rare word and often a typo. The idea is that you don't let _codespell_ fix typos blindly.

yarikoptic:
ok, let's keep it such. formulated more general idea then in https://github.com/codespell-project/codespell/issues/2766

PR #2767, opened 2023-03-03
Options that expect a file, should accept lists of files too

### LINKED ISSUE #2727: Multiple dictionaries (add default one) for `dictionary` config setting

DimitriPapadopoulos:
The rationale is that these options readily accept multiple files from the command line, because they can be specified multiple times. However, duplicate option keys are invalid in an INI config file. The alternative is to accept multiple values for each occurrence of an option key.

Fixes #2727.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -390,10 +390,9 @@ def parse_options(
         "-D",
         "--dictionary",
         action="append",
-        help="custom dictionary file that contains spelling "
-        "corrections. If this flag is not specified or "
-        'equals "-" then the default dictionary is used. '
-        "This option can be specified multiple times.",
+        help="comma-separated list of custom dictionary files that "
+        "contain spelling corrections. If this flag is not specified "
+        'or equals "-" then the default dictionary is used.',
     )
     builtin_opts = "\n- ".join(
         [""] + [f"{d[0]!r} {d[1]}" for d in _builtin_dictionaries]
@@ -423,26 +422,26 @@ def parse_options(
         "-I",
         "--ignore-words",
         action="append",
-        metavar="FILE",
-        help="file that contains words that will be ignored "
-        "by codespell. File must contain 1 word per line."
-        " Words are case sensitive based on how they are "
-        "written in the dictionary file",
+        metavar="FILES",
+        help="comma-separated list of files that contain "
+        "words to be ignored by codespell. Files must contain "
+        "1 word per line. Words are case sensitive based on "
+        "how they are written in the dictionary file.",
     )
     parser.add_argument(
         "-L",
         "--ignore-words-list",
         action="append",
         metavar="WORDS",
-        help="comma separated list of words to be ignored "
+        help="comma-separated list of words to be ignored "
         "by codespell. Words are case sensitive based on "
-        "how they are written in the dictionary file",
+        "how they are written in the dictionary file.",
     )
     parser.add_argument(
         "--uri-ignore-words-list",
         action="append",
         metavar="WORDS",
-        help="comma separated list of words to be ignored "
+        help="comma-separated list of words to be ignored "
         "by codespell in URIs and emails only. Words are "
         "case sensitive based on how they are written in "
         'the dictionary file. If set to "*", all '
@@ -494,11 +493,13 @@ def parse_options(
     parser.add_argument(
         "-x",
         "--exclude-file",
+        action="append",
         type=str,
-        metavar="FILE",
-        help="ignore whole lines that match those "
-        "in the file FILE. The lines in FILE "
-        "should match the to-be-excluded lines exactly",
+        metavar="FILES",
+        help="ignore whole lines that match those in "
+        "the comma-separated list of files EXCLUDE. "
+        "The lines in these files should match the "
+        "to-be-excluded lines exactly",
     )
 
     parser.add_argument(
@@ -1166,20 +1167,22 @@ def main(*args: str) -> int:
     else:
         ignore_word_regex = None
 
-    ignore_words_files = options.ignore_words or []
     ignore_words, ignore_words_cased = parse_ignore_words_option(
         options.ignore_words_list
     )
-
-    for ignore_words_file in ignore_words_files:
-        if not os.path.isfile(ignore_words_file):
-            print(
-                f"ERROR: cannot find ignore-words file: {ignore_words_file}",
-                file=sys.stderr,
-            )
-            parser.print_help()
-            return EX_USAGE
-        build_ignore_words(ignore_words_file, ignore_words, ignore_words_cased)
+    if options.ignore_words:
+        ignore_words_files = flatten_clean_comma_separated_arguments(
+            options.ignore_words
+        )
+        for ignore_words_file in ignore_words_files:
+            if not os.path.isfile(ignore_words_file):
+                print(
+                    f"ERROR: cannot find ignore-words file: {ignore_words_file}",
+                    file=sys.stderr,
+                )
+                parser.print_help()
+                return EX_USAGE
+            build_ignore_words(ignore_words_file, ignore_words, ignore_words_cased)
 
     uri_regex = options.uri_regex or uri_regex_def
     try:
@@ -1196,7 +1199,7 @@ def main(*args: str) -> int:
         itertools.chain(*parse_ignore_words_option(options.uri_ignore_words_list))
     )
 
-    dictionaries = options.dictionary or ["-"]
+    dictionaries = flatten_clean_comma_separated_arguments(options.dictionary or ["-"])
 
     use_dictionaries = []
     for dictionary in dictionaries:
@@ -1258,7 +1261,9 @@ def main(*args: str) -> int:
 
     exclude_lines: Set[str] = set()
     if options.exclude_file:
-        build_exclude_hashes(options.exclude_file, exclude_lines)
+        exclude_files = flatten_clean_comma_separated_arguments(options.exclude_file)
+        for exclude_file in exclude_files:
+            build_exclude_hashes(exclude_file, exclude_lines)
 
     file_opener = FileOpener(options.hard_encoding_detection, options.quiet_level)
 
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -342,6 +342,20 @@ def test_ignore_dictionary(
     fname = tmp_path / "ignore.txt"
     fname.write_text("abandonned\nabilty\r\nackward")
     assert cs.main("-I", fname, bad_name) == 1
+    # missing file in ignore list
+    fname_missing = tmp_path / "missing.txt"
+    result = cs.main("-I", fname_missing, bad_name, std=True)
+    assert isinstance(result, tuple)
+    code, _, stderr = result
+    assert code == EX_USAGE
+    assert "ERROR:" in stderr
+    # comma-separated list of files
+    fname_dummy1 = tmp_path / "dummy1.txt"
+    fname_dummy1.touch()
+    fname_dummy2 = tmp_path / "dummy2.txt"
+    fname_dummy2.touch()
+    assert cs.main("-I", fname_dummy1, "-I", fname, "-I", fname_dummy2, bad_name) == 1
+    assert cs.main("-I", f"{fname_dummy1},{fname},{fname_dummy2}", bad_name) == 1
 
 
 def test_ignore_words_with_cases(
@@ -495,6 +509,13 @@ def test_exclude_file(
     )
     assert cs.main(bad_name) == 18
     assert cs.main("-x", fname, bad_name) == 1
+    # comma-separated list of files
+    fname_dummy1 = tmp_path / "dummy1.txt"
+    fname_dummy1.touch()
+    fname_dummy2 = tmp_path / "dummy2.txt"
+    fname_dummy2.touch()
+    assert cs.main("-x", fname_dummy1, "-x", fname, "-x", fname_dummy2, bad_name) == 1
+    assert cs.main("-x", f"{fname_dummy1},{fname},{fname_dummy2}", bad_name) == 1
 
 
 def test_encoding(
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -169,6 +169,6 @@ max-complexity = 45
 [tool.ruff.lint.pylint]
 allow-magic-value-types = ["bytes", "int", "str",]
 max-args = 13
-max-branches = 49
+max-branches = 51
 max-returns = 11
 max-statements = 119


codespell_lib/_codespell.py
```diff
@@ -1028,7 +1029,10 @@ def main(*args: str) -> int:
     else:
         ignore_word_regex = None
 
-    ignore_words_files = options.ignore_words or []
+    ignore_words_files = []
+    if options.ignore_words:
+        for iw in options.ignore_words:
+            ignore_words_files.extend(iw.split(","))
```
cclauss, line 1035: Remove a branch...
```suggestion
    for iw in options.ignore_words to []:
        ignore_words_files.extend(iw.split(","))
```

codespell_lib/_codespell.py
```diff
@@ -1053,7 +1057,9 @@ def main(*args: str) -> int:
     uri_ignore_words = parse_ignore_words_option(options.uri_ignore_words_list)
 
     if options.dictionary:
-        dictionaries = options.dictionary
+        dictionaries = []
+        for d in options.dictionary:
+            dictionaries.extend(d.split(","))
     else:
         dictionaries = ["-"]
     use_dictionaries = []
```
cclauss, line 1065: Remove the `if options.dictionary:` line above and then:
```suggestion
    dictionaries = []
    for d in options.dictionary or []:
        dictionaries.extend(d.split(","))
    dictionaries = dictionaries or ["-"]
    use_dictionaries = []
```

codespell_lib/_codespell.py
```diff
@@ -1119,7 +1125,9 @@ def main(*args: str) -> int:
 
     exclude_lines: Set[str] = set()
     if options.exclude_file:
-        build_exclude_hashes(options.exclude_file, exclude_lines)
+        for exclude_file_list in options.exclude_file:
+            for exclude_file in exclude_file_list.split(","):
+                build_exclude_hashes(exclude_file, exclude_lines)
```
cclauss, line 1130: Remove the `if options.exclude_file:` line above and then
```suggestion
    for exclude_file_list in options.exclude_file or []:
        for exclude_file in exclude_file_list.split(","):
            build_exclude_hashes(exclude_file, exclude_lines)
```

cclauss (changes requested):
Let's reduce branches and complexity

codespell_lib/_codespell.py
```diff
@@ -984,6 +985,12 @@ def parse_file(
     return bad_count
 
 
+def flatten_comma_separated_arguments(
+    arguments: List(str),
+) -> List(str):
```
cclauss, line 990: These changes are great!  Simple but effective -- well done.

Add a doctest?
```suggestion
) -> List(str):
    """
    >>> flatten_comma_separated_arguments(None)
    []
    """
```

codespell_lib/_codespell.py
DimitriPapadopoulos: You fold the branch by changing `if` to `or`. Complexity is not really reduced.

I believe this a shortcoming of the complexity metrics, and this looks like a trick that makes use of that shortcoming, instead of tackling real complexity.

codespell_lib/_codespell.py
DimitriPapadopoulos: Yes, I'll look into it. This is just a draft, I'm not even sure it works well when an option is missing. I need tests.

DimitriPapadopoulos: Thank you for helping out, by the way. I hope this change will actually decrease code complexity.

codespell_lib/_codespell.py
```diff
@@ -984,6 +985,22 @@ def parse_file(
...
 
 
+def flatten_comma_separated_arguments(
+    arguments: List[str],
+) -> List[str]:
+    """
+    >>> flatten_comma_separated_arguments([])
+    []
+    >>> flatten_comma_separated_arguments(None)
+    None
```
cclauss, line 995: The type hint on this function promises to return a `list`.

DimitriPapadopoulos: Yes, that's disturbing. Does this mean that these type hints are useless?

codespell_lib/_codespell.py
```diff
@@ -984,6 +985,18 @@ def parse_file(
...
+def flatten_comma_separated_arguments(
+    arguments: List[str],
+) -> List[str]:
+    """
+    >>> flatten_comma_separated_arguments(["a,b,c", "d", "e"])
+    ['a', 'b', 'c', 'd', 'e']
+    >>> flatten_comma_separated_arguments([])
+    []
+    """
+    return [item for argument in arguments for item in argument.split(",")]
```
cclauss, line 997:
```suggestion
    >>> flatten_comma_separated_arguments(None)
    []
    """
    return [item for argument in arguments or [] for item in argument.split(",")]
```

cclauss (approved):

codespell_lib/_codespell.py
cclauss: Python is a dynamic / weakly typed language so type hints are just hints.  They have _no effect at runtime_.  They are designed to help tools like `mypy` to find type inconsistencies and bugs.  Using type hints is strongly recommended but they do not make Python a strongly typed language like Go, Rust, or TypeScript.

> __Note__  The Python runtime does not enforce function and variable type annotations. They can be used by third-party tools such as type checkers, IDEs, linters, etc.

https://docs.python.org/3/library/typing.html

DimitriPapadopoulos: Yes, I understand that. However, we do run mypy in a CI job. If mypy is unable to detect errors, what's the point of these type hints?

codespell_lib/_codespell.py
DimitriPapadopoulos: I had added `or []` in the arguments of `flatten_comma_separated_arguments`, but it's much better this way., it is better to make the function itself lenient.

DimitriPapadopoulos: But then it means we can pass `None` to a function that is documented to accept `List(str)` only. Again, I am puzzled by these type hints.

cclauss: You are not alone.  The tools are not perfect but the situation is better than before type hints.  The more hints you put into your functions and their return types, the better the type checkers get and finding issues.  Who is [the fifth most prolific contributor to `mypy`](https://github.com/python/mypy/graphs/contributors)?

sharonyogev:
if we touch this mechanism, maybe we want to combine it with my PR? https://github.com/codespell-project/codespell/pull/2759
Since I'm adding a new argument that can except file, we should at lease be aware of each other

cclauss:
Blocked by #2797

codespell_lib/_codespell.py
```diff
@@ -338,10 +338,9 @@ def parse_options(
         "-D",
         "--dictionary",
         action="append",
-        help="custom dictionary file that contains spelling "
-        "corrections. If this flag is not specified or "
-        'equals "-" then the default dictionary is used. '
-        "This option can be specified multiple times.",
+        help="comma-separated list of custom dictionary files that "
+        "contain spelling  corrections. If this flag is not specified "
```
larsoner, line 342:
```suggestion
        "contain spelling corrections. If this flag is not specified "
```

codespell_lib/_codespell.py
```diff
@@ -371,26 +370,26 @@ def parse_options(
...
-        metavar="FILE",
-        help="file that contains words that will be ignored "
-        "by codespell. File must contain 1 word per line."
-        " Words are case sensitive based on how they are "
-        "written in the dictionary file",
+        metavar="FILES",
+        help="comma-separated list of files that contain "
+        "words to be ignored by codespell. Files must contain "
+        "1 word per line. Words are case sensitive based on "
+        "how they are  written in the dictionary file.",
```
larsoner, line 377:
```suggestion
        "how they are written in the dictionary file.",
```

codespell_lib/_codespell.py
```diff
@@ -983,6 +984,20 @@ def parse_file(
...
+) -> List[str]:
+    """
+    >>> flatten_comma_separated_arguments(["a,b,c", "d", "e"])
+    ['a', 'b', 'c', 'd', 'e']
+    >>> flatten_comma_separated_arguments([])
+    []
+    >>> flatten_comma_separated_arguments(None)
+    []
+    """
+    return [item for argument in arguments or [] for item in argument.split(",")]
```
larsoner, line 998: Arguments is always list of str so why do you need the `arguments or []`?
```suggestion
    return [item for argument in arguments for item in argument.split(",")]
```

cclauss: Because the doctest on lines 995 and 996 will fail.

codespell_lib/_codespell.py
```diff
@@ -983,6 +984,20 @@ def parse_file(
...
 
+def flatten_comma_separated_arguments(
+    arguments: List[str],
+) -> List[str]:
+    """
+    >>> flatten_comma_separated_arguments(["a,b,c", "d", "e"])
+    ['a', 'b', 'c', 'd', 'e']
+    >>> flatten_comma_separated_arguments([])
+    []
+    >>> flatten_comma_separated_arguments(None)
```
larsoner, line 995: This one shouldn't be allowed based on the type hints (you'd need an `Optional`), and it's not used in code. Let's remove it

DimitriPapadopoulos: I understand, but unfortunately mypy is unable to enforce all hints, and I have always read that Pythonic functions should be lenient (probably read that before mypy).

larsoner: Sure but we don't need to be lenient here. I would only add leniency when it's needed. To put it differently, the `or []` branch isn't useful now, so why add it now? It seems like over-planning for a future, i.e., YAGNI

codespell_lib/_codespell.py
DimitriPapadopoulos: I was trying not to increase complexity by avoiding `if` where this function is called.

I had ended up adding an `or []` in the function arguments, then in the function itself to avoid multiple `or []` wherever it is called (and for leniency).

And now I am back to `if` because that's what's makes sense IMHO: do not call the function if that's not needed, and for that, you do need the `if`'s.

larsoner:
Looks like now you just need a regression test that fails on `master` but passes on this PR, and also to relax the Ruff checks

codespell_lib/_codespell.py
DimitriPapadopoulos: In all case, I was actually battling against complexity metrics that don't want new `if`s, and perceived complexity too.

First added `or []` in function arguments, wherever we call the function.

Then decided it might be better to have a single location for `or []`, at the expense of changing the signature of the function to accept a `None` argument too. But since mypy doesn't catch such violations, I thought that might be a feature and not a mypy shortcoming. So, I left the signature as is – to avoid increasing perceived complexity even more.

Then I went back to the first solution, but kept the second solution for leniency. But I agree that's over-engineering.

And finally, I'm back to the original code. And I think we do need to **accept increased complexity** when slightly modifying existing options or adding new options. The only right way to reduce complexity here is to modify the way we handle **all** options, **consistently**.

DimitriPapadopoulos:
It's not really a regression. Code coverage yells because I have to move a little bit already untested lines as far as I can see.

I could probably fix that by passing nonexistent files as arguments.

larsoner:
> It's not really a regression.

That's not what I mean -- by "regression test" I mean one that ensures we don't make any *future* regressions by breaking what you have added/changed here. In a test-driven development (TDD) sense, I want you to add a test that ensures that your comma parsing works (in a way that doesn't on current `master`).

> As far as I can see, code coverage screams because I have to move untested lines around.

This is a separate issue, but it would be nice also to fix this with a test if you could

DimitriPapadopoulos:
Ah, got it. Sure, tests are needed.

DimitriPapadopoulos:
By the way, I find current tests really hard to write. Some of them could rely a small set of permanent test files instead of creating temporary files on the fly. I find it much easier to create and maintain test file hierarchies using whatever tool fits the task best than programmatically creating them.

yarikoptic:
ping on a review for this PR.

larsoner:
Still needs a regression test

yarikoptic:
> By the way, I find current tests really hard to write. Some of them could rely a small set of permanent test files instead of creating temporary files on the fly. I find it much easier to create and maintain test file hierarchies using whatever tool fits the task best than programmatically creating them.

A salesman here: I was like you but more often we (in other projects like DataLad) started ran into situations where that "permanent test files" had "too many features" already in them thus making it hard to change them without introducing some regression etc.  So even if having some "prototypical" cases is still good for overall large integration testing, having succinct cases crafted specifically for tests are easier to understand/maintain etc.  So here I want to present a helper which we had and use for probably a decade:  `@with_tree` decorator which is handy to create arbitrary hierarchies of files and even some (e.g. `.tar.gz`) archives: https://github.com/datalad/datalad/blob/HEAD/datalad/tests/utils_pytest.py#L646 (decorator relies on some other helpers so is not yet as easily copy pasteable but could be made so I guess) with some examples of use

```python
@with_tree(tree={
     'somefile': 'file content',
     'subddir': {'file_in_sub': 'other'},
     'archive.tar.gz': {'file_in_archive': "content", "empty_folder": {}},
})
def test_smth(path=None):
```

You can [find more uses in datalad (now pytest) tests](https://github.com/datalad/datalad/search?q=with_tree&type=code)

dalito:
I modified one the test a little https://github.com/codespell-project/codespell/blob/4bc506312e780c2e1d56177c7c4dde7bf8babc27/codespell_lib/tests/test_basic.py#L1154 to cover configuring "skip" as a list in cfg or toml. The test does not pass for lists. I don't have more time to look deeper. Here is the code:

```python
@pytest.mark.parametrize(
    ("kind", "skip_conf"),
    (
        ("cfg", "skip = bad.txt, whatever.txt"),
        ("cfg", "skip =\n    bad.txt\n    whatever.txt\n"),
        ("toml", "skip = 'bad.txt,whatever.txt'"),
        ("toml", "skip = [\n  'bad.txt',\n  'whatever.txt',\n]"),
    ),
    ids=("cfg-str", "cfg-list", "toml-str", "toml-list"),
)
def test_config_toml(
    tmp_path: Path,
    capsys: pytest.CaptureFixture[str],
    kind: str,
    skip_conf: str,
) -> None:
    """Test loading options from a config file or toml."""
    d = tmp_path / "files"
    d.mkdir()
    (d / "bad.txt").write_text("abandonned donn\n")
    (d / "good.txt").write_text("good")

    # Should fail when checking both.
    result = cs.main(d, count=True, std=True)
    assert isinstance(result, tuple)
    code, stdout, _ = result
    # Code in this case is not exit code, but count of misspellings.
    assert code == 2
    assert "bad.txt" in stdout

    if kind == "cfg":
        conffile = tmp_path / "setup.cfg"
        args_ = ("--config", conffile)
        conffile.write_text(
            f"""\
[codespell]
{skip_conf}
count =
"""
        )
    else:
        assert kind == "toml"
        if sys.version_info < (3, 11):
            pytest.importorskip("tomli")
        tomlfile = tmp_path / "pyproject.toml"
        args_ = ("--toml", tomlfile)
        tomlfile.write_text(
            f"""\
[tool.codespell]
{skip_conf}
count = false
"""
        )

    # Should pass when skipping bad.txt
    result = cs.main(d, *args_, count=True, std=True)
    assert isinstance(result, tuple)
    code, stdout, _ = result
    assert code == 0
    assert "bad.txt" not in stdout

    # And both should automatically work if they're in cwd
    cwd = Path.cwd()
    try:
        os.chdir(tmp_path)
        result = cs.main(d, count=True, std=True)
        assert isinstance(result, tuple)
        code, stdout, _ = result
    finally:
        os.chdir(cwd)
    assert code == 0
    assert "bad.txt" not in stdout
```

DimitriPapadopoulos:
@dalito It works for me but lacks CI tests. Most probably, the above test don't test the right thing.

dalito:
Hmm, my projects use only toml since 2 years. But good to know the cfg works. I also saw that in my test but it was required to split list from cfg by newline and ignore empty newlines. I tried to debug the toml problem and saw toml is strangely handled: First toml lists are read correctly as python list but then they are converted to strings again and packed into a list. Then extracting paths from this strange string fails leading to ignoring the options in a toml list.

DimitriPapadopoulos:
It works for me like this:
```
ignore-words = "file1,file2"
```
Unfortunately, codespell has been reading its configuration using configparser. TOML is not directly compatible with configparser and INI files. Real support for TOML would require to a rewrite of the code handling configuration and probably add quite a few lines of code, as Python has not standard way to handle configuration from command-line, INI and TOML options. I tried to fix things a little bit in `_toml_to_parseconfig`, perhaps I can add list handlign in there.

DimitriPapadopoulos:
I just pushed a small change to `_toml_to_parseconfig()` to try to change lists into a string as expected by configparser. It doesn't work yet, needs more work. But you get the idea.

This means of course that you cannot specify file names that contain a comma, since the comma is considered as an argument separator. I don't know if it is possible to escape a comma in configparser.

DimitriPapadopoulos:
Actually, it looks like `ignore-words = "file1,file2"` is actually the format expected by codespell for lists!
https://github.com/codespell-project/codespell/blob/4bc506312e780c2e1d56177c7c4dde7bf8babc27/codespell_lib/tests/test_basic.py#L1193-L1196

Perhaps we can keep this strange TOML syntax in this PR, and defer better support for TOML to a different PR.

dalito:
Thanks for resuming work on this! My interest in the list support results from the need to skip several different files. Configuring several file paths in a long list gets hard to read quickly.

The toml I tried (as generated in the test above) was
```toml
[tool.codespell]
skip = [
  'bad.txt',
  'whatever.txt',
]
count = false
```

The cfg that needed additional splitting by newline was:
```ini
[codespell]
skip =
    bad.txt
    whatever.txt
count =
```
Both are using the standard way to add a list to the respective configuration format. If I understand this PR correctly its idea is to add support for exactly this?

DimitriPapadopoulos:
No, the idea of this PR is to modify options that didn't support lists, so that they can support lists in the same way as options that already support lists, like `skip` that expects a “_comma separated list of words_”`:
```toml
[tool.codespell]
skip = 'bad.txt,whatever.txt'
count = false
```
```ini
[codespell]
skip = bad.txt,whatever.txt
count =
```

dalito:
Ok. Thanks for the clarification.

SherAndrei:
🏓

larsoner:
I think we're still waiting for a regression test here

DimitriPapadopoulos:
Yes, I need to find some time to write the tests.

DimitriPapadopoulos:
@larsoner I have added tests. Could you have a look?

larsoner (approved):

larsoner:
Thanks @DimitriPapadopoulos !

PR #2771, opened 2023-03-06
ENH: Color matching wrong word in the interactive session

yarikoptic:
Aging me found it difficult to locate an offending word in the context line provided in interactive mode.  Coloring it makes it very obvious and worked like a charm on my sample cases.  But I think implementation should ideally be improved but I would need guidance on how to get detail of specific offending word location

- [x] Get a more specific way to locate the wrong word within the line or remove TODO


yarikoptic:
I guess smth like following patch would be needed

<details>
<summary>use re.finditer and operate on re.Match instead of on `findall` words</summary> 

```patch
diff --git a/codespell_lib/_codespell.py b/codespell_lib/_codespell.py
index 9052e427..f23b02c5 100644
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -680,11 +680,12 @@ def fix_case(word: str, fixword: str) -> str:
 
 def ask_for_word_fix(
     line: str,
-    wrongword: str,
+    match: re.match,
     misspelling: Misspelling,
     interactivity: int,
     colors: TermColors,
 ) -> Tuple[bool, str]:
+    wrongword = match.group()
     if interactivity <= 0:
         return misspelling.fix, fix_case(wrongword, misspelling.data)
 
@@ -695,12 +696,7 @@ def ask_for_word_fix(
             # TODO: this might be different from how word is parsed etc.
             # Ideally should operate on some information pointing to exact
             # location of misspelling in the line
-            line_ui = re.sub(
-                rf"\b({wrongword})\b",
-                rf"{colors.WWORD}\1{colors.DISABLE}",
-                line,
-                flags=re.IGNORECASE,
-            )
+            line_ui = rf"{line[:match.start()]}{colors.WWORD}{wrongword}{colors.DISABLE}{line[match.end():]}"
             print(f"{line_ui}\t{wrongword} ==> {fixword} (Y/n) ", end="", flush=True)
             r = sys.stdin.readline().strip().upper()
             if not r:
@@ -757,26 +753,34 @@ def extract_words(
     text: str,
     word_regex: Pattern[str],
     ignore_word_regex: Optional[Pattern[str]],
-) -> List[str]:
+    finditer: bool = False
+) -> List[str|re.Match]:
     if ignore_word_regex:
         text = ignore_word_regex.sub(" ", text)
-    return word_regex.findall(text)
+    if finditer:
+        return list(word_regex.finditer(text))
+    else:
+        return word_regex.findall(text)
 
 
 def apply_uri_ignore_words(
-    check_words: List[str],
+    check_matches: List[re.Match],
     line: str,
     word_regex: Pattern[str],
     ignore_word_regex: Optional[Pattern[str]],
     uri_regex: Pattern[str],
     uri_ignore_words: Set[str],
-) -> None:
+) -> List[re.Match]:
     if not uri_ignore_words:
-        return
+        return check_matches
     for uri in re.findall(uri_regex, line):
         for uri_word in extract_words(uri, word_regex, ignore_word_regex):
             if uri_word in uri_ignore_words:
-                check_words.remove(uri_word)
+                check_matches = [
+                    match for match in check_matches
+                    if match.group() == uri_word
+                ]
+    return check_matches
 
 
 def parse_file(
@@ -865,18 +869,18 @@ def parse_file(
         # outside, it will still be a spelling error.
         if "*" in uri_ignore_words:
             line = uri_regex.sub(" ", line)
-        check_words = extract_words(line, word_regex, ignore_word_regex)
+        check_matches = extract_words(line, word_regex, ignore_word_regex, finditer=True)
         if "*" not in uri_ignore_words:
-            apply_uri_ignore_words(
-                check_words,
+            check_matches = apply_uri_ignore_words(
+                check_matches,
                 line,
                 word_regex,
                 ignore_word_regex,
                 uri_regex,
                 uri_ignore_words,
             )
-
-        for word in check_words:
+        for match in check_matches:
+            word = match.group()
             lword = word.lower()
             if lword in misspellings:
                 context_shown = False
@@ -889,7 +893,7 @@ def parse_file(
                         print_context(lines, i, context)
                     fix, fixword = ask_for_word_fix(
                         lines[i],
-                        word,
+                        match,
                         misspellings[lword],
                         options.interactive,
                         colors=colors,

```
</details>

larsoner:
@yarikoptic I have no ideas for how to find the correct word but it looks like you've pushed a commit, should I test it locally now?

larsoner:
Seems to work at my end after a couple of tiny commits to make type hints + mypy happy:

Screenshot from 2023-03-07 12-35-52

@yarikoptic can you look at why the URIs are broken? If not I can continue to push

larsoner:
It looks like the remaining bug has to do with the case:
```
    with open(op.join(d, "flag.txt"), "w") as f:
        f.write("abandonned http://example.com/abandonned\n")
    # Test file has 2 invalid entries.
    assert cs.main(f.name) == 2
    # Ignoring the value in the URI won't ignore the word completely.
    assert cs.main(f.name, "--uri-ignore-words-list=abandonned") == 1
```
In other words, the code should only ignore one of the two errors, i.e., the one that occurs in the URI (not the one that precedes it).

yarikoptic:
@larsoner thanks for fixups -- I pushed 5ab652060c0afd9e70ece69e50eff84a492a5a27 for that outstanding test failure, thanks for describing it. See if you like it better/prettier.

larsoner (approved):

larsoner:
All green and seems to work locally, let's merge -- thanks this is cool @yarikoptic !

yarikoptic:
oh, that was fast! ;) Thank you @larsoner for quick review and acceptance. Thank you for all the work on codespell, I keep *making friends* introducing it virtually to every project I touch ;)

PR #2774, opened 2023-03-08
BUG: Fix bug with toml triage

larsoner:
Let's first reproduce https://github.com/codespell-project/codespell/pull/2625#issuecomment-1460644349 then fix it


pcaversaccio:
cool - let me know if you need further data to reproduce the bug.

larsoner:
d24b03f shows the correct failure pattern, will fix and then merge when all but the `actions-codespell/master` pass. That won't work until we cut a release that does... so I'll cut a release, then try a PR with no commits to verify it works

PR #2782, opened 2023-03-09
atalog typo (saw twice in a project)

yarikoptic:
analog is closeby word (one letter difference) but `t` and `n` far on keyboard so I think unlikely to happen. but can make it multi choice if someone thinks it is worth it

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3699,6 +3699,7 @@ atachment->attachment
 atachments->attachments
 atack->attack
 atain->attain
+atalog->catalog
 atatch->attach
 atatchable->attachable
 atatched->attached


luzpaz (approved):

DimitriPapadopoulos:
Perhaps add `catalog` as a second fix?

yarikoptic:
> Perhaps add `catalog` as a second fix?

ATM it is the one and only... ;-)

DimitriPapadopoulos:
:blush: Of course, sorry about that.

codespell_lib/data/dictionary.txt
```diff
@@ -3699,6 +3699,7 @@ atachment->attachment
 atachments->attachments
 atack->attack
 atain->attain
+atalog->catalog
```
DimitriPapadopoulos, line 3702: Perhaps add plural, I've seen a few occurrences as well:
```
atalogs->catalogs
```

yarikoptic:
```suggestion
atalog->catalog
atalogs->catalogs
```

PR #2785, opened 2023-03-10
Fix AssertionError when opening with chardet

n-thumann:
This PR fixes an AssertationError, when chardet is unable to detect the encoding.

Added in https://github.com/codespell-project/codespell/commit/8d0d82bcc07aaef5211127683aba7de67b923aac, it is checked if chardet has actually detected an encoding. If it's unable to detect one, an AssertationError is raised and codespell fails hard.

Example:
```
➜  ~ touch /tmp/empty_file
➜  ~ file /tmp/empty_file                 
/tmp/empty_file: empty
➜  ~ codespell /tmp/empty_file                                           
➜  ~ echo $?
0
➜  ~ codespell --hard-encoding-detection /tmp/empty_file                 
  File "/home/nthumann/.local/bin/codespell", line 8, in <module>
    sys.exit(_script_main())
  File "/home/nthumann/.local/lib/python3.10/site-packages/codespell_lib/_codespell.py", line 989, in _script_main
    return main(*sys.argv[1:])
  File "/home/nthumann/.local/lib/python3.10/site-packages/codespell_lib/_codespell.py", line 1178, in main
    bad_count += parse_file(
  File "/home/nthumann/.local/lib/python3.10/site-packages/codespell_lib/_codespell.py", line 871, in parse_file
    lines, encoding = file_opener.open(filename)
  File "/home/nthumann/.local/lib/python3.10/site-packages/codespell_lib/_codespell.py", line 218, in open
    return self.open_with_chardet(filename)
  File "/home/nthumann/.local/lib/python3.10/site-packages/codespell_lib/_codespell.py", line 231, in open_with_chardet
    assert encoding is not None
AssertionError
➜  ~ echo $?
1
```

With this patch we use always return `f.encoding` instead of `encoding`, which may be `None`, if chardet is unable to detect it. `f.encoding` either contains the encoding passed to `open` (if chardet detect it) or the encoding Python detected (if `encoding` is `None`).
Simply removing the assertation will make mypy unhappy, because the returned `encoding` can be `None`.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -228,7 +228,6 @@ class FileOpener:
                     break
         self.encdetector.close()
         encoding = self.encdetector.result["encoding"]
-        assert encoding is not None  # noqa: S101
 
         try:
             f = open(filename, encoding=encoding, newline="")
@@ -245,7 +244,7 @@ class FileOpener:
             lines = f.readlines()
             f.close()
 
-        return lines, encoding
+        return lines, f.encoding
 
     def open_with_internal(self, filename: str) -> Tuple[List[str], str]:
         encoding = None
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -384,6 +384,16 @@ def test_encoding(
     assert "WARNING: Binary file" in stderr
 
 
+def test_unknown_encoding_chardet(
+    tmp_path: Path,
+    capsys: pytest.CaptureFixture[str],
+) -> None:
+    """Test opening a file with unknown encoding using chardet"""
+    fname = tmp_path / "tmp"
+    fname.touch()
+    assert cs.main("--hard-encoding-detection", fname) == 0
+
+
 def test_ignore(
     tmp_path: Path,
     capsys: pytest.CaptureFixture[str],


codespell_lib/_codespell.py
```diff
@@ -228,7 +228,8 @@ def open_with_chardet(self, filename: str) -> Tuple[List[str], str]:
                     break
         self.encdetector.close()
         encoding = self.encdetector.result["encoding"]
-        assert encoding is not None
```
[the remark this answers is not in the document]
DimitriPapadopoulos: I clearly don't know much about mypy, I don't understand why it would change the signature of the function if we remove `assert`, especially since assert could be [optimised away](https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-assert_stmt).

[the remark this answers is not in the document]
DimitriPapadopoulos: Ah, got it. :blush:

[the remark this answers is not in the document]
DimitriPapadopoulos: I would favour [locale.getencoding()](https://docs.python.org/3/library/locale.html#locale.getencoding) because it's the default value of the `encoding` argument of open(). Perhaps someone else could look into this... @larsoner?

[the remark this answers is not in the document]
larsoner: `locale.getencoding()` seems reasonable to me assuming it works. Or get mypy to skip its check or whatever is fine, too...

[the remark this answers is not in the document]
n-thumann: Unfortunately [that method](https://docs.python.org/3.11/library/locale.html#locale.getencoding) is new in Python 3.11, so this will break with all lower Python versions.
How about we use [`getpreferredencoding(False)`](https://docs.python.org/3.11/library/locale.html#locale.getpreferredencoding) instead? This is what Python <= 3.10 use when `encoding` in `None` when calling `open`.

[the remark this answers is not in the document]
DimitriPapadopoulos: Yes, doing whatever `open` does internally looks like the best course of action.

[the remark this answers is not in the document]
n-thumann: Using `locale.getpreferredencoding(False)` now & rebased to fix merge conflict :v:

codespell_lib/_codespell.py
```diff
@@ -228,7 +229,8 @@ def open_with_chardet(self, filename: str) -> Tuple[List[str], str]:
                     break
         self.encdetector.close()
         encoding = self.encdetector.result["encoding"]
-        assert encoding is not None  # noqa: S101
+        if not encoding:
+            encoding = locale.getpreferredencoding(False)
```
larsoner, line 233: Thinking about this again (sorry I didn't realize this earlier!) -- I am a bit torn on this. On the one hand this might tend to make things work for people. On the other, this means behavior on one machine will differ from another. That makes bugs and reproducibility harder for people, especially because this will be silent. WDYT about just setting `encoding = 'utf8'` as the fallback instead?

n-thumann: Good point regarding the different behavior on different machines, but forcing a default encoding could lead to problems and I think it might make more sense to let Python cope with it. How about:
Instead of setting a encoding in before opening the file (either using a default or `getpreferredencoding`) we keep the old behavior. If we pass `None` as `encoding` to `open` Python will take care of it, just like before.
But instead of returning a possible `None` encoding, we return either `encoding` or `f.encoding`, because the latter will contain the encoding detected by Python after successfully opening the file.
```
>>> f = open("empty_file")
>>> f.encoding
'UTF-8'
>>> f = open("empty_file", encoding="ASCII")
>>> f.encoding
'ASCII'
```
This will have the same behavior like now, but letting Python take care of selecting the default encoding.

n-thumann: > it would be really nice to cover this line in a test

Done! Writing problematic bytes into the file isn't really necessary though, because the bug can be triggered by an empty file :)

larsoner: >  If we pass None as encoding to open Python will take care of it, just like before.

Works for me!

n-thumann: Done!
I also fixed the formatting error reported by black. There also were some other errors reported by ruff, but in lines I didn't even touch. I don't need to worry about them, do I?

DimitriPapadopoulos: If you rebase, ruff errors should disappear.

n-thumann: Ah, I already did and now they disappeared. Thanks :)

codespell_lib/_codespell.py
```diff
@@ -245,7 +244,7 @@ def open_with_chardet(self, filename: str) -> Tuple[List[str], str]:
             lines = f.readlines()
             f.close()
 
-        return lines, encoding
+        return lines, f.encoding
```
DimitriPapadopoulos, line 247: Excellent, we return [io.TextIOBase](https://docs.python.org/3/library/io.html#io.TextIOBase).[encoding](https://docs.python.org/3/library/io.html#io.TextIOBase.encoding), as actually set by [open](https://docs.python.org/3/library/functions.html#open) in text mode (the default).

DimitriPapadopoulos (approved):

codespell_lib/_codespell.py
```diff
@@ -228,7 +228,6 @@ def open_with_chardet(self, filename: str) -> Tuple[List[str], str]:
                     break
         self.encdetector.close()
         encoding = self.encdetector.result["encoding"]
-        assert encoding is not None  # noqa: S101
```
DimitriPapadopoulos, line 231: The assertion was introduced by `8d0d82b` / #2588 in an attempt to please mypy. That's not a proper fix, especially since Python can optimise assertions away, but mypy did give a good hint to look into this part of the code. And we now have the proper fix!

PR #2787, opened 2023-03-10
ruff: Use format specifiers instead of percent format

cclauss:
% `ruff --select=UP031 --fix .`
```
Found 3 errors (3 fixed, 0 remaining).
```
% `ruff rule UP031`
# printf-string-formatting (UP031)

Derived from the **pyupgrade** linter.

Autofix is always available.

Message formats:
* Use format specifiers instead of percent format


larsoner:
Thanks @cclauss !

PR #2803, opened 2023-03-18
Kwanza = currency of Angola

### LINKED ISSUE #2794: Incorrect spelling suggestion for the Angolan currency Kwanza
https://en.wikipedia.org/wiki/Angolan_kwanza
https://en.wikipedia.org/wiki/Kwanza,_Kenya
https://en.wikipedia.org/wiki/Kwanzaa

The newest release of codespell requests this change:
* web_programming/currency_converter.py:26: Kwanza ==> Kwanzaa

Clearly, both are valid spellings for different things.
* https://en.wikipedia.org/wiki/Kwanza_(disambiguation)

DimitriPapadopoulos:
Fixes  #2794.


peternewman (approved):
LGTM.

We should really work out a method to block words like this from ever getting back in, a dictionary of words banned from codespell would be the easiest way to implement it...

PR #2804, opened 2023-03-20
Add tweek(s)->tweak(s) correction

cfi-gb:
Note: Not 100% sure if we should place this in `dictionary_code.txt` instead because there seems to be a `Tweek` lib and some times also a usage of `tweek` some kind of abbreviation for a `week` variable.

See e.g.:
- https://grep.app/search?q=tweeks&words=true
- https://grep.app/search?q=tweek&words=true

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -39162,6 +39162,10 @@ tust->trust
 tution->tuition
 tutoriel->tutorial
 tutoriels->tutorials
+tweek->tweak
+tweeked->tweaked
+tweeking->tweaking
+tweeks->tweaks
 tweleve->twelve
 twelth->twelfth
 two-dimenional->two-dimensional


cfi-gb:
> A `tweek` variable would be the kind of typos we aim to flag.

But `tweek` might be also completely valid as some kind of abbreviation like seen e.g. here:

https://github.com/redmine/redmine/blob/92bd67c279516f339866e95c702b39cee325ef05/app/models/time_entry.rb

That was the reason why i'm not that sure but maybe such special cases are acceptable and the project in question could add exclusions?

> Perhaps add all verbal forms?
> 
> ```
> tweeked-> tweaked
> tweeking->tweaking
> ```

Ah, indeed. Thanks a lot for the hint, added it now :+1:

DimitriPapadopoulos:
Any typo could be seen as a valid abbreviation.

DimitriPapadopoulos (approved):

cfi-gb:
> Any (or almost any) typo could be seen as a valid abbreviation.

Fair enough, thanks a lot for the explanation :+1:

PR #2812, opened 2023-03-28
Nome = A city in Alaska.

### LINKED ISSUE #2811: Incorrect spelling suggestion for "nome" to "gnome" - Nome, Alaska is a real place
PR #2675 added a mapping for "nome" to "gnome".

I have no doubt that people make this mistake frequently, but Nome, Alaska does exist and is listed in the IANA Timezone database's [zone.tab](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones#NOME) file.

jfx2006:
Fixes #2811.

### DIFF
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -137,7 +137,7 @@ moue->mouse
 multistory->multistorey, multi-storey,
 nickle->nickel
 noes->nose, knows, nodes, does,
-nome->gnome
+nome->gnome, none,
 ochry->ochery, ochrey,
 oerflow->overflow
 oerflowed->overflowed
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -137,7 +137,6 @@ moue->mouse
 multistory->multistorey, multi-storey,
 nickle->nickel
 noes->nose, knows, nodes, does,
-nome->gnome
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Perhaps it could be moved to _rare_ instead. There might be 500,000 inhabited places in the world. If we take into account each one of them, I suspect plenty of common typos might disappear from the dictionary. Because Nome is listed in the IANA Timezone database, I am not against changing the typo database. But how often does `nome` stand for the city of `Nome` and not for `none`? That's why I'm suggesting moving it to _rare_.

Note that [nome](https://en.wikipedia.org/wiki/Nome_(Egypt)), from νομός, was a territorial division in Ancient Egypt. A reason more to change the current state of the database, but _rare_ might be the right balance.

Most occurrences of “nome” in GitHub appear to stand for “name” in Portuguese. I don't have the time to compare the frequency of “Nome” as the city and “nome” as a typo of “none”.

DimitriPapadopoulos:
Ah, my wrong, `nome` is indeed already in _rare_.

I also vote for changing `nome->gnome` to `nome->gnome, none,`. It will stop auto-correction but keep flagging a possible typo.

DimitriPapadopoulos:
Isn't `nome->gnome, none,` preferable? Both look like valid fixes, plus this would disable autocorrection.

DimitriPapadopoulos (approved):

skangas (approved):

PR #2815, opened 2023-03-29
Add vai->via, vie

peternewman:


DimitriPapadopoulos:
I guess `vie` would a fix for a phonetic misspelling, while `via` would be a fix for a keyboard typo, right?

DimitriPapadopoulos (approved):

peternewman:
> I guess `vie` would a fix for a phonetic misspelling, while `via` would be a fix for a keyboard typo, right?

Indeed!

PR #2818, opened 2023-04-01
Add gadjet->gadget and friend

peternewman:
I don't know if we should avoid it because of this? https://en.wiktionary.org/wiki/gadjet


luzpaz:
Try again :wink:

peternewman:
:facepalm: I was err just testing our CI...

Thanks for the spot @luzpaz !

DimitriPapadopoulos (approved):

PR #2820, opened 2023-04-16
some fixes

marksmayo:
simplified conditionals
added maxsplit to a split
simplified an if not check


larsoner (approved):
LGTM, updating with `master` and marking for merge when green. Thanks in advance @marksmayo !

PR #2845, opened 2023-04-27
Add 'requestor->requester'

sirosen:


DimitriPapadopoulos:
Note that other native speakers might think `requestor` is correct English:
https://github.com/openssl/openssl/pull/20910#discussion_r1188611363

From dictionaries:
* [Oxford Learner's Dictionary](https://www.oxfordlearnersdictionaries.com/spellcheck/english/?q=requester) lacks both forms
* [Cambridge](https://dictionary.cambridge.org/dictionary/english/requester?q=requestor) suggests **_requester_**, but _requestor_ is also acceptable.
* [Collins](https://www.collinsdictionary.com/dictionary/english/requestor) suggests **_requester_** too, but _requestor_ is "a rare spelling of _[requester](https://www.collinsdictionary.com/dictionary/english/request)_".
* [Merriam-Webster](https://www.merriam-webster.com/dictionary/requestor) knows only **_requester_**.

Google Ngram Viewer tends to agree it is [_requester_ over _requestor_](https://books.google.com/ngrams/graph?content=requestor%2Crequester&year_start=1800&year_end=2019&corpus=en-2019&smoothing=3):

requestor

From [What's the difference between "requester" and "requestor"?](https://english.stackexchange.com/questions/29254/whats-the-difference-between-requester-and-requestor):
> Both are in [dictionaries](http://en.wiktionary.org/wiki/requestor). I've heard people insist "requester" is correct for a person who requests something, and that "requestor" is wrong there, leaving me to wonder how it is used. Requestor happens to be marked as an error in my browser's spellchecker.
> 
> [COCA for requester/requestor](https://www.english-corpora.org/coca/?c=coca&q=10713921) indicates that requester is used for people and requestor is used in technical senses like a part of a program. A [google ngram for requester/requestor](https://books.google.com/ngrams/graph?content=requester%2Crequestor&year_start=1900&year_end=2008&corpus=15&smoothing=3) shows that requester is also a bit more common.

Not sure we should keep this fix in codespell though. Opinions?

sirosen:
That's interesting background. I wasn't aware that "requestor" could be valid.

One of the uses of codespell which I care about is ensuring that documentation -- which may be generated from source -- is consistent and correct.
For that purpose, preferring the significantly more common form is useful.

The harm in disallowing a very rare spelling is that it may read as a false negative to someone who has that spelling and is using it intentionally. This seems similar to the problem with obscure proper nouns (e.g. "Nome"). I think this is acceptable, but, of course, I'm the one who proposed the check.

I don't know if this has any bearing on the issue, but, having now seen the distinction above between a person and a machine or mechanism, I would strongly discourage that usage if I encountered it in code reviews or a similar context. It's sufficiently obscure that we cannot expect a reader to know it, so it's better to find another way to make the distribution.

PR #2858, opened 2023-05-17
Add spelling correction for siver->silver/sliver/diver

cfi-gb:
See e.g. https://grep.app/search?q=siver&words=true

Note: There are names like "Kay Siver" but also various typos like e.g. the ones below so hope it is correct to add it like this.

https://github.com/heuermh/lick/blob/6b9b57a935c22d7c499ca5a9f926ef378f11d58d/lick/interval/Intervals.ck#L1148-L1149C14

https://github.com/astefanutti/kms-glsl/blob/f385a052b3a1495108b0e3e2702d446b00e2ef14/examples/subdivided_grid_truchet_parttern.glsl#L199

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -35158,6 +35158,7 @@ siutable->suitable
 siute->suite
 sive->sieve, save,
 sived->sieved, saved,
+siver->silver, sliver, diver,
 sives->sieves, saves,
 sivible->visible
 siving->sieving, saving,


DimitriPapadopoulos:
Don't worry, any typo could be a name (of a person, of a place, of a fictive character or place). We wouldn't have a dictionary of _possible_ typos if we were to take that into account.

Perhaps add `diver`?

cfi-gb:
> Don't worry, any typo could be a name (of a person, of a place, of a fictive character or place). We wouldn't have a dictionary of _possible_ typos if we were to take that into account.
> 
> Perhaps add `diver`?

Thanks a lot for the additional opinion about the names and similar, sounds absolutely reasonable to handle it like this :+1: 

I have also added `diver` as well, sounds absolutely valid.

PR #2862, opened 2023-05-25
While lambda is prevalent, Unicode uses lamda

### LINKED ISSUE #2860: Just a variant
https://github.com/codespell-project/codespell/blob/0f5c8fd59545fe2423dc03514e19cedad7623b8c/codespell_lib/data/dictionary.txt#L22488
Unicode uses the (Modern Greek-based) spelling "lamda" in character names, instead of "lambda", due to "the pre-existing names in ISO 8859-7, as well as preferences expressed by the Greek National Body".
https://en.wikipedia.org/wiki/Lambda#Character_encodings

DimitriPapadopoulos:
* [Greek and Coptic](https://www.unicode.org/charts/PDF/U0370.pdf)
* [Mathematical Alphanumeric Symbols](https://www.unicode.org/charts/PDF/U1D400.pdf)

The only exception in Unicode is **ƛ**, “LATIN SMALL LETTER LAMBDA WITH STROKE” in the [Latin Extended-B](https://www.unicode.org/charts/PDF/U0180.pdf) chart, for historical reasons.

Fixes #2860.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -24430,8 +24430,6 @@ laguagues->languages
 laguange->language
 laguanges->languages
 laiter->later
-lamda->lambda
-lamdas->lambdas
 lanaguage->language
 lanaguge->language
 lanaguges->languages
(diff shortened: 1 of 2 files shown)


skangas (approved):
LGTM.

PR #2865, opened 2023-06-01
move 'astroid' to code dictionary

robin-wayve:
`astroid` is a [python package](https://pypi.org/project/astroid/) which powers the commonly used linter `pylint`.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -3630,7 +3630,6 @@ astiimation->estimation
 astrix->asterisk
 astrixes->asterisks
 astrixs->asterisks
-astroid->asteroid
 asudo->sudo
 asume->assume
 asumed->assumed
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -2,6 +2,7 @@ agrv->argv
 alloced->allocated
 amin->main
 apoint->appoint
+astroid->asteroid
 atend->attend
 atending->attending
 bre->be, brie,


robin-wayve:
I disagree, but OK.

DimitriPapadopoulos:
Oh, I am not the maintainer, so you can challenge my comment.

However, note that codespell relies on a list of _possible_ typos. The very name of this package is a misspelling on purpose, a wordplay, like for other software such as Jupyter. If we were to take into account all location or product names, persons' names (fictional or not), trademarks, we might end up emptying our dictionaries. See for example https://github.com/codespell-project/codespell/issues/2811.  We need to maintain a balance, which could consider the ratio between false positives and false negatives, and not necessarily a ratio of 1 as the boundary.

robin-wayve:
My understanding that this is exactly the purpose of the `code` dictionary, and a brief search of github at least confirms my suspicion that the majority of occurrences are the python package https://github.com/search?q=astroid&type=code

DimitriPapadopoulos:
Exactly, but we need a balance between the notoriety and the frequency of the name involved in the false positives, and the frequency of false negatives if we remove the misspelling – and again I would recommend that the boundary is not a frequency ratio of 1.

I have no hard figures to back my impression that astroid is not frequent or notorious enough to move it to the code dictionary, unlike Jupyter, but you don't have hard figures either. I really don't know. You could re-open this PR and let the maintainer decide, time permitting. I just don't feel like merging myself.

robin-wayve:
@peternewman can re-open depending on what he makes of it.

Issue #2871, opened 2023-06-06
Issues with leading escape sequences in strings


cryptomilk:
codespell 2.2.4

As codespell checks strings and if a string looks like the following:

```
f.write("\tread only = no\n")
```

It complains that `tread` is incorrect as it doesn't detect the `t` as part of the escape sequence.

DimitriPapadopoulos:
See https://github.com/codespell-project/codespell/issues/1774.

Issue #2809, opened 2023-03-25
Ignore escape sequences


gforcada:
Hi! 👋🏾 

Today I found a string like this: `"{}\n\n!=\n\n{}\n\nIN {}({})"`

Notice the `\nIN ` that's what codespell was reporting to me:

```
Products/PortalTransforms/tests/test_transforms.py:103: nIN ==> inn, min, bin, nine
```

Would it be possible to detect this escape sequences, in the example above `\n` 🤔 

🍀

gforcada:
Sorry, duplicated of #1774

Issue #2508, opened 2022-09-29
Ignore new line character when checking for a word


jooola:
I have a print statement that starts with a new line char and some text. Code spell is failing as it includes the `n` from `\n` in the next word:

```
print("\nWe have ...\n")
```

Will suggest `nWe ==> new`

peternewman:
Agreed, but it's a duplicate of #1774

PR #2875, opened 2023-06-14
Ignore 'misspellings' due to string escapes

### LINKED ISSUE #17: codespell ignores C/C++ tabs ('\t') in strings
I found out, that codespell gives wrong answers in case of a tab is used within a C/C++ string. For example:
test.c:

``` c++
int main()
{
    std::cout<< "\tTest\n";
}
```

dict.txt:

```
ttest->test
```

Scanning with test.c with codespell gives a wrong result:

```
$ ../codespell.py dict.txt test.c 
test.c:3: tTest  ==> test
```

### LINKED ISSUE #1774: Escaped characters in strings not ignored in 2.0 version
Hello! I use the latest version of codespell in CI. Yesterday everything was ok, but the new release gives errors about escaped characters.

A minimal reproducible example is simple:
```
cat hello.c
#include <stdio.h>

int main() {
  printf("CodeSpell do not ignore escaped chars\nin strings.");
  return 0;
}
```

codespell==1.17.1 sees no problem in that file.

codespell==2.0.0 gives this message:

```
hello.c:4: nin ==> inn, min, bin, nine
```

Zac-HD:
This issue has been reported _many_ times - #17, https://github.com/codespell-project/codespell/issues/111,  https://github.com/codespell-project/codespell/issues/233#issuecomment-451950067, https://github.com/codespell-project/codespell/issues/1421, https://github.com/codespell-project/codespell/issues/1774, https://github.com/codespell-project/codespell/issues/2508, https://github.com/codespell-project/codespell/issues/2809, and https://github.com/codespell-project/codespell/issues/2871 at least.  It would also let me enforce `codespell` at work, which would be nice!

Design:  in order to make this super-fast, we do an O(1) check _of each detected misspelling_, with no new configuration or dictionary entries.  If the match is preceeded by a backslash and begins with a letter which (with backslash) constitutes a common escape string, _and_ the remainder of the word is not a misspelling, we skip it; otherwise we report exactly as we would now, with no special handling for the escapes.  It's a minimalist change meant to address a common complaint, accepting a very few missed alarms in exchange for lower complexity for both maintainers and users.

Fixes #17, fixes #1774, and closes #174.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -897,6 +897,19 @@ def parse_file(
             word = match.group()
             lword = word.lower()
             if lword in misspellings:
+                # Sometimes we find a 'misspelling' which is actually a valid word
+                # preceeded by a string escape sequence.  Ignore such cases as
+                # they're usually false alarms; see issue #17 among others.
+                char_before_idx = match.start() - 1
+                if (
+                    char_before_idx >= 0
+                    and line[char_before_idx] == "\\"
+                    # bell, backspace, formfeed, newline, carriage-return, tab, vtab.
+                    and word.startswith(("a", "b", "f", "n", "r", "t", "v"))
+                    and lword[1:] not in misspellings
+                ):
+                    continue
+
                 context_shown = False
                 fix = misspellings[lword].fix
                 fixword = fix_case(word, misspellings[lword].data)
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -898,7 +898,7 @@ def parse_file(
             lword = word.lower()
             if lword in misspellings:
                 # Sometimes we find a 'misspelling' which is actually a valid word
-                # preceeded by a string escape sequence.  Ignore such cases as
+                # preceded by a string escape sequence.  Ignore such cases as
                 # they're usually false alarms; see issue #17 among others.
                 char_before_idx = match.start() - 1
                 if (
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -98,6 +98,9 @@ def test_basic(
         f.write("this is a test file\n")
     assert cs.main(fname) == 0, "good"
     with fname.open("a") as f:
+        f.write("var = '\\nDoes not error on newline'\n")
+    assert cs.main(fname) == 0, "with string escape"
+    with fname.open("a") as f:
         f.write("abandonned\n")
     assert cs.main(fname) == 1, "bad"
     with fname.open("a") as f:


DimitriPapadopoulos:
This change would be a change of paradigm: _codespell_ is currently file type-agnostic. This change introduces a feature that is specific to files types that that support `\` as an escape character, without even relying on the current regexp mechanism. That said, such file formats are common enough that we can give this a thought.

Also, this feature cannot be disabled, it is not optional.

codespell_lib/_codespell.py
```diff
@@ -897,6 +897,19 @@ def parse_file(
             word = match.group()
             lword = word.lower()
             if lword in misspellings:
+                # Sometimes we find a 'misspelling' which is actually a valid word
+                # preceeded by a string escape sequence.  Ignore such cases as
```
larsoner, line 901: Hah, `codespell` got you :)
```suggestion
                # preceded by a string escape sequence.  Ignore such cases as
```

larsoner (commented):
Agreed @DimitriPapadopoulos this might cause some misses in some rare circumstances (binary files maybe?) but 1) it should help with false alarms and 2) it seems like 99% of people will want it "on" so I don't think we need to add an option yet. I'm inclined to fix the test (will commit shortly) then merge

larsoner:
Okay let's give it a shot, thanks @Zac-HD !

larsoner:
In theory https://github.com/codespell-project/codespell/releases/tag/v2.2.5 should cause a push to pip soon

codespell_lib/tests/test_basic.py
```diff
@@ -97,6 +97,9 @@ def test_basic(
     with fname.open("a") as f:
         f.write("this is a test file\n")
     assert cs.main(fname) == 0, "good"
+    with fname.open("a") as f:
+        f.write("var = '\\nDoes not error on newline'\n")
```
peternewman, line 101: This test relied on ndoes being in the dictionary, which I don't think it is.

I also found when adding a positive test that it doesn't match/trigger anyway because of the apostrophe at the start.

I've added some more thorough tests in #2879 .

codespell_lib/_codespell.py
```diff
@@ -897,6 +897,19 @@ def parse_file(
...
             if lword in misspellings:
+                # Sometimes we find a 'misspelling' which is actually a valid word
+                # preceded by a string escape sequence.  Ignore such cases as
+                # they're usually false alarms; see issue #17 among others.
+                char_before_idx = match.start() - 1
+                if (
+                    char_before_idx >= 0
+                    and line[char_before_idx] == "\\"
+                    # bell, backspace, formfeed, newline, carriage-return, tab, vtab.
+                    and word.startswith(("a", "b", "f", "n", "r", "t", "v"))
```
peternewman, line 908: This won't deal with anything where someone has put in e.g. \\ to render a backslash (or \\\\ etc) and the following (unrelated) word is a typo.

peternewman:
> This change would be a change of paradigm: _codespell_ is currently file type-agnostic. This change introduces a feature that is specific to files types that that support `\` as an escape character

> Agreed @DimitriPapadopoulos this might cause some misses in some rare circumstances (binary files maybe?)

RTF seems like it might possibly be the biggest/first hit (or maybe it's delimiters save it):
https://en.wikipedia.org/wiki/Rich_Text_Format#Code_syntax

> Also, this feature cannot be disabled, it is not optional.

> 2) it seems like 99% of people will want it "on" so I don't think we need to add an option yet.

Personally I think it would be nice to be able to turn it off, if someone finds an issue with it, or they just want to be more thorough. It wasn't very hard to find an edge case in #2880 .

PR #2889, opened 2023-06-20
Add spelling correction for 'prodive'

adrien-berchet:
https://github.com/search?q=prodive&type=code => 6.4K hits
https://github.com/search?q=prodived&type=code => 836 hits

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -29713,6 +29713,8 @@ procudes->produces, procures,
 procuding->producing, procuring,
 prodceding->proceeding
 prodecure->procedure
+prodive->provide
+prodived->provided
 producable->producible
 producables->producible
 produceds->produces, produced,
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

DimitriPapadopoulos (changes requested):
Perhaps add all verbal forms
```
prodives->provides
prodiving->providing
```
as well as:
```
prodiver->provider, pro diver,
prodivers->providers, pro divers,

adrien-berchet:
I added several forms.
For the `pro diving` forms I'm not sure it is relevant because it is a very specific context and it will make codespell to switch to multiple choice mode which, IMO, is more cumbersome to use. But I will add it if you think it's better.

DimitriPapadopoulos (approved):
I followed the links to GitHub and all typos in source code seem to unequivocally need to be fixed to `provide`, never to `pro dive`. Let's give it a try like this.

adrien-berchet:
Ok, cool. I let you merge then.
Thanks for the review!

Issue #2877, opened 2023-06-14
Dev Container


korverdev:
I was curious if the maintainers of this project have given any thought to using dev containers to simplify getting a development environment setup. I've been using this for a couple of projects I work on and the experience has been great, especially for getting new contributors set up quickly. I'd be happy to work on this, but I wanted to gauge interest before putting this together.

DimitriPapadopoulos:
What would a dev container look like? Right now, you “just” need to:
* install Python,
* run `pip install -e ".[dev]"`.

Casey-Kiewit:
@DimitriPapadopoulos That is likely. Codespell doesn't have many requirements, so it would be a relatively simple config file. The nice thing with dev containers is the integration with VSCode and a few other editors that allows you to setup a consistent, containerized development environment in one button click. It's definitely not a need, but I've found them very helpful when juggling multiple development environments between projects.

DimitriPapadopoulos:
I must admit my ignorance, I don't know what a “dev container” for IDEs looks like. Is the expected format standard?

Should it be maintained in this repository, or perhaps a new repository under https://github.com/codespell-project? How do other projects manage that?

DimitriPapadopoulos:
So this would just involve adding a small **`.devcontainer/devcontainer.json`** file, wouldn't it?

@larsoner Would you agree to add such a file?

Casey-Kiewit:
That is correct. Unless you actively use dev containers, it'll just mean one more config file in the repo and wouldn't impact current local development setups. Wanted to run this by you first since this is a somewhat opinionated technology, but if y'all are open to this, I'd be happy to set this up. I find it extremely useful as a sort of venv alternative.

larsoner:
Having one more file to maintain has a cost, but if it can automatically pull in deps and such (e.g., using `pip install .[dev]` or whatever inside the container) it sounds painless enough. So +1 from me

PR #2903, opened 2023-06-26
Dev Container

korverdev:
Adds a dev container config to allow for easy development environment setup and support for GitHub Codespaces.

Resolves https://github.com/codespell-project/codespell/issues/2877.


.devcontainer/post_create.sh
```diff
@@ -0,0 +1,12 @@
...
+sudo apt-get update
+sudo apt-get install -y libaspell-dev
+
+pip install --upgrade \
+    aspell-python-py3 \
+    pip \
+    setuptools \
+    setuptools_scm \
+    wheel
+pip install -e '.[dev]'
```
DimitriPapadopoulos, line 12: Out of curiosity, why a dev install? So that end-users can modify the dictionary or even the code locally?

korverdev: Yup. That's exactly right. The goal is to give you a single-click development environment for a local copy of Codespell. For your context, your local Codespell folder is automatically mounted to the default directory which is the context where these commands are being run.

I'm looking to mirror [this development setup guide](https://github.com/codespell-project/codespell#development-setup) with the addition of support for all optional tests.

korverdev:
Let's hold off with this until https://github.com/codespell-project/codespell/pull/2917 is addressed. This will let me leap straight to the latest Debian version.

.devcontainer/post_create.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+sudo apt-get update
+sudo apt-get install -y libaspell-dev
```
DimitriPapadopoulos, line 4: Does `libaspell-dev` install `aspell-en` automatically? Is `aspell-en` installed by default in image `mcr.microsoft.com/devcontainers/python:1-3-bullseye`?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "Codespell Development Environment",
+    // TODO: Switch to image 'mcr.microsoft.com/devcontainers/python:1-3' once
+    // https://github.com/codespell-project/codespell/issues/2897 has been
+    // resolved.
+    "image": "mcr.microsoft.com/devcontainers/python:1-3-bullseye",
+    "customizations": {
+        "vscode": {
+            "extensions": [
+                "github.vscode-github-actions",
```
DimitriPapadopoulos, line 10: Why is the above extension needed?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,17 @@
...
+                "ms-python.python",
+                "python.vscode-pylance"
```
DimitriPapadopoulos, line 12: According to https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance, this should be `ms-python.vscode-pylance`.

korverdev: Appreciate you pointing that out. I was apparently using the old name which I pulled from a previous config. Both still resolve correctly, but Microsoft has been in the process of renaming these for a while. Will update this.

.devcontainer/devcontainer.json
korverdev: None of these extensions are strictly needed, but they set default extensions for this dev container in VSCode and Codespaces. This is an official extension for working with GitHub's actions and I've found it pretty useful, but it can definitely be removed and installed manually if the user desires. Really just a matter of personal preference.

.devcontainer/post_create.sh
korverdev: That is correct. Here's the full list of what this command installs:
- aspell
- aspell-en
- dictionaries-common
- emacsen-common
- libaspell-dev
- libaspell15
- libtext-iconv-perl

This image won't be published anywhere, so if we wanted, we could likely skip most of these recommended dependencies and save some space.

korverdev:
@DimitriPapadopoulos Really appreciate the review and all your help with my last couple of issues. I'm flipping this out of draft since we're now able to use Debian bookworm. All your comments should be addressed or answered, so this should be ready for another look whenever you have the time.

.devcontainer/post_create.sh
DimitriPapadopoulos: I was just curious, we need `aspell` to run all tests locally.

.devcontainer/devcontainer.json
DimitriPapadopoulos: It's just that [GitHub.vscode-github-actions](https://github.blog/2023-03-28-announcing-the-github-actions-extension-for-vs-code/) seems orthogonal to the basic purpose of this container. GitHub Actions are necessary to modify codespell locally.

korverdev: None of these plugins are strictly needed to modify Codespell, but Codespell does make use of GitHub Actions and this is a recommended plugin for those files. One of the primary benefits of dev containers in my mind is to give a single-button development environment will all the plugins and settings for VSCode that are considered standard for a project already configured.

In my mind, for anyone working on Codespell's actions, this plugin will be appreciated and VSCode will prompt them to install it regardless. And for anyone who isn't, they probably won't even notice it's there.

DimitriPapadopoulos: Perhaps just change from
```
github.vscode-github-actions
```
to
```
GitHub.vscode-github-actions
```
to match the [GitHub.vscode-github-actions](https://marketplace.visualstudio.com/items?itemName=github.vscode-github-actions) page.

korverdev: I'm definitely not opposed to the change, but VSCode handles these identified weirdly. I pulled that identifier from the page in VSCode where it's still lower-cased and I've confirmed it's functioning as expected:
image

I suspect upper-case will likely work, so if you have a strong opinion there, I can definitely test the alternative.

DimitriPapadopoulos: Microsoft tools are usually case-insensitive. My idea was to stick to the documentation, but if the documentation itself is inconsistent, let's keep as it is. I prefer lowercase myself.

DimitriPapadopoulos (approved):

PR #2909, opened 2023-06-29
Add spelling corrections for upperace and lowerace.

cfi-gb:
See e.g. (Note: `grep.app` search calls have less hits):
- https://github.com/search?q=upperace&type=code
- https://github.com/search?q=lowerace&type=code
- https://grep.app/search?q=upperace
- https://grep.app/search?q=lowerace

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -23716,6 +23716,9 @@ losted->listed, lost, lasted,
 lotation->rotation, flotation,
 lotharingen->Lothringen
 lowd->load, low, loud,
+lowerace->lowercase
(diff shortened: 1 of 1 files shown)


luzpaz:
Would you mind adding the `-ed` suffix ? TIA!

cfi-gb:
> Would you mind adding the `-ed` suffix ? TIA!

Sore, absolutely. I have also added the `-ing` variants for these based on:
- https://www.thefreedictionary.com/uppercasing
- https://www.merriam-webster.com/dictionary/uppercase

and also added them for another already existing `upercase` correction.

luzpaz (approved):

PR #2919, opened 2023-07-04
Add typos for "general", "generate", "generic"

int-y1:


codespell_lib/data/dictionary.txt
```diff
@@ -19452,78 +19460,161 @@ gayity->gaiety
...
 geneation->generation
+geneations->generations
+geneator->generator
+geneators->generators
 geneic->generic, genetic,
 geneological->genealogical
 geneologies->genealogies
 geneology->genealogy
 generaates->generates
+generae->generate
```
DimitriPapadopoulos, line 19529: Perhaps `genera`, the plural of `genus`?
```
generae->generate, genera,
```
But then it's probably uncommon. Thoughts?

codespell_lib/data/dictionary.txt
```diff
@@ -19452,78 +19460,161 @@ gayity->gaiety
...
+genereally->generally
 genereate->generate
 genereated->generated
 genereates->generates
 genereating->generating
+genereation->generation
+genereations->generations
+genereator->generator
+genereators->generators
 genered->generated
```
DimitriPapadopoulos, line 19573: Just a suggestion, I'm letting you decide:
```
genered->generated, gendered,
```

codespell_lib/data/dictionary.txt
int-y1: I couldn't find an example of `generae->genera`, so I won't include this.

codespell_lib/data/dictionary.txt
int-y1: I found a few examples of `genered->gendered` and `genered->gender`, so I added both.

DimitriPapadopoulos (approved):

PR #2927, opened 2023-07-08
Add postifx->postfix to code dictionary

luzpaz:
[Postfix](https://www.postfix.org/) is a FOSS mail transfer agent

### DIFF
--- a/codespell_lib/data/dictionary_code.txt
+++ b/codespell_lib/data/dictionary_code.txt
@@ -50,6 +50,7 @@ objext->object
 od->of
 outputof->output of, output-of,
 packat->packet
+postifx->postfix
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

peternewman:
Sorry why did this go into the code dictionary? In what context is `postifx` a valid term that we don't want to correct (in code or otherwise)?

The code dictionary is to stop e.g. `postfix` being turned into `post fix`, its disabled by default so would only make that change when explicitly asked.

DimitriPapadopoulos:
Ah, that's true. @luzpaz Could you move `postifx->postfix` to `dictionary.txt`?

PR #2936, opened 2023-07-10
Add suffixes to typos that start with "l"

int-y1:


codespell_lib/data/dictionary.txt
```diff
@@ -25201,13 +25212,15 @@ liasing->liaising
...
+libararies->libraries
 libarary->library
 libaries->libraries
 libary->library
 libell->libel
 liberaries->libraries
 liberary->library
 liberoffice->libreoffice
+liberries->libraries
 liberry->library
```
DimitriPapadopoulos, line 25224: While I agree library is the most probable fix, I wonder whether this could `liberty` or even `bilberry`.

codespell_lib/data/dictionary.txt
```diff
@@ -25426,25 +25492,46 @@ litterally->literally
 litterals->literals
 litterate->literate
 litterature->literature
+litttle->little
 liturature->literature
 liuke->like
+livecycle->lifecycle
+livecycles->lifecycles
```
DimitriPapadopoulos, line 25499: Perhaps add `livetime`.

DimitriPapadopoulos (commented):
Just this single suggestion.

codespell_lib/data/dictionary.txt
int-y1: `liberty`: added
`bilberry`: not added, because this word is rare

DimitriPapadopoulos (approved):

PR #2937, opened 2023-07-10
Add suffixes to typos that start with "m"

int-y1:
`mange` and `manged` were in aspell

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25657,6 +25657,7 @@ maintaned->maintained
 maintanence->maintenance
 maintaner->maintainer
 maintaners->maintainers
+maintaning->maintaining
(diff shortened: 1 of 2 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -26530,6 +26570,7 @@ miminize->minimize
 miminized->minimized
 miminizes->minimizes
 miminizing->minimizing
+miminum->minimum
```
DimitriPapadopoulos, line 26573: Perhaps also `miminums->minimums` despite the formal plural is `minima`?

int-y1: I added both `miminums->minimums` and `mimina->minima`.

DimitriPapadopoulos (approved):

PR #2951, opened 2023-07-18
Add suffixes to typos from "sa" to "se"

int-y1:


codespell_lib/data/dictionary.txt
```diff
@@ -37781,24 +37888,56 @@ separatly->separately
 separato->separator
 separatos->separators
 separatring->separating
+separe->separate
 separed->separated
+separely->separately
+separes->separates
```
DimitriPapadopoulos, line 37894: Perhaps
```
separes->separates, spares,
```
I haven't looked if this typo exists out there, though.

int-y1: I couldn't find an example of `separe->spare`. I tried searching the next closest typo (`separingly->sparingly`), but this typo is too rare to be useful.

DimitriPapadopoulos (approved):

Issue #2960, opened 2023-07-23
Ruff checks failing on `master`


korverdev:
It appears something's changed with the latest version of ruff, [causing the server-side checks to fail with the following output](https://github.com/codespell-project/codespell/actions/runs/5639052570/job/15273804468):
```
codespell_lib/_codespell.py:1003:5: PLR0912 Too many branches (49 > 48)
```

I was able to recreate this locally. `v0.0.254` of the Pre-Commit hook passes all checks, but the latest version, `v0.0.280`, raises the same error.

korverdev:
Opened a separate issue to discuss a potential long-term solution for this and similar issues:
https://github.com/codespell-project/codespell/issues/2961

DimitriPapadopoulos:
Indeed ruff is still evolving rapidly resulting in new CI alerts without actual changes in our codebase. You just need to change:
```
max-branches = 48
```
to:
```
max-branches = 49
```

Or, alternatively, you can reorganise the code to remove branches, although I don't see anything wrong with multiple similar branches when parsing command line options.

PR #2962, opened 2023-07-23
Remove a branch from `main()` to pass Ruff CI check.

korverdev:
It appears that a recent update to Ruff has resulted in `master` failing the CI Ruff check. Resolves https://github.com/codespell-project/codespell/issues/2960.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -1063,10 +1063,8 @@ def main(*args: str) -> int:
         return EX_USAGE
     uri_ignore_words = parse_ignore_words_option(options.uri_ignore_words_list)
 
-    if options.dictionary:
-        dictionaries = options.dictionary
-    else:
-        dictionaries = ["-"]
+    dictionaries = options.dictionary if options.dictionary else ["-"]
+
     use_dictionaries = []
     for dictionary in dictionaries:
         if dictionary == "-":


DimitriPapadopoulos:
I would rather find a way to shut up this ruff warning in this function.

I don't find this specific warning very useful when parsing command line options.

DimitriPapadopoulos:
Besides, this is clearly a ruff shortcoming. You are **not** reducing the number of `if`s, just fooling ruff into not seeing one of them.

korverdev:
A few alternatives for this:
- [Silence the check for `main()`.](https://github.com/astral-sh/ruff/discussions/3442)
- [Increase the limit here.](https://github.com/codespell-project/codespell/blob/master/pyproject.toml#L144C2-L144C2)
- Disable the check entirely.

codespell_lib/_codespell.py
```diff
@@ -1063,10 +1063,8 @@ def main(*args: str) -> int:
         return EX_USAGE
     uri_ignore_words = parse_ignore_words_option(options.uri_ignore_words_list)
 
-    if options.dictionary:
-        dictionaries = options.dictionary
-    else:
-        dictionaries = ["-"]
+    dictionaries = options.dictionary if options.dictionary else ["-"]
```
larsoner, line 1066: This could be equivalently expressed more compactly like this but what you have is fine, too
```suggestion
    dictionaries = options.dictionary or ["-"]
```

larsoner:
Good enough I think, thanks @korverdev !

@DimitriPapadopoulos we might indeed need to quiet ruff a bit but at least for now we can avoid it :)

codespell_lib/_codespell.py
korverdev: Good to know! Haven't seen that syntax before.

PR #2969, opened 2023-07-25
Remove redundant linting CI workflows handled by pre-commit.ci

kurtmckee:
This PR removes the black and ruff CI workflows that are now run by pre-commit.ci.

One note: there's a `ruff-annotation` job in `codespell-private.yml`; I'm not sure if this should be removed as well.


kurtmckee:
It appears that the black and ruff workflows are configured as "required". If this PR merges, some repo settings may need to be modified.

larsoner:
I think we shouldn't need the `ruff-annotation` run either because `pre-commit.ci` should push fixes automatically for people

larsoner:
... and hopefully we can also get rid of the separate `mypy` run

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
@larsoner I cannot merge because of the missing GitHub workflows. And I would like you to have a look any way.

larsoner:
Thanks @kurtmckee !

PR #2973, opened 2023-07-27
Use pre-commit to sort the dictionaries

### LINKED ISSUE #2689: Add pre-commit.ci and use local hook to auto-run `make sort-dictionaries` on PRs
pre-commit.ci supports [`local` hooks](https://pre-commit.com/#repository-local-hooks) which could be used in this repo to safely run `make sort-dictionaries` on PRs.

It's a bit tedious to have to download the repo (I prefer https://github.dev for small PRs) just to get a terminal from which to run sorting manually.

kurtmckee:
This changes the sort order of specific characters (like `"_"`) but is cross-platform and can be fixed by pre-commit.ci.

Closes #2689

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 ---
-files: ^(.*\.(py|json|md|sh|yaml|yml|in|cfg|txt|rst|toml|precommit-toml))$
+files: ^(.*\.(py|json|md|sh|yaml|yml|in|cfg|txt|rst|toml|precommit-toml|wordlist))$
 exclude: ^(\.[^/]*cache/.*)$
 repos:
   - repo: https://github.com/executablebooks/mdformat
@@ -47,7 +47,7 @@ repos:
       - id: check-case-conflict
       - id: check-toml
       - id: file-contents-sorter
-        files: dictionary.*\.txt$
+        files: dictionary.*\.txt$|\.wordlist$
         args: [--ignore-case]
   - repo: https://github.com/adrienverge/yamllint.git
     rev: v1.32.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -46,6 +46,9 @@ repos:
       - id: fix-byte-order-marker
       - id: check-case-conflict
       - id: check-toml
+      - id: file-contents-sorter
+        files: dictionary.*\.txt$
+        args: [--ignore-case]
   - repo: https://github.com/adrienverge/yamllint.git
     rev: v1.32.0
     hooks:
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5,6 +5,9 @@
 3rt->3rd
 3st->3rd
 4rd->4th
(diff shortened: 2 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -46,6 +46,9 @@ repos:
       - id: fix-byte-order-marker
       - id: check-case-conflict
       - id: check-toml
+      - id: file-contents-sorter
+        files: dictionary.*\.txt$
```
DimitriPapadopoulos, line 50: May I ask you add the new files `*-additionnal.wordlist`? They live under `codespell_lib/tests/data` and are currently not sorted.

peternewman: Or even `*-additional.wordlist` as per #2976 !

peternewman:
> This changes the sort order of specific characters (like `"_"`) but is cross-platform and can be fixed by pre-commit.ci.

For anyone wondering why the hell sort and python don't return the same results, it seems to be around LC_ALL as per https://stackoverflow.com/questions/8776807/is-pythons-sort-function-the-same-as-linuxs-sort-with-lc-all-c

It seems a shame pre-commit doesn't offer an option to be consistent, but it might not be worth chasing for now.

As long as we're not setting ourselves up for issues when someone runs in another locale on their machine?

codespell_lib/data/dictionary.txt
```diff
@@ -11247,8 +11250,8 @@ cought->caught, cough, fought,
 coul->could
 could'nt->couldn't
 could't->couldn't
-couldent->couldn't
 coulden`t->couldn't
```
peternewman, line 11253:
```suggestion
coulden`t->couldn't
coulden't->couldn't
```

codespell_lib/data/dictionary.txt
```diff
@@ -47125,8 +47128,8 @@ woudl->would
 woudn't->wouldn't
 would'nt->wouldn't
 would't->wouldn't
-wouldent->wouldn't
 woulden`t->wouldn't
```
peternewman, line 47131:
```suggestion
woulden`t->wouldn't
woulden't->wouldn't
```

.pre-commit-config.yaml
```diff
@@ -46,6 +46,9 @@ repos:
...
+        args: [--ignore-case]
```
peternewman, line 51: We didn't have the ignore case option passed to sort before.

kurtmckee: It was. `SORT_ARGS`, defined in the Makefile and passed to `sort` when called, includes `-f`, which is the supremely non-obvious equivalent of sort's `--ignore-case` argument.

codespell_lib/data/dictionary.txt
kurtmckee: Dictionary changes seem outside of the scope of this PR. I can introduce this change if needed, but I would prefer to keep this PR tightly scoped.

DimitriPapadopoulos (approved):

larsoner:
Thanks @kurtmckee !

PR #2976, opened 2023-07-27
 Fix the name of the extra word lists we load

peternewman:
I'm a bit unclear why this fix didn't trip it up TBH:
https://github.com/codespell-project/codespell/tree/peternewman-test-filename-checking

Ideas welcome, and would be sort of good to solve before we just merge this...


DimitriPapadopoulos:
Indeed:
```
$ cd codespell_lib/tests/data
$ 
$ ls
en_GB-additionnal.wordlist
en_US-additionnal.wordlist
$ 
$ codespell -f
$ 
```

DimitriPapadopoulos:
It's the dash, and I think it's expected:
```
$ cat > additionnal.txt
additionnal
$ 
$ cat > list-additionnal.txt
list-additionnal    
$ 
$ codespell  -f
./additionnal.txt: additionnal ==> additional
./additionnal.txt:1: additionnal ==> additional
$ 
```

peternewman:
> It's the dash, and I think it's expected:

Yeah I'd expect the hyphen to, for e.g. semi-skimmed, but I tried to change it to another underscore hear and it still didn't catch it:
https://github.com/codespell-project/codespell/tree/peternewman-test-filename-checking

Which might be an outstanding issue around snake_case checking...

DimitriPapadopoulos:
Indeed, **`-`** is part of the default word definition, but not **`_`**:
https://github.com/codespell-project/codespell/blob/47064e05418df07f54dacbfdd7e3a99088866afe/codespell_lib/_codespell.py#L31

It looks like **`_`** is part of **`\w`**:
```python
>>> import re
>>> 
>>> word_regex_def = "[\\w\\-'’`]+"
>>> word_regex = re.compile(word_regex_def)
>>> 
>>> word_regex.findall("list-additional")
['list-additional']
>>> 
>>> word_regex.findall("list_additional")
['list_additional']
>>> 
>>> word_regex.findall("list additional")
['list', 'additional']
>>> 
>>> word_regex.findall("list.additional")
['list', 'additional']
>>> 
```

DimitriPapadopoulos:
Is there an easy way to get **`\w`** except **`_`** in the non-[`ASCII`](https://docs.python.org/3/library/re.html#re.ASCII) case?

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
See #2979 for `snake_case` checking.

Issue #2981, opened 2023-07-28
Default word regex and grave accent (`)


DimitriPapadopoulos:
Why include the grave accent (**&#x0060;**) in the default word regex?

https://github.com/codespell-project/codespell/blob/2eeb8b992e7fec4940a12bdac5e8837f0ffd6c98/codespell_lib/_codespell.py#L31

This messes up Markdown checking, just for the sake of catching these two typos:
```
coulden`t->couldn't
woulden`t->wouldn't
```
These two fixes don't really make sense, we don't even catch the same typos with the apostrophe (**&#x0027;**) or acute accent (**&#x00B4;**), :
```
coulden't->couldn't
coulden´t->couldn't
woulden't->wouldn't
woulden´t->wouldn't
```

luzpaz:
CC @larsoner do you mind weighing in ?

PR #2985, opened 2023-07-28
Generate alternative typos with a translation table

### LINKED ISSUE #2984: Misspellings with an apostrophe
Would it make sense to duplicate all misspellings that contain an apostrophe (**'**), replacing the apostrophe (**'**) with an acute accent (**’**) in both the misspelling and fix(es)?

That way, we wouldn't have to duplicate the typos in the dictionary itself:
```
coulden't->couldn't
coulden´t->couldn't
woulden't->wouldn't
woulden´t->wouldn't
```

DimitriPapadopoulos:
This way we can catch misspellings with alternative characters, typically an acute accent (**´**) instead of an apostrophe (**'**). In this case, the alternative character is a valid character and will be used both in the misspelling and the fix(es).

The above is different from detecting Unicode phishing, where some characters like `A` are intentionally, or not, replaced by lookalikes such as `A`, `Α`,  `А`,  `ᗅ`, `ᴀ`,  `Ａ`. In this case, the alternative character is invalid and should be replaced by its valid counterpart in the fix.

The latter use case is deferred to a different merge request.

Fixes #2984.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -36,6 +36,9 @@ uri_regex_def = (
     "(\\b(?:https?|[ts]?ftp|file|git|smb)://[^\\s]+(?=$|\\s)|"
     "\\b[\\w.%+-]+@[\\w.-]+\\b)"
 )
+# Pass all misspellings through this translation table to generate
+# alternative misspellings and fixes.
+alt_chars = (("'", "’"),)
 encodings = ("utf-8", "iso-8859-1")
 USAGE = """
 \t%prog [OPTIONS] [file1 file2 ... fileN]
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
@larsoner This is a proof of concept, I will add tests if there is interest (typically to detect and avoid replaced characters in dictionaries).

peternewman:
Seems much more sensible than my suggestion!

Microsoft "smart" quotes too?

Will also want some checks that we don't duplicate in the dictionary with different variants. Likewise if someone ignores with a different escape it will need to still match and remove it.

Presumably correction will have our deemed acceptable one?

How does this interact with the escape stuff someone did a little while ago?

DimitriPapadopoulos:
By adding alternatives while reading the dictionaries, it really works as additional entries in the dictionary. Hence, we cannot, and I suggest we should not, ignore alternatives:
1. There might be niche cases where we want to ignore only the specified alternative.
2. In most cases, only a single alternative will be present in the dictionary.
3. For the above reasons, it is not worth the additional complexity.

Duplicates will be avoided by checking all target characters are missing from the dictionaries. For example, all misspellings of acute accent (**´**) would have to be removed from the dictionaries, or the acute accent replaced by an apostrophe.

DimitriPapadopoulos:
@peternewman Which escape stuff are you referring too?

I have not added smart single quotes because they are not part of the default word regex, which is probably used in the vast majority of cases:
https://github.com/codespell-project/codespell/blob/2eeb8b992e7fec4940a12bdac5e8837f0ffd6c98/codespell_lib/_codespell.py#L31

However, I could add smart single quotes to both the default word regex and my alternative to the apostrophe.

Similarly, I have left out the grave accent because of #2981 and Markdown.

larsoner:
Can you modify some clear/informative test that fails on `master` but passes on this PR? It will both act as a regression test and help reviewers understand the use case concretely...

DimitriPapadopoulos:
The Windows error is related to encoding, obviously: According to the Python documentation, when writing:
> _encoding_ is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever [`locale.getencoding()`](https://docs.python.org/3/library/locale.html#locale.getencoding) returns),

Need to find how the file is then decoded.

DimitriPapadopoulos:
When reading files, the `FlleOpener` tries these encodings:
https://github.com/codespell-project/codespell/blob/adf6ff6f2744dd75b08993fb085da67e58cbb50e/codespell_lib/_codespell.py#L39

I will try to explicitly use the `utf-8` encoding when writing, to match the first encoding `utf-8` tried while reading.

DimitriPapadopoulos:
@larsoner Can you have a look? Obviously, reading dictionaries will be slower, as I have added a function to process and sometimes duplicate dictionary lines.

PR #2986, opened 2023-07-28
Simpler dictionary parsing

DimitriPapadopoulos:
I find the code easier to read, and it's even (very slightly) faster under Python 3.11:

```
$ git checkout misspellings_array
Switched to branch 'misspellings_array'
Your branch is up to date with 'origin/misspellings_array'.
$ 
$ time for run in {1..500}; do python3.11 /home/username/.local/bin/codespell /tmp ; done

real	0m40,160s
user	0m33,300s
sys	0m6,875s
$ 
$ 
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ 
$ time for run in {1..500}; do python3.11 /home/username/.local/bin/codespell /tmp ; done

real	0m41,794s
user	0m34,998s
sys	0m6,799s
$ 
```

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -637,19 +637,14 @@ def build_dict(
             if key in ignore_words:
                 continue
             data = data.strip()
-            fix = data.rfind(",")
 
-            if fix < 0:
+            if "," in data:
+                fix = False
+                data, reason = data.rsplit(",", 1)
+                reason = reason.lstrip()
+            else:
                 fix = True
                 reason = ""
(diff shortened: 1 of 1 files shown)


larsoner:
Admittedly I was skeptical that the `in` followed by another search (if present) would be fastest but I checked some other options (including `master`) with:
```
python -m timeit "import codespell_lib; codespell_lib.main('./codespell')"
```
and indeed the implementation here seemed fastest! Thanks @DimitriPapadopoulos !

PR #2991, opened 2023-07-31
Add suffixes to typos that start with "ap"

int-y1:


codespell_lib/data/dictionary.txt
```diff
@@ -3696,7 +3696,15 @@ aotomatically->automatically
 aotomation->automation
 aovid->avoid
 apach->apache
+apapt->adapt
```
DimitriPapadopoulos, line 3699: Could it be `apart` in some cases?

codespell_lib/data/dictionary.txt
```diff
@@ -3776,15 +3813,26 @@ apostrafies->apostrophes
...
+appache->apache
 appar->appear
+apparance->appearance
+apparances->appearances
 apparant->apparent
 apparantly->apparently
+appareance->appearance
+appareances->appearances
 appared->appeared
+apparen->apparent
```
DimitriPapadopoulos, line 3825: Perhaps `apparel` too? **L** and **N** are close on Dvorak keyboards, for what it's worth.

codespell_lib/data/dictionary.txt
int-y1: I couldn't find `apapt->apart`.

codespell_lib/data/dictionary.txt
int-y1: I couldn't find `apparen->apparel`.

PR #3022, opened 2023-08-09
Get rid of typos with apostrophe replaced by semicolon

### LINKED ISSUE #2982: Default word regex and semicolon (;)
The **default** word regex does not include the semicolon (**;**) and I agree with that:

https://github.com/codespell-project/codespell/blob/2eeb8b992e7fec4940a12bdac5e8837f0ffd6c98/codespell_lib/_codespell.py#L31

But then, why maintain a long list of typos that contain a semicolon (**;**), since the default word regex won't be able to match such typos any way?
```
aren;t->aren't
arent;->aren't
can;t->can't
cant;->can't
couldn;t->couldn't
couldnt;->couldn't
didn;t->didn't
didnt;->didn't
doesn;t->doesn't
doesnt;->doesn't
dosen;t->doesn't
dosent;->doesn't
dosn;t->doesn't
get;s->gets
hasn;t->hasn't
haven;t->haven't
i;ll->I'll
isn;t->isn't
isnt;->isn't
nightfa;;->nightfall
shouldn;t->shouldn't
shouldnt;->shouldn't
that;s->that's
thats;->that's
wan;t->want, wasn't,
want;s->wants
was;t->wasn't
wasn;t->wasn't
wasnt;->wasn't
which;s->which's
won;t->won't
wouldn;t->wouldn't
wouldnt;->wouldn't
```

Is it OK to discard those typos, or at least move them to a `dictionary_no_match.txt` archive file?

DimitriPapadopoulos:
The default regex does not match "words" with semicolons in them anyway.

Fixes #2982.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -4866,10 +4866,8 @@ are'nt->aren't
 aready->already
 areea->area
 aren's->aren't
-aren;t->aren't
 arent'->aren't
 arent->aren't
-arent;->aren't
 areodynamics->aerodynamics
 argement->argument
 argements->arguments
@@ -8821,7 +8819,6 @@ campatibilities->compatibilities
 campatibility->compatibility
 campatible->compatible
 campatibly->compatibly
-can;t->can't
 canabel->cannibal
 canabels->cannibals
 canabelyse->cannibalise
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
@larsoner Could you have a look at this merge request and its associated issue https://github.com/codespell-project/codespell/issues/2982?

larsoner:
> Generally speaking, would it make sense to add a test to detect (and discard) typos that don't match the default word regex?

I would say "yes" and I think you've done that here so in it goes, thanks @DimitriPapadopoulos !

PR #3034, opened 2023-08-15
Add suffixes to typos from "dc" to "den"

int-y1:


codespell_lib/data/dictionary_rare.txt
```diff
@@ -69,7 +69,6 @@ deffer->differ, defer,
 degrate->degrade
 degrates->degrades
 dependant->dependent
-derails->details
```
luzpaz, line 72: This is here for a reason. 'r' can be a typo of 't'.

codespell_lib/data/dictionary.txt
```diff
@@ -15884,7 +15898,16 @@ deaults->defaults
 deauthenication->deauthentication
 debain->Debian
 debateable->debatable
+debbug->debug
```
DimitriPapadopoulos, line 15901: Perhaps `bedbug` too? Same for plural.

codespell_lib/data/dictionary.txt
```diff
@@ -15960,7 +15997,10 @@ decerning->discerning
 decerns->discerns
 deciaml->decimal
 deciamls->decimals
-decices->decides
+decice->device, decide,
```
DimitriPapadopoulos, line 16000: Perhaps `decile` too, although **C** and **L** are far apart on most keyboards? Plural too.

codespell_lib/data/dictionary.txt
```diff
@@ -16638,11 +16773,17 @@ denomitators->denominators
...
+dentically->identically
+dentification->identification
 dentified->identified
 dentifier->identifier
 dentifiers->identifiers
 dentifies->identifies
+dentify->identify
 dentifying->identifying
+dentities->identities, densities, entities,
+dentity->identity, density, entity,
```
DimitriPapadopoulos, line 16786: Perhaps `dentistry` too?

codespell_lib/data/dictionary_rare.txt
DimitriPapadopoulos: Indeed. Is this incompatible with `defails->details, derails,` in the main dictionary?

Also, I'm wondering why the singular was missing.

int-y1: This entry doesn't conflict with `defails->details, derails,` in the main dictionary.
I deleted this entry because most uses of `derails` on GitHub were correct.

What should I do with this entry?

codespell_lib/data/dictionary.txt
int-y1: `decice->decile` is rare, but exists. I think this is caused by accidentally using `c` twice in a row.

codespell_lib/data/dictionary.txt
int-y1: I couldn't find `dentity->dentistry`

DimitriPapadopoulos (approved):

codespell_lib/data/dictionary_rare.txt
DimitriPapadopoulos: @luzpaz What do you think? Should we keep `derails->details`, and if so add `derail->detail`?

int-y1: I added this entry back.

DimitriPapadopoulos:
@larsoner There"s a bug in the codecov/project CI job. Can you bypass the failure and merge for us?

PR #484, opened 2018-05-06
Letters 'j, k, and l' from #436

luzpaz:
Continuing to separate ~#428~ #436 in to separate PRs for each letter. Also removed/corrected some entries that were not relevant/precise and augmented others. This PR has been sorted as well.

PR #3038, opened 2023-08-20
Remove the valid expression "keep-alives"

fxlb:
The IETF RFCs currently contain:
209 keep-alives
469 keep-alive

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -29145,7 +29145,6 @@ karismatically->charismatically
 kazakstan->Kazakhstan
 keboard->keyboard
 keboards->keyboards
-keep-alives->keep-alive
 keept->kept
 keesh->quiche
 kenel->kernel, kennel,


DimitriPapadopoulos:
Originates in https://github.com/codespell-project/codespell/pull/436 / https://github.com/codespell-project/codespell/pull/484.

The fact that it is used in RFCs doesn't make it proper English. It can be used as a shorthand in some cases, but then ignore the codespell suggestion. I'll leave it to native speakers.

DimitriPapadopoulos:
Note that aspell itself doesn't like the word:
```
$ cat > file
keep-alives
$ 
$ aspell check -l en --size 90 file
 


keep-alives








                                                                                                                                    
1) alive                                                          6) elves
2) lives                                                          7) alcoves
3) olives                                                         8) Alice's
4) Olive's                                                        9) laves
5) olive's                                                        0) Alice
i) Ignore                                                         I) Ignore all
r) Replace                                                        R) Replace all
a) Add                                                            l) Add Lower
b) Abort                                                          x) Exit
                                                                                                                                    
? 
$ 
```
You could raise an issue with the [SCOWL wordlist](https://github.com/en-wl/wordlist) instead, which appears to be the _de facto_ reference for open source English spell checkers.

PR #3056, opened 2023-08-31
Add more typos for "initialize" and similar

int-y1:


DimitriPapadopoulos:
Would it be worth adding typos of `initiate`, where applicable?

int-y1:
I avoided `initiate` because this pr is getting too big.

DimitriPapadopoulos (approved):

PR #3062, opened 2023-09-02
Add refinement to 'draing' typo

luzpaz:
Added 'draining' as alternative solution. Example in the wild:  
https://github.com/gphoto/libgphoto2/blob/0b93f2d0d5305e941017d2d7cc6e04952261e126/camlibs/ptp2/library.c#L5662


codespell_lib/data/dictionary.txt
```diff
@@ -20177,7 +20177,7 @@ draconain->draconian
 dragable->draggable
 draged->dragged
 draging->dragging
-draing->drawing
+draing->drawing, draining,
```
DimitriPapadopoulos, line 20180: Indeed, but in this example I think the intended word was `drain`, not `draining`.
```suggestion
draing->drawing, drain, draining,
```

luzpaz: Screenshot_20230902_104053

Can't seem to commit via GH interface

luzpaz: Added manually in adb7ec0

DimitriPapadopoulos: Yes, the GitHub interface is broken, it cannot modify files > 1 MB.

Perhaps it's time to split dictionary.txt into multiple files.

DimitriPapadopoulos (approved):

PR #3067, opened 2023-09-05
Add spelling correction for "exis".

cfi-gb:
Did that typo myself recently so adding it. See also e.g. the following below from https://grep.app/search?q=exis&words=true:

> spec "err log should exist with null in cluster mode"

or:

> print("File ",weigths_filename," does not exis. Retraining... ")

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -22938,6 +22938,7 @@ exipration->expiration
 exipre->expire
 exipred->expired
 exipres->expires
+exis->exist, exists, exit, exits, axis, lexis, exes,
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

DimitriPapadopoulos (commented):
Perhaps add `lexis`, and even `exes`, plural of `ex`.

cfi-gb:
> Perhaps add `lexis`, and even `exes`, plural of `ex`.

Absolutely makes sense, thanks for the suggestion. :+1: These have been added now.

DimitriPapadopoulos (approved):

Issue #3088, opened 2023-09-20
Codespell pre-commit hooks totally failing following the release of setuptools-scm==8.0.0


andrewmwilson:
Following the release of `setuptools-scm==8.0.0` around an hour ago, I believe all newly-installed codespell pre-commit hooks are failing with the following error.

```Traceback (most recent call last):
  File "/Users/a/.cache/pre-commit/repoczrniamn/py_env-python3.11/bin/codespell", line 5, in <module>
    from codespell_lib import _script_main
  File "/Users/a/.cache/pre-commit/repoczrniamn/py_env-python3.11/lib/python3.11/site-packages/codespell_lib/__init__.py", line 1, in <module>
    from ._codespell import _script_main, main
  File "/Users/a/.cache/pre-commit/repoczrniamn/py_env-python3.11/lib/python3.11/site-packages/codespell_lib/_codespell.py", line 29, in <module>
    from ._version import __version__ as VERSION  # noqa: N812
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/a/.cache/pre-commit/repoczrniamn/py_env-python3.11/lib/python3.11/site-packages/codespell_lib/_version.py", line 4
    __version__ : str = version : str = '0.1.dev1+g355e50e'
                                ^
SyntaxError: invalid syntax

andrewmwilson:
Looks like this is an issue with setuptools-scm, so codespell might not have to do much other than wait for a fix.

https://github.com/pypa/setuptools_scm/issues/905
https://github.com/pypa/setuptools_scm/issues/906

andrewmwilson:
Looks like this has been resolved upstream.

tomrijntjes:
Not so. Opened a PR to fix the immediate issue: https://github.com/codespell-project/codespell/pull/3100

PR #3100, opened 2023-09-28
fixing setuptool_scm dependency because the latest version is broken

tomrijntjes:
The upstream fix is not released yet. Pinning the version has been proposed as a workaround [here](https://github.com/pypa/setuptools_scm/issues/908#issue-1904797205).

Tested using `make check` and by invoking the pre-commit hook.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ repository = "https://github.com/codespell-project/codespell"
 
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = ["setuptools>=64", "setuptools_scm[toml]>=7.1.0"]
+requires = ["setuptools>=64", "setuptools_scm[toml]>=6.2, != 8.0.0"]
 
 [tool.setuptools_scm]
 write_to = "codespell_lib/_version.py"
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
How recent is this problem? I haven't noticed it yet.

pyproject.toml
```diff
@@ -65,7 +65,7 @@ repository = "https://github.com/codespell-project/codespell"
 
 [build-system]
 build-backend = "setuptools.build_meta"
-requires = ["setuptools>=64", "setuptools_scm[toml]>=6.2"]
+requires = ["setuptools>=64", "setuptools_scm[toml]>=7.1.0"]
```
DimitriPapadopoulos, line 68: Perhaps I have misunderstood the issue at hand, but:
```suggestion
requires = ["setuptools>=64", "setuptools_scm[toml]<=7.1.0"]
```

tomrijntjes: good one, let me test that. 

The problem has been open since at least september 20th, see this [issue](https://github.com/codespell-project/codespell/issues/3088).

tomrijntjes: Tested and works fine. Good suggestion, this is more precise than what I proposed.

larsoner (approved):

PR #3104, opened 2023-10-01
Update and rename _codespell.py to codespell.py

Vansh16aug:


larsoner:
Prefixing by `_` is intentional to indicate that everything inside that file is private from an API standpoint

Vansh16aug:
ooo okok now i understood

PR #3116, opened 2023-10-03
Add schematrion->schematron

AirQuick:
[`Schematron`](https://en.wikipedia.org/wiki/Schematron) is a schema language for XML.
I've found several instances of its typo, `Schematrion`, so thought it might be worth adding.

- xspec/xspec#1778
- https://github.com/apache/camel/blob/main/components/camel-schematron/src/main/docs/schematron-component.adoc#headers:~:text=The-,schematrion,-report%20body%20in
- https://stackoverflow.com/questions/38268457/schematrion-case-insensitive-sqfstringreplace#:~:text=for%20your%20Teams%3F-,Schematrion,-case%2Dinsensitive%20sqf

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -45990,6 +45990,7 @@ schedulier->scheduler
 schedulling->scheduling
 scheduluing->scheduling
 schem->scheme
+schematrio->schematron
 schematrion->schematron
 schemd->schemed
 schems->schemes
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):
I thought about suggesting an uppercase **`S`**, but there are many cases in programming where the intent is not to use an uppercase first letter:
```
schematrion->Schematron
```

PR #3126, opened 2023-10-08
fix: aesthetic(s) should be kept as is

### LINKED ISSUE #3124: aesthetic is used in the US
The [British to American English spelling dictionary](https://github.com/codespell-project/codespell/blob/master/codespell_lib/data/dictionary_en-GB_to_en-US.txt) "corrects" aesthetic to esthetic even though aesthetic is also used in the US. This should not be the case. The spelling "esthetic" on the other hand is used only in the US, though, so if the intention is for it to correct to "aesthetic" in British English, then this is an issue where this word has an additional spelling in American English and not British English, unlike the other words as far as I can see. Regardless, "aesthetic" should be allowed by the spell checker when writing in American English.

EDIT: From looking at the code, it appears like it only converts British English to American English, so I think these lines should just be removed. I am willing to open a PR to do so if the developers feel like this is an acceptable outcome.

SimonVerhoeven:
fixes #3124

### DIFF
--- a/codespell_lib/data/dictionary_en-GB_to_en-US.txt
+++ b/codespell_lib/data/dictionary_en-GB_to_en-US.txt
@@ -1,10 +1,8 @@
 acknowledgement->acknowledgment
 acknowledgements->acknowledgments
-aesthetic->esthetic
 aesthetically->esthetically
 aesthetician->esthetician
 aestheticians->estheticians
-aesthetics->esthetics
 aggrandise->aggrandize
 aggrandised->aggrandized
 aggrandisement->aggrandizement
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

codespell_lib/data/dictionary_en-GB_to_en-US.txt
```diff
@@ -1,10 +1,8 @@
 acknowledgement->acknowledgment
 acknowledgements->acknowledgments
-aesthetic->esthetic
 aesthetically->esthetically
```
DimitriPapadopoulos, line 3: Also remove:
```
aesthetically->esthetically
aesthetician->esthetician
aestheticians->estheticians
```

SimonVerhoeven: I was a bit on the fence, but sure.

DimitriPapadopoulos (approved):

PR #3129, opened 2023-10-08
Added a -F/--filenames-only flag and support.

rreeves-web:


wangsiqidahaoren:
Has this problem been solved?

PR #3162, opened 2023-10-18
Fix ruff alerts (currently) not caught by pre-commit

DimitriPapadopoulos:


DimitriPapadopoulos:
I cannot reproduce this error locally:
```
codespell_lib/_codespell.py:19:1: I001 Import block is un-sorted or un-formatted
```
That's one of the reasons I don'y like pre-commit. Too complex.

codespell_lib/_codespell.py
```diff
@@ -268,7 +270,7 @@ def open_with_internal(self, filename: str) -> Tuple[List[str], str]:
                     break
         else:
             msg = "Unknown encoding"
-            raise Exception(msg)
+            raise LookupError(msg)
```
DimitriPapadopoulos, line 273: @larsoner Any suggestion on how to build a file that cannot be interpreted as `utf-8` or `iso-8859-1`?

DimitriPapadopoulos: I suspect there is no illegal byte in `latin1`. I'll write all bytes between 0 and 255 into a file and check whether the file can be read.

DimitriPapadopoulos: It looks like reading any file with encoding `iso-8859-1` cannot fail, or at least will not throw an exception such as `UnicodeDecodeError`:
```python
>>> with open("test.bin", "wb") as f:
...     f.write(bytes(range(0, 256)))
... 
256
>>> with open("test.bin", encoding="latin1") as f:
...     f.readlines()
... 
['\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n', '\x0b\x0c\n', '\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !"#$%&\'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~\x7f\x80\x81\x82\x83\x84\x85\x86\x87\x88\x89\x8a\x8b\x8c\x8d\x8e\x8f\x90\x91\x92\x93\x94\x95\x96\x97\x98\x99\x9a\x9b\x9c\x9d\x9e\x9f\xa0¡¢£¤¥¦§¨©ª«¬\xad®¯°±²³´µ¶·¸¹º»¼½¾¿ÀÁÂÃÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖ×ØÙÚÛÜÝÞßàáâãäåæçèéêëìíîïðñòóôõö÷øùúûüýþÿ']
>>> 
```
Therefore, the only way to reach "Unknown encoding" would be to modify the encodings global variable, removing  `"iso-8859-1"`:
https://github.com/codespell-project/codespell/blob/4a122a459c80c0b62a4fca6a755c5098bdace3c5/codespell_lib/_codespell.py#L42

DimitriPapadopoulos: I have added `#pragma: no cover` to avoid the Codecov warning. Also refactored the code a little bit.

DimitriPapadopoulos:

larsoner:
Thanks @DimitriPapadopoulos

PR #3165, opened 2023-10-19
Create speech.py

abhuman:


luzpaz:
@DimitriPapadopoulos maybe using a different label called 'hacktoberfest-spam' because the current one makes this look like a legit hacktoberfest PR..?

luzpaz:
@abhuman reporting this to github

PR #3174, opened 2023-10-23
Add Gelma's typos that start with "e"

int-y1:


codespell_lib/data/dictionary.txt
```diff
@@ -23296,13 +23394,21 @@ execising->exercising
 execpt->except
 execption->exception
 execptions->exceptions
+exect->exact, expect, eject,
```
DimitriPapadopoulos, line 23397: Perhaps `elect`, `execute`, `exert` too?

Same with final `s` below.

int-y1: I'll add `exert`. I couldn't find the others.

DimitriPapadopoulos (approved):

PR #3178, opened 2023-10-24
Add alwats->always correction.

cfi-gb:
See e.g. https://grep.app/search?q=alwats&words=true


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3179, opened 2023-10-24
Add entries to rare dictionary

DimitriPapadopoulos:
Split off from #2266

This is a copycat of #2421 without the `whit` → `wit` suggestion, which can be addressed specifically in that repository after rebasing.


DimitriPapadopoulos:
@larsoner @luzpaz Could you have a look and merge this PR so that we can specifically address the circularity in dictionary tests after rebasing #2421?

PR #3190, opened 2023-10-30
 Fibonacci Series

Megghaanaa:

PR #3192, opened 2023-10-30
 Finding Largest Value In Each Tree Row

Megghaanaa:


luzpaz:
Hacktoberfest spam. Reported in #3190

Issue #3196, opened 2023-11-02
"Interaction" with pre-commit: ideally should skip the skipped files.


yarikoptic:
I ran into it before but forgot to report, and now didn't find open issue. Currently inspired by https://github.com/PrefectHQ/prefect/pull/10893 where running  `codespell` in CLI directly results in no hits, but `pre-commit` reports:

```
codespell................................................................Failed
- hook id: codespell
- exit code: 65

.github/ISSUE_TEMPLATE/5_maintenance_ticket.yaml:1: maintenence ==> maintenance
.github/workflows/python-tests.yaml:62: automaticlly ==> automatically
.github/workflows/python-tests.yaml:246: automaticlly ==> automatically
.github/ISSUE_TEMPLATE/2_ui_bug_report.yaml:69: Browers ==> Browsers
ui/package-lock.json:2999: OT ==> TO, OF, OR, NOT
ui/package-lock.json:8411: OT ==> TO, OF, OR, NOT
src/prefect/_vendor/fastapi/utils.py:131: unecessary ==> unnecessary
``` 

**Problems:** 

1. If I define files to  `skip` in codespell configuration (e.g. in `.codespellrc`), those files will **not** be skipped in `pre-commit run --all` since ( I guess) pre-commit would pass those paths explicitly to `codespell` invocation and thus it would not account for `skip`.

2. Also, complimentary, may be `codespell` should check in `.dotdirs` like `.github` etc?

DimitriPapadopoulos:
1. It looks like we do skip explicitly passed files that match the skip list:
   ```
   $ echo unecessary > foobar
   $ 
   $ codespell foobar
   foobar:1: unecessary ==> unnecessary
   $ 
   $ codespell -S foobar foobar
   $ 
   ```
   Could it be that `.codespellrc` is not taken into account when run by pre-commit?
2. Does option `-H`/`--check-hidden` help?

yarikoptic:
> 1. Could it be that `.codespellrc` is not taken into account when codespell is run by pre-commit?

I think it might be more of the case of directories listed in `skip`

```shell
❯ grep skip .codespellrc
skip = .git,*.pdf,*.svg,versioneer.py,package-lock.json,_vendor,*.css
❯ codespell
❯ codespell **/_vendor
src/prefect/_vendor/fastapi/utils.py:131: unecessary ==> unnecessary
```

> 2\. Does option `-H`/`--check-hidden` help?

[...]

PR #3198, opened 2023-11-03
Test we properly skip argument when relevant

DimitriPapadopoulos:


DimitriPapadopoulos:
@larsoner Can you have a look? This PR should fix #3196 reported by @yarikoptic.

codespell_lib/_codespell.py
```diff
@@ -1168,6 +1168,16 @@ def main(*args: str) -> int:
         if is_hidden(filename, options.check_hidden):
             continue
 
+        # skip argument if any piece of the path is in the skip list
```
DimitriPapadopoulos, line 1171: Perhaps I can make this a function to avoid increasing pylint acceptable metrics too much.

larsoner: sure, some `if _match_path_parts(...): continue` or so would help `main` be more readable

DimitriPapadopoulos: Done with the suggested name.

larsoner:
Tests are failing after a very long timeout. 

1. Can you see why?
2. One possibility is that `while` never exits. I usually prefer `for ii in range (100): ...` or similar just in case something misbehaves.
3. Can you change the timeout on the jobs to be like 10-20 minutes? For example like this: https://github.com/mne-tools/mne-python/blob/ec87fd8976a82ef7541c3114cd7955083eb278d6/.github/workflows/tests.yml#L20

DimitriPapadopoulos:
Strange, it did work from the command line, but probably before I changed the existing code into a function :smile:

PR #3213, opened 2023-11-11
Ignore ill-formed INI files instead of crashing

### LINKED ISSUE #3212: codespell exits brutally on ill-formed config file
```
$ cat CONFIG 
check-hidden =
$ 
$ codespell --config CONFIG
  File "/home/username/.local/bin/codespell", line 8, in <module>
    sys.exit(_script_main())
             ^^^^^^^^^^^^^^
  File "/path/to/codespell/codespell_lib/_codespell.py", line 1019, in _script_main
    return main(*sys.argv[1:])
           ^^^^^^^^^^^^^^^^^^^
  File "/path/to/codespell/codespell_lib/_codespell.py", line 1024, in main
    options, parser, used_cfg_files = parse_options(args)
                                      ^^^^^^^^^^^^^^^^^^^
  File "/path/to/codespell/codespell_lib/_codespell.py", line 584, in parse_options
    _cfg.read(cfg_file)
  File "/usr/lib/python3.11/configparser.py", line 713, in read
    self._read(fp, filename)
  File "/usr/lib/python3.11/configparser.py", line 1101, in _read
    raise MissingSectionHeaderError(fpname, lineno, line)
configparser.MissingSectionHeaderError: File contains no section headers.
file: 'CONFIG', line: 1
'check-hidden =\n'
$ 
```

DimitriPapadopoulos:
Fixes #3212.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -118,6 +118,7 @@ _builtin_default = "clear,rare"
 EX_OK = 0
 EX_USAGE = 64
 EX_DATAERR = 65
+EX_CONFIG = 78
 
 # OPTIONS:
 #
@@ -586,7 +587,7 @@ def parse_options(
             used_cfg_files.append(cfg_file)
 
     # Use config files
-    config.read(cfg_files)
+    config.read(used_cfg_files)
     if config.has_section("codespell"):
         # Build a "fake" argv list using option name and value.
         cfg_args = []
@@ -1021,7 +1022,14 @@ def _script_main() -> int:
 
 def main(*args: str) -> int:
     """Contains flow control"""
-    options, parser, used_cfg_files = parse_options(args)
+    try:
+        options, parser, used_cfg_files = parse_options(args)
+    except configparser.Error as e:
+        print(
+            f"ERROR: ill-formed config file: {e.message}",
+            file=sys.stderr,
+        )
+        return EX_CONFIG
 
     # Report used config files
     if not options.quiet_level & QuietLevels.CONFIG_FILES:
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -13,7 +13,13 @@ from typing import Any, Generator, Optional, Tuple, Union
 import pytest
 
 import codespell_lib as cs_
-from codespell_lib._codespell import EX_DATAERR, EX_OK, EX_USAGE, EX_CONFIG, uri_regex_def
+from codespell_lib._codespell import (
+    EX_CONFIG,
+    EX_DATAERR,
+    EX_OK,
+    EX_USAGE,
+    uri_regex_def,
+)
 
 
 def test_constants() -> None:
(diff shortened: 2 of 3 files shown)


codespell_lib/_codespell.py
```diff
@@ -581,12 +581,15 @@ def parse_options(
     used_cfg_files = []
     for cfg_file in cfg_files:
         _cfg = configparser.ConfigParser()
-        _cfg.read(cfg_file)
+        try:
+            _cfg.read(cfg_file)
+        except configparser.Error:
+            continue  # ignore unparsable config file!
```
peternewman, line 587: I think it ought to tell you it's failed to read it (with the error) at least for some level of verbose logging. Is setup.cfg likely to be invalid generally, or only when someone has failed to write it properly.

Otherwise it's going to be hard to troubleshoot what's going on/why it's not working for you as expected.

peternewman (changes requested):
LGTM, I just don't think it should silently fail to read them.

codespell_lib/_codespell.py
DimitriPapadopoulos: I had left out logging because this function did not log until now, just returning a status. But yes, some message or error status is needed.

Need to add tests too.

peternewman: It sort of inadvertently did via the stack trace! 😆

DimitriPapadopoulos:
I have changed codespell to fail upon broken INI config files:
```
$ cat CONFIG 
check-hidden =
$ 
$ codespell --config=CONFIG
ERROR: ill-formed config file: File contains no section headers.
file: 'CONFIG', line: 1
'check-hidden =\n'
$ 
```

codespell_lib/_codespell.py
```diff
@@ -1021,7 +1022,14 @@ def _script_main() -> int:
...
     """Contains flow control"""
-    options, parser, used_cfg_files = parse_options(args)
+    try:
+        options, parser, used_cfg_files = parse_options(args)
+    except configparser.Error as e:
+        print(
+            f"ERROR: ill-formed config file: {e.message}",
+            file=sys.stderr,
+        )
+        return EX_CONFIG
```
peternewman, line 1032: I think this will mean you can't ever use Codespell if your setup.cfg has an invalid config for example (as you can't override just append these):
https://github.com/codespell-project/codespell/blob/ca44052fa85a02609492e6b51771a058f79844e2/codespell_lib/_codespell.py#L547-L548

Not sure what the best solution is, but thought it's worth flagging before we merge...

larsoner: Maybe warn if it's the root-level `setup.cfg` and error otherwise?

DimitriPapadopoulos: That was already the case, with an exception raised. Nobody has complained.

larsoner:
Thanks @DimitriPapadopoulos !

codespell_lib/_codespell.py
peternewman: > That was already the case, with an exception raised. Nobody has complained.

Oops, I didn't think of that! :facepalm:

PR #3219, opened 2023-11-14
Ignore BrokenPipeError exception

jcubic:
I've also found that when using:

```
codespell | less
```

and pressing `q` (to quit) it throws exception `BrokenPipeError`.
It is also safe to ignore this exception.


codespell_lib/__main__.py
```diff
@@ -5,5 +5,7 @@
 if __name__ == "__main__":
     try:
         sys.exit(_script_main())
+    except BrokenPipeError:
+        pass
```
DimitriPapadopoulos, line 9: Perhaps just:
```python
    except (KeyboardInterrupt, BrokenPipeError):
        pass
```

jcubic: yes, that's better I'm not that good a Python

jcubic: done

DimitriPapadopoulos:
Looks perfect @jcubic. Thank you for fixing the script.

PR #3224, opened 2023-11-17
Add --stdin-single-line option

Jendker:
Adds an option to output just a single line for each misspelling. Current only option with stdin outputs 2 lines per error which makes parsing difficult, e.g. here it requires extensive code to customize the parser to how codespell in stdin behaves: https://github.com/mfussenegger/nvim-lint/pull/457

Moreover current stdin behavior with output of 2 lines is not in line with the usual output of codespell, which normally outputs just a single line for each misspelling.

Suggested change does not change the default. Added logic is optional and useful.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -535,6 +535,11 @@ def parse_options(
         metavar="LINES",
         help="print LINES of surrounding context",
     )
+    parser.add_argument(
+        "--stdin-single-line",
+        action="store_true",
+        help="output just a single line for each misspelling in stdin mode",
+    )
     parser.add_argument("--config", type=str, help="path to config file.")
     parser.add_argument("--toml", type=str, help="path to a pyproject.toml file.")
     parser.add_argument("files", nargs="*", help="files or directories to check")
@@ -993,6 +998,8 @@ def parse_file(
                         f"{cfilename}:{cline}: {cwrongword} "
                         f"==> {crightword}{creason}"
                     )
+                elif options.stdin_single_line:
+                    print(f"{cline}: {cwrongword} ==> {crightword}{creason}")
                 else:
                     print(
                         f"{cline}: {line.strip()}\n\t{cwrongword} "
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
I understand the use case is:
```
$ cat file.txt 
abandonned
$ 
$ codespell file.txt 
file.txt:1: abandonned ==> abandoned
$ 
$ codespell - < file.txt 
1: abandonned
	abandonned ==> abandoned
$ 
```
I don't know about this. It seems possible to simplify the code in https://github.com/mfussenegger/nvim-lint/pull/457 to start with, without changing codespell. I would recommend starting with that. I am not the maintainer, so I won't have the last word on this, but I would recommend against adding more complexity in codespell to avoid complexity in a single use case - especially when it seems quite possible to avoid complexity in that single use case by improving the Lua code there.

Jendker:
I see it more like a streamlining of codespell. Currently a single line is output per misspelling if we pass a file as an arugment but two lines are provided per misspeling if stding mode (`-` argument) is used. This option unifies the output format.

DimitriPapadopoulos:
Precisely, I don't know about that. The current behaviour is not by accident, it is on purpose. Your suggestion to "unify" (even optionally) breaks that purpose.

I would myself remove the two lines output when reading stdin if I were to re-write from scratch. But I feel that it's too late to change the behaviour and that the complexity of this additional option is not worth it.

Jendker:
I completely agree with you, thank you for the feedback. But importantly it would be just an optional change which wouldn't modify the default behavior for backward compatibility. Let's see what is the opinion of maintainers on that.

larsoner:
To me it's not much additional maintenance burden to support the single-line case and it does make things easier in some cases, so +1 for adding it

larsoner:
... but we will need a test that it works as designed :)

Jendker:
Great, glad to hear it! I'm adding tests for this functionality.

Jendker:
@larsoner @peternewman I have added stdin test. Please let me know if anything is missing. Thank you!

larsoner (approved):

larsoner:
Thanks @Jendker !

Issue #2682, opened 2022-12-29
Acknowledge dictionary sources


DimitriPapadopoulos:
Probably to be added at the end of [License](https://github.com/codespell-project/codespell#license):
> `dictionary.txt` and the other `dictionary_*.txt` files are derivative works of English Wikipedia and are released under the [Creative Commons Attribution-Share-Alike License 3.0](https://creativecommons.org/licenses/by-sa/3.0/).

DimitriPapadopoulos:
* [Robert Heckendorn's List of Hard to Spell Words](http://marvin.cs.uidaho.edu/misspell.html)
* [vlajos/misspell-fixer](https://github.com/vlajos/misspell-fixer)
* [Fossies](https://fossies.org/codespell_usage_hints.html)
*  Linux kernel's [spelling.txt](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/scripts/spelling.txt) file
* ...

skangas:
See also https://github.com/codespell-project/codespell/pull/2653#issuecomment-1364173193. I think the challenge here is to make it in a way that is both clear, yet still not confusing.

PR #3231, opened 2023-11-22
chore(license): update to use spdx id

chenrui333:
it would be good to use spdx license ID to avoid confusion.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ name = "codespell"
 description = "Codespell"
 readme = { file = "README.rst", content-type = "text/x-rst" }
 requires-python = ">=3.8"
-license = {text = "GPL v2"}
+license = {text = "GPL-2.0-only"}
 authors = [
     {name = "Lucas De Marchi", email = "lucas.de.marchi@gmail.com"},
 ]


DimitriPapadopoulos:
I agree that is clearer, but the reality is that the license for the Python module and the dictionaries are not the same. See https://github.com/codespell-project/codespell/issues/2682. I think we'd better spend time on explaining that complexity, instead of over-simplifying.

DimitriPapadopoulos:
@larsoner I will let you decide on this one. Either over-simplify and use [SPDX](https://spdx.org/licenses/) identifiers as described in [PEP 639](https://peps.python.org/pep-0639/) (licenses had been left out of [PEP 621](https://peps.python.org/pep-0621/) on purpose to let a future PEP specify support for [SPDX](https://spdx.org/licenses/) identifiers).

chenrui333:
based on my observations of this license declaration, some folks intend to do the exhausted acknowledgements, some folks just do it for the project itself. Either way, I think this change would be a good move (maybe part of the series of changes)

larsoner:
To me this is just using a different (better, more standard) name, so +1 from me. I know `grayskull` for example does SPDX matching so this might help some people/tools

larsoner (approved):

PR #3242, opened 2023-12-08
black → ruff format

DimitriPapadopoulos:
Rationale:
- Use less tools, as we already use ruff.
- While black has always been very opinionated, ruff does provide a few options.

See for example https://github.com/pypa/setuptools/pull/4125#issuecomment-1822842090 and https://github.com/jaraco/skeleton/pull/99.

Also add new [pyupgrade (UP) ruff rules](https://docs.astral.sh/ruff/rules/#pyupgrade-up).


larsoner:
Thanks @DimitriPapadopoulos !

PR #3247, opened 2023-12-11
ot is a typo also for it, which i is close to o

yarikoptic:


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
@yarikoptic Thank you for your regular contributions, and sorry for not always be able to rapidly follow up on your suggestions.

PR #3250, opened 2023-12-13
Get rid of autoflake

DimitriPapadopoulos:
We have already moved to [ruff](https://astral.sh/ruff), no need to keep [autoflake](https://github.com/PyCQA/autoflake) as far as I can see.

@cclauss Why did you keep autoflake in #2779? Doesn't ruff cover autoflake (https://github.com/astral-sh/ruff/issues/1647)?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,10 +62,6 @@ repos:
     hooks:
       - id: ruff
       - id: ruff-format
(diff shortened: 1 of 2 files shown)


larsoner (approved):

larsoner:
Thanks @DimitriPapadopoulos @cclauss !

PR #3252, opened 2023-12-14
Add two choices for verision typo fix

yarikoptic:
Originally detected/fixed in
https://github.com/zenodo/zenodo/pull/2519

Apparently there is 13k hits on github:
https://github.com/search?q=verision&type=code

The other possible close match is derision, which is a legit word and is used in about of 45k hits on github: https://github.com/search?q=derision&type=code

'd' is nearby 'v', so even though unlikely, I have decided that best to have multiple choices here.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -61756,6 +61756,8 @@ verions->versions
 veriosn->version
 veriosns->versions
 verious->various
+verision->version, revision, derision,
+verisions->versions, revisions,
 verison->version
 verisoned->versioned
 verisoner->versioner


codespell_lib/data/dictionary.txt
```diff
@@ -57862,6 +57862,7 @@ verions->versions
 veriosn->version
 veriosns->versions
 verious->various
+verision->version,derision
```
DimitriPapadopoulos, line 57865: Perhaps add:
```
verisions->versions, derisions,
```
or:
```
verisions->versions
```

DimitriPapadopoulos: Could it be a misspelling of `revision` too?

yarikoptic: ah, missed the syntax... 

> Could it be a misspelling of `revision` too?

ha -- great catch -- just swapping of 2 letters, I didn't immediately see it even! nice. I even see it possibly happening more likely than for some rare `derision` so will place it 2nd

ok, let's do/accept suggestion

```suggestion
verision->version, revision, derision,
```

yarikoptic: grr... github is giving me hard time... need to run, will try to merge again later:

image

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @yarikoptic.

PR #3259, opened 2023-12-21
Add support for ANSI colors on Windows

perillo:
Windows Terminal is the Default in Windows 11, so there is no reason to not support ANSI colors on Windows.

Explicitly disable ANSI colors on WASI, since ANSI escape characters are disabled for security reasons.

Tested on a Windows 10.0.19045.0 laptop, with Windows Terminal installed.


perillo:
Fixed code style.

perillo:
Update the code to open `CONOUT$`, instead of using `GetStdHandle(STD_OUTPUT_HANDLE)`.

perillo:
@DimitriPapadopoulos: `pytest-cov` complains about the coverage of the `_supports_ansi_colors` function, but IMHO it is impossible to fix since the result is fixed.

Is it possible to exclude the function from the total coverage?

peternewman:
> Update the code to open `CONOUT$`, instead of using `GetStdHandle(STD_OUTPUT_HANDLE)`.

That looks like it will test the console, rather than where STDOUT has been redirected to, the latter may not have ANSI colour support. It might be nice if it works properly, if I pipe top, I get the ANSI sequences as plain characters, it would probably be better if it dropped them.

I'm not surprised GitHub is confused as it's probably redirecting it. What happens if you redirect locally, do you get the same error?

perillo:
> > Update the code to open `CONOUT$`, instead of using `GetStdHandle(STD_OUTPUT_HANDLE)`.
> 
> That looks like it will test the console, rather than where STDOUT has been redirected to, the latter may not have ANSI colour support. It might be nice if it works properly, if I pipe top, I get the ANSI sequences as plain characters, it would probably be better if it dropped them.

You are right, thanks.
Maybe the simple solution is to return `False` instead of raising an exception.

> 
> I'm not surprised GitHub is confused as it's probably redirecting it. What happens if you redirect locally, do you get the same error?

What do you means by "redirecting locally"?

As a wrote in a previous comment, I did not get an error when running `codespell` from a Linux PC connected to a Windows 10 Laptop via SSH.  Colors where printed correctly on my local terminal.

perillo:
Revert last change, using `STD_OUTPUT_HANDLE` again.

Don't raise an exception in case `GetConsoleMode` fails, just return `False` (or maybe print a warning with the error message on `stderr`).

perillo:
Looking at the Zig language implementation of `supportsAnsiEscapeCodes`, it is a good idea to returns `False` when the `TERM` environment variable is equal to `dumb`.

DimitriPapadopoulos:
Looks good to me. Does it sound feasible to add tests?

perillo:
> Looks good to me. Does it sound feasible to add tests?

I think that we can only test that the `_supports_ansi_colors` function does not raise an exception.

We can also add an older windows runner (`windows-2019`) to the matrix.

An open issue is `wasi` support, since it is not tested; should we remove it?

perillo:
There is an unofficial build of Python for wasm32-wasi: https://github.com/tiran/cpython-wasm-test/releases/download/v3.11.0/Python-3.11.0-wasm32-wasi-16.zip

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
About `wasi` support, I guess hearsay is sufficient for now: no need to test on an actual platform.

I haven't looked into the TERM variable and the details of the Zip implementation. What about `return posix.isCygwinPty(self.handle)` in that Zig implementation?

perillo:
> About `wasi` support, I guess hearsay is sufficient for now: no need to test on an actual platform.
> 
> I haven't looked into the TERM variable and the details of the Zip implementation. What about `return posix.isCygwinPty(self.handle)` in that Zig implementation?

`isCygwinPty` is messy: https://github.com/ziglang/zig/blob/master/lib/std/os.zig#L3305
### 
I don't think it is necessary to support cygwin.

As for disabling colors if the terminal is `dumb`, it is probably not necessary, since the user can just use the `--disable-colors` option.

Thanks.

DimitriPapadopoulos:
@perillo Thank you for your contribution and the thorough review of possible solutions.

PR #3270, opened 2024-01-02
Make typos lowercase

DimitriPapadopoulos:
Typo matching is case insensitive, it doesn't make sense to use uppercase in this context.

While we would like a mechanism to suggest that "manuel" is a typo but "Manuel" is not, writing the typo as "Manuel" doesn't help in any way.

We keep uppercase in suggested fixes, that could be useful some day.

See also https://github.com/codespell-project/codespell/issues/1578#issuecomment-1874666853.


DimitriPapadopoulos:
@larsoner or @peternewman Could you have a look?

This does create a backwards compatibility issue: some modified typos may appear in existing lists of words to ignore, using the former capitalisation of the first letter. However, I feel it is worth the pain.

larsoner:
> I feel it is worth the pain.

Indeed this is likely to cause problems for people.

What is the gain here exactly? If it's just logical and cleaner but doesn't help end users, I'm not sure it's worth inflicting the pain.

DimitriPapadopoulos:
The gain is that the case of `-L` arguments would be simpler to explain: "_all lowercase_" instead of "_the case used in the dictionary the word is in_". I had left such a documentation change for later.

larsoner:
One backward compatible way to accomplish this would be to check if the ignore matches the upper or lower cased version

luzpaz (approved):

PR #3272, opened 2024-01-03
Case ignore

vEnhance:
Starting on #1578, WIP


DimitriPapadopoulos:
#3270 makes all typos (but not suggested fixes) lowercase. Does your implantation require #3270 as a prior?

vEnhance:
> #3270 makes all typos (but not suggested fixes) lowercase. Does your implantation require #3270 as a prior?

It should be independent. The implementation of this PR only affects the reading of `-I` and `-L` options and leaves everything else the same.

In fact, the `build_dict` function that parses the built-in dictionaries currently reads (this is not changed by this PR):

```python
def build_dict(
    filename: str,
    misspellings: Dict[str, Misspelling],
    ignore_words: Set[str],
) -> None:
    with open(filename, encoding="utf-8") as f:
        translate_tables = [(x, str.maketrans(x, y)) for x, y in alt_chars]
        for line in f:
            [key, data] = line.split("->")
            # TODO for now, convert both to lower. Someday we can maybe add
            # support for fixing caps.
            key = key.lower()
            data = data.lower()
            if key not in ignore_words:
                add_misspelling(key, data, misspellings)
            # generate alternative misspellings/fixes
            for x, table in translate_tables:
                if x in key:
                    alt_key = key.translate(table)
                    alt_data = data.translate(table)
                    if alt_key not in ignore_words:
                        add_misspelling(alt_key, alt_data, misspellings)
```
that is, when reading the dictionaries included with codespell, actually everything is already converted into lowercase anyway.

codespell_lib/_codespell.py
```diff
@@ -655,24 +655,37 @@ def parse_options(
...
-                word.strip() for word in comma_separated_words.split(",")
-            )
-    return ignore_words
+            for word in comma_separated_words.split(","):
+                word = word.strip()
+                if word == word.lower():
+                    ignore_words.add(word)
+                else:
+                    ignore_words_cased.add(word)
+    return (ignore_words, ignore_words_cased)
```
DimitriPapadopoulos, line 671: We now have two categories of words to ignore when reading text:
* `ignore_words`: words with lowercase characters only
* `ignore_words_cased`: words with at least an uppercase character

This will indeed help if we want to ignore `Manuel` but not `manuel`, which is a common requirement.

codespell_lib/_codespell.py
```diff
@@ -655,24 +655,37 @@ def parse_options(
...
+    filename: str, ignore_words: Set[str], ignore_words_cased: Set[str]
+) -> None:
     with open(filename, encoding="utf-8") as f:
-        ignore_words.update(line.strip() for line in f)
+        for line in f:
+            word = line.strip()
+            if word == word.lower():
+                ignore_words.add(word)
+            else:
+                ignore_words_cased.add(word)
```
DimitriPapadopoulos, line 688: Perhaps factorize common code in a single function:
```python
def classify_ignore_words(words: Iterable[str], ignore_words: Set[str], ignore_words_cased: Set[str]) -> None:
    for word in words:
        word = words.strip()
        if word == word.lower():
            ignore_words.add(word)
        else:
            ignore_words_cased.add(word)
```
Not sure whether file `f` qualifies as an "iterable of `str`".

codespell_lib/_codespell.py
```diff
@@ -1133,7 +1154,10 @@ def main(*args: str) -> int:
         )
         parser.print_help()
         return EX_USAGE
-    uri_ignore_words = parse_ignore_words_option(options.uri_ignore_words_list)
+    uri_ignore_words_lowercase, uri_ignore_words_cased = parse_ignore_words_option(
+        options.uri_ignore_words_list
+    )
+    uri_ignore_words = uri_ignore_words_lowercase | uri_ignore_words_cased
```
DimitriPapadopoulos, line 1160: Perhaps replace with:
```python
    uri_ignore_words = {itertools.chain(parse_ignore_words_option(options.uri_ignore_words_list))}
```

codespell_lib/_codespell.py
vEnhance: good idea, I implemented this

codespell_lib/_codespell.py
vEnhance: this snippet didn't work for me as written but
```python
set().union(*parse_ignore_words_option(options.uri_ignore_words_list))
```
did work and has the advantage it doesn't need another import of `itertools`.

DimitriPapadopoulos: Try this instead:
```python
    uri_ignore_words = set(itertools.chain(parse_ignore_words_option(options.uri_ignore_words_list)))
```
`itertools` is a standard library and chaining is what we do here.

```python
>>> import itertools
>>> set1 = {"a", "b", "c"}
>>> set2 = {"1", "2", "3"}
>>> {itertools.chain(set1, set2)}
{<itertools.chain object at 0x7fc52e24c700>}
>>> set(itertools.chain(set1, set2))
{'1', 'c', 'b', '3', 'a', '2'}
>>> 
```

vEnhance: OK I can do that (but I think you need a `*` before `parse_ignore_words_option` for it to work).

DimitriPapadopoulos (approved):
LGTM. @larson do you want to have a look?

larson:
> LGTM. @larson do you want to have a look?

I think you meant to tag @larsoner

DimitriPapadopoulos:
@larsoner Would you like to have a look?

larsoner:
Code changes look reasonable. There is some risk that someone with custom dictionaries will get different behavior with this PR, though. Safest would be to add a switch to make this opt-in like `--ignore-words-case-sensitive` or similar. Less safe but maybe useful (or maybe YAGNI) would be to add a `--ignore-words-case-insensitive` that could get back the old behavior, but people can also get the old behavior just by modifying their dictionaries. I think I can live with the backward incompatible change but wanted to make sure we thought about it first!

vEnhance:
> There is some risk that someone with custom dictionaries will get different behavior with this PR, though.

So I believe the changes in this PR should only affect words passed by config or command line options, which is why I stated in https://github.com/codespell-project/codespell/pull/3272#issuecomment-1875827271 this was independent of #3270. Similarly, there should be no backwards-incompatability with custom dictionaries either, because those are read using `build_dict` which has remained unchanged.

The only users who should see any changes are people who have specified, say, something like `SEH` in `-I` which, as mentioned previously (e.g. https://github.com/codespell-project/codespell/issues/1578#issuecomment-1763478663), actually did nothing. I can't actually imagine any situations where someone would add a word into an ignore list and want/expect it to have no effect.

Please let me know if I misunderstood.

larsoner:
Thanks @vEnhance !

PR #3274, opened 2024-01-03
fix: typos in comments

vEnhance:
Ironically, the word "mispelled" is misspelled in some doc comments. I _think_ this is not intentional?


DimitriPapadopoulos:
Nice catch @vEnhance!

PR #3282, opened 2024-01-10
dictionnary: persan can be persian.

mdeweerd:
Found a case where persan should be persian.


codespell_lib/data/dictionary.txt
```diff
@@ -40293,7 +40293,7 @@ perposing->proposing, purposing,
 perpsective->perspective
 perpsectives->perspectives
 perrror->perror
-persan->person
+persan->person, persian,
```
DimitriPapadopoulos, line 40296: It's probably more often a typo of persian than person.

DimitriPapadopoulos (approved):

PR #3283, opened 2024-01-11
Unvalidate can be devalidate.

mdeweerd:
Unvalidate can be devalidate as in something that was validated is devalidated in the UI to allow modifications and the validated again.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -56969,7 +56969,7 @@ unusued->unused
 unvailability->unavailability
 unvailable->unavailable
 unvalid->invalid
-unvalidate->invalidate
+unvalidate->invalidate, devalidate,
 unverfified->unverified
 unversal->universal
 unversality->universality


DimitriPapadopoulos (commented):
Why would you suggest `devalidate` instead of `unvalidate`?

First the initial intent was probably to write `invalidate`. Then `devalidate` is missing from mainstream dictionaries:
* [OED](https://www.oed.com/search/dictionary/?q=devalidate)
* [Cambridge](https://dictionary.cambridge.org/spellcheck/english/?q=devalidate)
* [Collins](https://www.collinsdictionary.com/spellcheck/english?q=devalidate)
* [Longman](https://www.ldoceonline.com/spellcheck/english/?q=devalidate)
* [Merriam-Webster](https://www.merriam-webster.com/dictionary/devalidate)
* [Oxford Learner's](https://www.oxfordlearnersdictionaries.com/spellcheck/english/?q=devalidate)

DimitriPapadopoulos:
The Google Ngram Viewer also shows that `devalidate` is quite uncommon:
https://books.google.com/ngrams/graph?content=invalidate%2Cdevalidate%2Cunvalidate&year_start=1800&year_end=2019&corpus=en-2019

I understand you would like to use `devalidate` for the specific case of check-boxes: _devalidate a check-box_. But then what is wrong with _invalidate a check-box_?

Note that if you insist on using `devalidate`, codespell will not show it a typo.

DimitriPapadopoulos:
I also tried Google searches:
* invalidate checkbox: [382,000 hits](https://www.google.com/search?q=invalidate+checkbox)
* devalidate checkbox: [958 hits](https://www.google.com/search?q=devalidate+checkbox)

mdeweerd:
It's not just about invalidating a checkbox - that is in itself perfect for me.

In this case the intent is to indicate "To remove validation from" as indicated here:
https://www.wordsense.eu/devalidate/ 

In this case it's related to invoices:

- The order can be "draft": neither validated nor invalidated;
- The order could then be validated - it can not be changed anymore, unless,
- The order is devalidated: it returns from a validated state to a pending/draft state;
- If the order is invalidated, this means that it is not accepted/refused, the opposite of validated.

So 'devalidate' is certainly rare because in most circumstances invalidated works fine and is appropriate.


In the cases of a checkbox related to a function, we can say that we enable/disable the function or tick/untick or check/uncheck the tickbox.

If someone wrote 'unvalidate' it needs reflection to make sure which is the better replacement.

DimitriPapadopoulos:
But that's precisely the meaning of `invalidate` itself.

DimitriPapadopoulos:
i'm not saying that the word is incorrect or that I wouldn't understand it if I saw it. However:
* The implicit rule is that we rely on mainstream curated dictionaries.
* The intent is not 100% accuracy. That's doesn't sound possible in English spellchecking, without AI and just based on plain dictionaries of word misspellings, especially if you want to take into account words coined for use in specific domains or contexts.
* I believe the intent behind occurrences of `unvalidate` was to use `invalidate`, in the vast majority of cases.

Eventually, the reference for English spelling is [SCOWL (And Friends)](http://wordlist.aspell.net/) and [`devalidate`](http://app.aspell.net/lookup?dict=en_US-large;words=devalidate) is not part of its dictionaries, even large ones.

PR #3284, opened 2024-01-12
ci: Add pre-commit workflow

mdeweerd:
This runs the pre-commit setup in a github workflow that will also annotate the PRs.

### DIFF
--- a/.github/logToCs.py
+++ b/.github/logToCs.py
@@ -1,0 +1,378 @@
+#!/usr/bin/env python3
+# pylint: disable=invalid-name
+"""
+Convert a log to CheckStyle format.
+
+Url: https://github.com/mdeweerd/LogToCheckStyle
+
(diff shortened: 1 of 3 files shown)


larsoner:
I think inline comments are neat but I'm worried about adding a bunch of new infrastructure that we need to maintain. Ideally contributors would `pre-commit install --install-hooks` locally and see the errors. And if they don't do that, then click the "details" button on the failed CI.

mdeweerd:
Seeing the annotation in context is IMHO more efficient, not just neat.  
If the annotations break in the future, the related steps can just be disabled/commented/removed.

Annotations are fairly new - I think that we'll see more features coming in the github ui.  I suppose that at some point in time an annotation can be turned into an issue in the UI - something that I do not expect in the textual report/details view.

PR #3294, opened 2024-01-16
Enable lists in TOML config file

### LINKED ISSUE #3206: Enable defining settings using comma-separated-values as lists
This is a proposal for a minor update in defining some settings.

Settings involving sequences of values (e.g. words to be ignored, files to be skipped) could be declared not as comma-separated strings, but as lists.

The idea particularly regards `pyproject.toml`, but could be potentially applicable to pre-commit hook as well.

Of course, I think we can all live with the way it's implemented now, but I guess using lists would make the `codespell` configuration more consistent with the remaining tools set up in the config files.

Anyway, the tool is great! I thank all the contributors for their efforts! 🙇🏻‍♂️ 👍🏻

DimitriPapadopoulos:
Strings in a list are changed into a string containing comma-separated substrings. Of course, this means option arguments cannot contain a comma.

Fixes #3206.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -325,7 +325,11 @@ class NewlineHelpFormatter(argparse.HelpFormatter):
 
 def _toml_to_parseconfig(toml_dict: Dict[str, Any]) -> Dict[str, Any]:
     """Convert a dict read from a TOML file to the parseconfig.read_dict() format."""
-    return {k: "" if v is True else v for k, v in toml_dict.items() if v is not False}
+    return {
+        k: "" if v is True else ",".join(v) if isinstance(v, list) else v
+        for k, v in toml_dict.items()
+        if v is not False
+    }
 
 
 def _supports_ansi_colors() -> bool:
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -1150,7 +1150,7 @@ def test_ill_formed_ini_config_file(
     assert "ill-formed config file" in stderr
 
 
-@pytest.mark.parametrize("kind", ("toml", "cfg"))
+@pytest.mark.parametrize("kind", ("cfg", "toml", "toml_list"))
 def test_config_toml(
     tmp_path: Path,
     capsys: pytest.CaptureFixture[str],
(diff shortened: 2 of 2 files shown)


DimitriPapadopoulos:

larsoner:
Going to a list to comma-separated strings just to eventually split on commas seems a bit odd. But I'm assuming it would be more work to make our code operate natively on lists deeper in the code so I can live with the current approach until someone is motivated to refactor. Thanks @DimitriPapadopoulos !

larsoner (approved):

DimitriPapadopoulos:
I can recall starting working on changing internal comma-strings into lists. I'll try to find the relevant branch. But that would be for a different PR.

dalito:
@DimitriPapadopoulos - Thanks a lot for doing this!

DimitriPapadopoulos:
@larsoner Actually, we cannot switch from comma-separated substrings to lists of strings because we feed the config to [configparser](https://docs.python.org/3/library/configparser.html) which expects exactly that: plain strings.

Switching to lists would imply:
* either interpreting TOML config files separately, which would result in extra code to manually extract the config from a TOML dict,
* or moving away from [configparser](https://docs.python.org/3/library/configparser.html) towards code capable of extracting a config from both INI and TOML files, based on the same description of settings.

Do you know of a Python library implementing the latter?

larsoner:
Don't know of any but haven't looked. If it's not trivial to make the switch we can consider doing it later. Would be good to add a comment in our code at some point about why we map to comma-separated string just to separate back out later

DimitriPapadopoulos:
Actually, we transform whatever settings we read from INI or TOML config files into their equivalent command line argument and feed them to [argparse.ArgumentParser.parse_args](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_args) — that's the function that expects plain strings:
https://github.com/codespell-project/codespell/blob/4587d788a7d1724b3cef26011b956f1d6951b307/codespell_lib/_codespell.py#L647-L661

Therefore, switching to lists would imply validating the dicts read from INI or TOML config files separately, instead of validating them with [argparse.ArgumentParser](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser).

I eventually found [ConfigArgParse](https://github.com/bw2/ConfigArgParse) in 
[Parse config files, environment, and command-line arguments, to get a single collection of options](https://stackoverflow.com/questions/6133517/parse-config-files-environment-and-command-line-arguments-to-get-a-single-col#14743159). It understands command-line options, config and INI files, TOML files, and more. It's yet unclear how it handles lists.

paduszyk:
@DimitriPapadopoulos Thanks!

PR #3299, opened 2024-01-20
Enable multiline lists in config files

DimitriPapadopoulos:
Interpret multiline values in config files as lists, as a follow up to #3294 which enables lists in TOML files.

Fixes https://github.com/codespell-project/codespell/issues/3206#issuecomment-1902089813.


DimitriPapadopoulos:

DimitriPapadopoulos:
Superseded by #3400.

PR #3304, opened 2024-01-22
Add spelling corrections for sphere

adrien-berchet:
https://github.com/search?q=hemispere&type=code => 640 hits
https://github.com/search?q=hemisperes&type=code => 137 hits
https://github.com/search?q=sperical&type=code => 5.3k hits
For `spere`/`speres` the search returns a lot of irrelevant hits so it's hard to count true hits...

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -27821,6 +27821,8 @@ hemipshere->hemisphere
 hemipsheres->hemispheres
 hemishpere->hemisphere
 hemishperes->hemispheres
+hemispere->hemisphere
+hemisperes->hemispheres
 hemlholtz->Helmholtz
 hemmorhage->haemorrhage, hemorrhage,
 hemmorhaged->haemorrhaged, hemorrhaged,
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -51105,7 +51107,10 @@ speration->separation
 sperations->separations
 sperator->separator
 sperators->separators
+spere->sphere
```
DimitriPapadopoulos, line 51110: Perhaps suggest `spare` and even `spire` and `spore`.

adrien-berchet: I added them but I don't know if it really makes sense because these suggestions don't look as simple typos to me. I mean, one have to be very tired to write `spere` instead of `spore` because the `o` is very far from the `e` on the keyboard :smile: (at least on the keyboard layouts I know)

DimitriPapadopoulos: Yes, I wasn't certain about `spire` and `spore`, but `spare` might be the intended word when typing with usual keyboards.

All these suggestions become more probable when typing using a Dvorak layout:

Dvorak keyboard layout

DimitriPapadopoulos (approved):

codespell_lib/data/dictionary.txt
adrien-berchet: Ah indeed, I'm not familiar with Dvorak 😅
Thanks for this suggestion!

DimitriPapadopoulos:
Thank you @adrien-berchet.

PR #3311, opened 2024-01-27
Add corrections for all *in->*ing words

jdufresne:
In my experience and in my personal projects, a common mistake is to write an "*ing" word (e.g. jumping) without the final "g" (e.g. jumpin).

Using a quick and dirty script, for every *ing word in the dictionary, there is now a correction when it is missing a final "g". If the word without a final "g" appeared in the aspell dictionary, it was excluded.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -54,6 +54,7 @@ abandonds->abandons
 abandone->abandon
 abandones->abandons
 abandonig->abandoning
+abandonin->abandoning
 abandonne->abandonment, abandon,
 abandonned->abandoned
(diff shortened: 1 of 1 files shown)


jdufresne:
There are a lot of words here so I understand if the review is tedious. If it would help, I'm happy to break this down by the word's first letter. That is, a PR for word starting with "A", then one for words starting with "B", etc.

DimitriPapadopoulos:
Yes, please break down into smaller merge requests. My web browser cannot handle such large diffs, reviewing the current PR would be really difficult.

jdufresne:
Sounds good. Will do this.

PR #3335, opened 2024-01-27
Add corrections for all *in->*ing words starting with "Z"

jdufresne:
In my experience and in my personal projects, a common mistake is to write an
"*ing" word (e.g. jumping) without the final "g" (e.g. jumpin).

Using a quick and dirty script, for every *ing word in the dictionary, there is
now a correction when it is missing a final "g". If the word without a final
"g" appeared in the aspell dictionary, it was excluded.

This changes only contains words starting with the letter
"Z" to contain the scope.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -59431,6 +59431,7 @@ zeored->zeroed
 zeores->zeroes
 zeoring->zeroing
 zeors->zeros
+zeroin->zeroing, zero in,
 zick-zack->zig-zag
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

jdufresne:
@DimitriPapadopoulos Thank you so much for finding the time to go through and resurrect these PRs. I had originally closed them due to time constraint but am quite happy to see they have moved along. :tada:

DimitriPapadopoulos:
@jdufresne Thank you for collating this useful list of typos in the first place.

PR #3339, opened 2024-01-30
mian->main

MercuryDemo:
A common typo when my classmates and I were first learning programming

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -35753,6 +35753,7 @@ mezmorizing->mesmerizing
 miagic->magic
 miagical->magical
 mial->mail
+mian->main
 mices->mice
 Michagan->Michigan
 micorcode->microcode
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -34732,6 +34732,7 @@ mezmorizing->mesmerizing
 miagic->magic
 miagical->magical
 mial->mail
+mian->main
```
DimitriPapadopoulos, line 34735: Perhaps `moan` as well?
```suggestion
mian->main, moan,
```

MercuryDemo: Compared to "main", "moan" doesn't seem to be a commonly used programming word, but it seems to work.
Thanks!

DimitriPapadopoulos: codespell is not used exclusively for code. Not sure what you mean by "it seems to work".

MercuryDemo: Adding "moan" is a good suggestion, thanks!

DimitriPapadopoulos: Then could you apply the change?

DimitriPapadopoulos (approved):

Issue #3220, opened 2023-11-15
"--interactive" useless without "-w"


buhtz:
I tried `codespell -i 3` to check errors and select modifications offered by codespell.
But the files are not modified even if I have chosen `Y`.

It only works when I add `-w`.

From a users perspective I would expect that `-i` implicates `-w`.

MercuryDemo:
hi, can i work on this issue？I'm looking for my first issue on github.

buhtz:
Damn. Run again into this bug and wasted half hour of my lifetime.

Suggestion: If you disagree that "-i" should implicate "-w" then minimally there should be warning when using "-i" without "-w".

DimitriPapadopoulos:
I'd be happy to merge #3220 once the issues have been fixed.

DimitriPapadopoulos:
Closed by #3340.

PR #3340, opened 2024-01-30
fix issue #3220: interactive model & write-changes

MercuryDemo:
fix issue #3220
By modifying the initial value of interactive and directly modifying the value of write_changes when the user selects the interactive model
But I think this is only a temporary solution, because codespell doesn't make any limitations on the value of interactive entered by the user, while the execution only supports 0, 1, 2, 3

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -1126,6 +1126,9 @@ def main(*args: str) -> int:
         for ifile, cfg_file in enumerate(used_cfg_files, start=1):
             print(f"    {ifile}: {cfg_file}")
 
+    if options.interactive > 0:
+        options.write_changes = True
+
     if options.regex and options.write_changes:
         return _usage_error(
             parser,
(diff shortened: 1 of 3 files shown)


DimitriPapadopoulos (approved):

DimitriPapadopoulos (commented):
Looks good to me, but tests don't pass. Not sure why at this point. Could you have a look?

Also, please change the commit message to something more meaningful, such as "_enforce --write-changes in interactive mode_". This may actually help you identify why tests don't pass.

codespell_lib/_codespell.py
```diff
@@ -1122,6 +1122,9 @@ def main(*args: str) -> int:
         for ifile, cfg_file in enumerate(used_cfg_files, start=1):
             print(f"    {ifile}: {cfg_file}")
 
+    if options.interactive >= 0:
```
DimitriPapadopoulos, line 1125:
```suggestions
    if options.interactive > 0:
```

MercuryDemo:
> Looks good to me, but tests don't pass. Not sure why at this point. Could you have a look?
> 
> Also, please change the commit message to something more meaningful, such as "_enforce --write-changes in interactive mode_". This may actually help you identify why tests don't pass.

Thanks! I'll recheck for test errors.

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @MercuryDemo. I rebased, slightly modified the PR, and fixed tests to pass.

PR #3350, opened 2024-02-04
dict: False/true typos

mdeweerd:
Add fixes for fals and tru.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25084,7 +25084,7 @@ fallhrough->fallthrough
 fallthough->fallthrough
 fallthruogh->fallthrough
 falltrough->fallthrough
-fals->false
+fals->fails, falls, false,
 falsh->flash, false,
 falshed->flashed
 falshes->flashes
@@ -55628,7 +55628,7 @@ trrigger->trigger
 trriggered->triggered
 trriggering->triggering
 trriggers->triggers
-tru->true
+tru->through, true,
 trubador->troubadour
 trubadors->troubadours
 trubadour->troubadour
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25084,6 +25084,7 @@ fallhrough->fallthrough
 fallthough->fallthrough
 fallthruogh->fallthrough
 falltrough->fallthrough
+fals->false
 falsh->flash, false,
 falshed->flashed
 falshes->flashes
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -25084,6 +25084,7 @@ fallhrough->fallthrough
 fallthough->fallthrough
 fallthruogh->fallthrough
 falltrough->fallthrough
+fals->false
```
DimitriPapadopoulos, line 25087:
```suggestion
fals->false, fails, falls, foals,
```

codespell_lib/data/dictionary.txt
```diff
@@ -55627,6 +55628,7 @@ trrigger->trigger
 trriggered->triggered
 trriggering->triggering
 trriggers->triggers
+tru->true
```
DimitriPapadopoulos, line 55631: Although uncommon:
```suggestion
tru->true, trug,
```
Perhaps even `through` as a fix for `thru`.

codespell_lib/data/dictionary.txt
mdeweerd: Ok for fails and falls, but do you think that 'foals' is common in a coding context?

codespell_lib/data/dictionary.txt
mdeweerd: Like 'foals', I find 'trug' very unlikely to be present in code, and even less misspelled.

I was not under the impression that it was the policy to have rare words in the dictionnary.

Again, I'll add that if confirmed that I should.

codespell_lib/data/dictionary.txt
DimitriPapadopoulos: The context is not only coding. Additionally, the vocabulary depends on the application domain, and some of us work in life sciences.

In this specific case, I guess you may leave `foals` out, because we already suggest a few words.

codespell_lib/data/dictionary.txt
DimitriPapadopoulos: Both [`foal`](http://app.aspell.net/lookup?words=foal) and [`trug`](http://app.aspell.net/lookup?words=trug) can be found in [SCOWLS](http://wordlist.aspell.net/), which I see as the reference for English spellchecking.

In this specific case, adding `tru->through` might be enough, because suggesting multiple words force users to give the proposed fixes a thought, and handle rare word themselves:
```suggestion
tru->true, through,
```

DimitriPapadopoulos (approved):

PR #3362, opened 2024-02-17
Add (dis)similar

mdeweerd:
'simulate' might not seem close to 'similare' but for a foreigner that for a non-native speaker that could be what he meant.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -2328,6 +2328,8 @@ aformentioned->aforementioned
 afrer->after
 afterall->after all
 afterw->after
+afteward->afterward
+aftewards->afterwards
 afther->after, father,
 aftherward->afterward
 aftherwards->afterwards


codespell_lib/data/dictionary.txt
```diff
@@ -20166,6 +20166,7 @@ dissimialr->dissimilar
 dissimialrity->dissimilarity
 dissimialrly->dissimilarly
 dissimiar->dissimilar
+dissimilare->dissimilar, dissimilarly,
```
DimitriPapadopoulos, line 20169: Not sure about the `dissimilarly` suggestion. Have you ever seen such a typo out there?

codespell_lib/data/dictionary.txt
```diff
@@ -49689,6 +49690,7 @@ similair->similar
 similairity->similarity
 similairly->similarly
 similairty->similarity
+similare->similar, similarly, simulate,
```
DimitriPapadopoulos, line 49693: Same about the `similarly` suggestion.

As for the `simulate` suggestion, I guess you thought of the Latin/Italian/Spanish verb `simulare`. I don't know, it doesn't strike me as a common error. Any typos out there?

codespell_lib/data/dictionary.txt
mdeweerd: I have seen the similare typo and usually I am asked to add all variants of typos, so I added dissimilare as well.

codespell_lib/data/dictionary.txt
mdeweerd: Yes, I was thinking of that - I often see false friends used.   French will use 'eventually' as a false friend of 'eventuellement' which means possibly.  But that is just the most common example I see.

codespell_lib/data/dictionary.txt
DimitriPapadopoulos: Do you mean you have seen `similare` as a typo of `similarly`? If so, the suggestion would obviously be OK. If not, I am not certain about it - but let's keep it!

codespell_lib/data/dictionary.txt
DimitriPapadopoulos: I see, but that would be a double error, wouldn't it? 
- `simulare` instead of `simulate`confusion
- `similare` instead of `simulare` typo

Have you actually encountered it?

mdeweerd: I saw: similare instead of similar.  But that could be a typo (not observed) for simulate, or similarly.

In the future I will only submit what I have observed and then you can add what you want to add according to the policy for codespell which I have a hard time understanding.  When I submit only what I see, I have to add other cases and when I add other cases I have to keep only what I see.

It's diffucult.

The PR is editable by the maintainers and I think that's the perfect function to manage that.

DimitriPapadopoulos: I sympathise concerning the difficulty. There's no real written rule, and I have myself moved from adding lots of possible suggestions to limiting myself to probable ones. A few thoughts:
* The distance between the typo and the suggestion is a straightforward metric. In your case, the distance is 2.
* I don't know if there are distances taking keyboard layouts into account out there, but I try to estimate the probability of a keyboard typo on common keyboard layouts.
* The type(s) of misspelling may be important too: plain misspelling (such as `supercede` instead of `supersede`), keyboard typo, words used by non-native speakers, etc. In your case, the combination of a keyboard typo and non-native speaker error seems less probable to me.

DimitriPapadopoulos: Maintainers don't have much time. In many projects, they mostly review suggestions from the community.

DimitriPapadopoulos:
```suggestion
similare->similar, similarly,
simulare->simulate,
```

mdeweerd: Unfortunately, this commit suggestion button does not have the desired effect.
image
(Requiring me tocheckout the branch, find the file, the location, update, squash, rebase, recommit, push).

I appreaciate that the maintainers doe not have a lot of time, neither do I...  So for adding a few suggestions, its a bit long - it would be nice to streamline it.

DimitriPapadopoulos: Suggestions for streamlining?

mdeweerd: - Maybe it's a setting to allow accepting the "Suggested changes";
- Give more rights to reviewers like you on the repository (?);
- Another method for submitting new spellings - but the issue is that I do not see a very low effort for that (example: a system dedicated specifically to the spelling suggestions where the commits would be grouped over time).  Another method could be to rely on existing platforms like transifex, but then one needs to create an account there.

I think that if I were managing this, I'ld probably use the issues to accept proposals, add and commit them to my branch and have a script that would order the dictionnary for me, and occasionnaly push.

One could also automate based on the issues where a maintainer could add a specific sequence in his comment like "ACCEPT: misspeling -> misspelling" that would be extracted and trigger the dictionnary update with a commit message that references (and implictly closes) the issue number so that github links them.

DimitriPapadopoulos (approved):

codespell_lib/data/dictionary.txt
DimitriPapadopoulos: Thank you @mdeweerd for these ideas.

Until we can try them, I suggest you add typos in your own repository, and open a PR after you have accumulated enough.

PR #3368, opened 2024-02-22
Add metadataa typo

yarikoptic:
I found a few cases on github but not that many actually. Quote often people just make a variable metadataA and alike.

may be should not bother? or move to another dictionary?


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @yarikoptic. Let's give it a try. We can remove it later if too many false positives.

PR #3370, opened 2024-02-23
Add correction for spoofing and spoofed.

cfi-gb:
See e.g.:
- https://github.com/search?q=spoffing&type=code
- https://github.com/search?q=spoffed&type=code
- https://grep.app/search?q=spoffing
- https://grep.app/search?q=spoffed


codespell_lib/data/dictionary.txt
```diff
@@ -51229,6 +51229,8 @@ spoace->space
 spoaced->spaced
 spoaces->spaces
 spoacing->spacing
+spoffed->spoofed
```
DimitriPapadopoulos, line 51232:
```suggestion
spoff->spoof, spiff, spaff, scoff,
spoffed->spoofed, spaffed, scoffed,
```

codespell_lib/data/dictionary.txt
```diff
@@ -51229,6 +51229,8 @@ spoace->space
...
+spoffing->spoofing
```
DimitriPapadopoulos, line 51233:
```suggestion
spoffing->spoofing, spaffing, scoffing,
spoffs->spoofs, spaffs, scoffs,
```

codespell_lib/data/dictionary.txt
cfi-gb: Thanks a lot for the suggestion. :+1:

Unfortunately i wasn't able to apply the suggestion directly in GitHub because it always showed me a "Diff changed, please refresh" and showed that recurring after each refresh (probably some kind of bug in GitHub) so i had applied the suggestions now in / via 98693210b75bceb24a0dff16ad220e551318be81

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3374, opened 2024-02-28
Add 'quotted->quoted'

sirosen:
I just made this typo in a project; easy addition to justify.
Fits the existing pattern with 'quottes->quotes'. I'm not aware of any valid use of 'quotted'.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -44590,6 +44590,7 @@ quotent->quotient
 quotin->quoting, quot in,
 quotted->quoted
 quottes->quotes
+quotting->quoting
 quried->queried
 quries->queries
 quroum->quorum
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -44588,6 +44588,7 @@ quotaions->quotations
 quoteed->quoted
 quotent->quotient
 quotin->quoting, quot in,
+quotted->quoted
 quottes->quotes
```
DimitriPapadopoulos, line 44592: Time permitting, could you add all verbal forms?
```suggestion
quottes->quotes
quotting->quoting
```

sirosen: Yep! I considered doing it, but since the variations were already slightly inconsistent, I didn't know if it would be welcome. I'll expand this PR with a few of these; starting with this suggestion.

sirosen: I've pushed a new revision with several more of these.

DimitriPapadopoulos (approved):

PR #3377, opened 2024-03-06
Add typos for expration(s)

fishilico:
It seems very common to write `expration` instead of `expiration`: https://grep.app/search?q=expration&words=true

Nevertheless, it could also be a misspelling of `exploration`. So put the two options.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -25224,6 +25224,8 @@ exportin->exporting, export in,
 exposin->exposing, expos in,
 expport->export
 exppressed->expressed
+expration->expiration
+exprations->expirations
 exprect->expect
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Perhaps `expiration` is sufficient here. Have you actually found such typos of `exploration` out there?

DimitriPapadopoulos (approved):

fishilico:
Thanks for your quick feedback!

> Perhaps `expiration` is sufficient here. Have you actually found such typos of `exploration` out there?

I have not found such typos and I agree `expiration` should be sufficient. I have updated my Pull Request accordingly.

DimitriPapadopoulos:
Thank you @fishilico.

PR #3388, opened 2024-03-25
Add 'repeatition->repetition' and variants

fishilico:
`repeatition` and the variants are used in many projects. Here are some of them.

`repeatition` is used in
https://github.com/FFmpeg/FFmpeg/blob/ecdc94b97f809d5f2b88640842fd0541951ad295/libavcodec/ilbcdec.c#L1106

`repeatitions` is used in
https://github.com/NVIDIA/cuda-samples/blob/3559ca4d088e12db33d6918621cab5c998ccecf1/Samples/5_Domain_Specific/p2pBandwidthLatencyTest/p2pBandwidthLatencyTest.cu#L204

`repeatitive` is used in
https://github.com/php/php-src/blob/f69d540541b73ae73f3d125d542f9de6ea457888/ext/standard/tests/array/array_flip_variation5.phpt#L2

`repeatitively` is used in
https://github.com/microsoft/lisa/blob/c38b340d3f133874095d0b41bc325bfbd8eb0ff7/microsoft/testsuites/network/networksettings.py#L585

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -47176,6 +47176,10 @@ repeadet->repeated
 repeadetly->repeatedly
 repeates->repeats
 repeatin->repeating, repeat in,
+repeatition->repetition
+repeatitions->repetitions
+repeatitive->repetitive
+repeatitively->repetitively
 repeatly->repeatedly
 repect->respect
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Thank you @fishilico.

PR #3390, opened 2024-03-25
docs: add the documentation of the new inline ignore comments

12rambau:
Doccument the new inline ignore capability from #2400.

Related to #3387

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -135,6 +135,27 @@ The words to ignore can be passed in two ways:
 
        codespell -L word1,word2,word3,word4
 
+Inline ignore
+-------------
+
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

larsoner:
Thanks!

I'll cut a release today unless there are other PRs we should get in first

DimitriPapadopoulos:
By the way, GitHub tends to underline code blocks in reST documents:

code-block

That's unrelated to this merge request, but I thought I'd mention it. It seems to happen with most of our code blocks (`python` and `ini`) but `toml` for example. Strange.

PR #3391, opened 2024-03-26
Workaround for reST rendering

DimitriPapadopoulos:
GitHub does not render reST `code::block::` properly, with most language specifiers. Remove language specifiers for now, until GitHub fix this issue.

Fixes https://github.com/codespell-project/codespell/pull/3390#issuecomment-2020705461.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -151,10 +151,10 @@ Words should be separated by a comma.
 
 2. ignore multiple words:
 
-    .. code-block:: python
+   .. code-block:: python
 
-         def wrod(wrods) # codespell:ignore
-              pass
+       def wrod(wrods) # codespell:ignore
+           pass
 
 Using a config file
 -------------------


DimitriPapadopoulos:
@larsoner Coudl please have a look at this PR:
- works around broken reST rendering
- fixes indentation problem in the documentation for inline ignores

larsoner:
> works around broken reST rendering

Keep in mind we use it in our sdist / on PyPI via `readme`:

https://github.com/codespell-project/codespell/blob/a6b4e1e84d3342299c51187b0d7e2bf4b9b1f698/pyproject.toml#L6

Renders into this page on release

https://pypi.org/project/codespell/

And it looks like `pygments` rendering is working okay on a per-language basis at least from spot checking a few of them e.g.:

image

One bugfix is that the `pip` lines should probably be `console` rather than `sh` I think?

In any case I think I'd prefer to keep the correct format specifiers and just wait for GH to fix their rendering at some point, and keep the nicer rendering on PyPI (and any editors that render `.rst` on the fly correctly, etc.).

12rambau:
Agreed with @larsoner I think it's a bug from Gh rendering

DimitriPapadopoulos:
Yes, that's a bug in GitHub rendering. It occurs with all of `sh`, `console`, `python`, `ini`, `yaml`, but not `toml`.

I find decent rendering of the README in GitHub is more important than fancy syntax highlighting. Any way, you decide.

At least let's merge the second commit, which ensures better rendering through consistent indentation of reST code blocks.

larsoner:
Okay based on your edit how about a compromise where we change `sh`->`console` and keep `toml`?

DimitriPapadopoulos:
Let's keep it as is then. The real language type would be `sh`. I guess GitHub doesn't support `console` at all which is why it does not even try fancy rendering.

PR #3400, opened 2024-04-10
Handle multiline options

### LINKED ISSUE #3399: Splitting multiple 'skip' paths onto indented config file lines behaves differently than when on a single line
The INI file format claims to support an option value that is split across multiple lines, if the lines after the first are indented. However, I see different behavior with the two settings

```
skip=*.adoc,./build_tests/*,
```

```
skip=*.adoc,
    ./build_tests/*,
```

The first works (skips a file with misspellings under build_tests/BAD), the second fails (reports the misspelling). Am I missing something about the INI file format that would allow this to work?

Attaching a zipfile with this minimal example.

[codespell-config.zip](https://github.com/codespell-project/codespell/files/14926060/codespell-config.zip)

DimitriPapadopoulos:
Clean up options expecting lists before using them, as they may contain newlines.

Examples:
* Enclosing command-line arguments in quotes may introduce newlines in option values:
  ```console
  $ codespell -S "A, B,
  >               C, D, E"
  ```
* INI files may contain multiline values:
  ```ini
  [codespell]
  skip = A, B,
         C, D, E,
   ```

In all the above cases, the option parsing mechanism keeps the newlines (and spaces) . We need to clean up, by splitting on commas and stripping each resulting item from newlines (and spaces).

Fixes #3399.

Very much related to #2767. Both pull requests introduce an option to flatten the options list after splitting on commas, and both need tests!

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -160,15 +160,11 @@ class QuietLevels:
 
 
 class GlobMatch:
-    def __init__(self, pattern: Optional[List[str]]) -> None:
-        self.pattern_list: Optional[List[str]] = pattern
+    def __init__(self, pattern: List[str]) -> None:
+        self.pattern_list: List[str] = pattern
 
     def match(self, filename: str) -> bool:
-        return (
-            any(fnmatch.fnmatch(filename, p) for p in self.pattern_list)
-            if self.pattern_list
-            else False
-        )
+        return any(fnmatch.fnmatch(filename, p) for p in self.pattern_list)
 
 
 class Misspelling:
@@ -493,6 +489,7 @@ def parse_options(
         "accepts globs as well. E.g.: if you want "
         "codespell to skip .eps and .txt files, "
         'you\'d give "*.eps,*.txt" to this option.',
+        default=list(),
     )
 
     parser.add_argument(
@@ -1106,19 +1103,20 @@ def parse_file(
 
 
 def flatten_clean_comma_separated_arguments(
-    arguments: Optional[List[str]],
-) -> Optional[List[str]]:
+    arguments: List[str],
+) -> List[str]:
     """
     >>> flatten_clean_comma_separated_arguments(["a, b ,\n c, d,", "e"])
     ['a', 'b', 'c', 'd', 'e']
     >>> flatten_clean_comma_separated_arguments([])
-    >>> flatten_clean_comma_separated_arguments(None)
+    []
     """
-    return (
-        [item.strip() for argument in arguments for item in argument.split(",") if item]
-        if arguments
-        else None
-    )
+    return [
+        item.strip()
+        for argument in arguments
+        for item in argument.split(",")
+        if item
+    ]
 
 
 def _script_main() -> int:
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
@larsoner Would agree to merge this PR that fixes a codespell bug? Full tests already added.

larsoner:
@DimitriPapadopoulos I simplified the logic by avoiding the `Optional` stuff and just always having a `List[str]` and making use of stuff like `any([]) == False`, feel free to merge if you're happy with the change

codespell_lib/_codespell.py
```diff
@@ -493,6 +489,7 @@ def parse_options(
         "accepts globs as well. E.g.: if you want "
         "codespell to skip .eps and .txt files, "
         'you\'d give "*.eps,*.txt" to this option.',
+        default=list(),
```
DimitriPapadopoulos: Default arguments shouldn't be mutable.

larsoner: Default arguments of functions sure. But for a `parser` instance that gets created and used once here I don't see a problem with it (it will get mutated as things get appended but that's fine)?

DimitriPapadopoulos: Or pass a tuple instead, which is good enough here since we just need to iterate, so anything `Iterable` is OK.

DimitriPapadopoulos: You're right, t's not a problem if the function/method is used only once. But linters may raise issue (I think there's a ruff/refurb rule for that). Any way, the fix is so simple that I went ahead and fixed it.

DimitriPapadopoulos: The fix is not so simple :grin:

Note that parser.add_argument("-S", ...) might be called multiple times. Isn't that a **potential** problem? I know it is not a problem here, but it does require some knowledge of the inner workings of [ArgumentParser.add_argument](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument), doesn't it?

larsoner: > Note that parser.add_argument("-S", ...) might be called multiple times. 

To be clear I don't think the `add_argument` will be called multiple times but rather that the user might pass multiple `-S`'s, right? Internally `parser` will just call `var.append(...)` which seems fine. If you didn't supply an argument the default is `None` so it would just on the first `-S` create its own list then mutate that.

Think about it this way if we did:
```
mylist = []
parser.add_arg('--whatever', default=mylist)
parser.parse()
```
now you can access the args via `parser.whatever` but that will match `mylist`, i.e. `mylist == parser.whatever` but this isn't a problem (and it's also not a problem that we don't bother to store `mylist` in the local variables and just use `parser.whatever`).

I'm not sure what the fear is here. Typically the "no mutable default" is used for *function arguments* because people didn't realize that doing:
```
def fun(a=list()):
    a.append('b')
    return a
```
would do
```
>>> fun()
['b']
>>> fun()
['b', 'b']
```
because of the mutable default argument of the function, i.e., multiple calls have an unexpected behavior. But we aren't dealing with a mutable function argument here, but rather passing a mutable variable to a method that we know later will have `.append` called on it when we call `parser.parse()` (and we call that once!).

DimitriPapadopoulos: You're right of course. This is function **call**, not a function **definition**. The name `default` somehow made me switch to "function definition" mode.

Any way, I think the final code is equivalent in terms of complexity.

PR #3421, opened 2024-05-08
Add `arragne->arrange` typo

korverdev:
Added typo and variants.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -5297,6 +5297,10 @@ arragement->arrangement
 arragements->arrangements
 arrages->arranges
 arraging->arranging
+arragne->arrange
+arragned->arranged
+arragnes->arranges
+arragning->arranging
 arraies->arrays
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

DimitriPapadopoulos (commented):
I wonder whether it could be a misspelling of `arraign`.

korverdev:
@DimitriPapadopoulos It's an interesting question. This misspelling is one variant away from `arrange` and two from `arraign`, so I think it's much more likely to be a misspelling of `arrange`. Is there a set standard for this project on how "close" corrections should be from their misspellings?

DimitriPapadopoulos:
No standard. Actually, the standard could be the existence of typos out there. Also, the dictionaries are not only about typos, but any kind of misspelling. I believe non-native speakers may confuse `arraign` and `arragne` on a phonetic basis.

DimitriPapadopoulos:
Thank you @korverdev.

PR #3425, opened 2024-05-17
Add Python spellchecking API

### LINKED ISSUE #3419: Python API support
Hi,

Would you be open to supporting a public stable python API for codespell. Ideally for me, one where I as a consumer can feed codespell with words I want spellchecked and then is given back a `valid` or `invalid, here are the corrections available. If you auto-correct, use this choice` (or "Not safe for auto-correcting" if that is the data available).

My use case is that I am working on a Language Server (LSP, https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification) and I want to provide spell checking. So far, I have relied on hunspell because it had Python bindings but its accuracy on technical documents with the dictionary I found leaves a bit to be wanted.

In my use case, being able to identify the exact range of the problem is an absolute necessity as my code need to provide the text range for the editor, such that it can show the user exactly where in the open document the problem is. If word-by-word checking is not supported, then API can be pass lines of text to be spellchecked provided the result identifies exactly in the range where the problem is (that is, I need start + end index).

In this case, I would probably also need the API docs to state a bit about why the line by line text is important, because I might need to extract the underlying text from formatting to create the a synthetic line to be spellchecked. As an example, my current code tries to identify and hide common code/file references like `usr/share/foo`. In a word-by-word check, I just skip the the spellcheck call for that word. But if I need to pass a line of text to codespell, I would need to removed the ignored and here it is relevant to know how to do that replacement such that the user does not get a false positive because I attempted to avoid another false-positive.

Alternatively, a parser for the dictionaries plus the underlying dictionaries might also be an option a "light weight API" assuming they are easier to keep stable.

I have noted that `codespell` can do spell checking from stdin to stdout. However, that is a bit too heavy handed for me to easily replace my `hunspell` integration.

That is my wishlist. :) Would this be something that you would be open to supporting?

Note: By stable API, I assumed nothing was stable since `__all__ = ["_script_main", "main", "__version__"]` does not look it contains reusable APIs.

nthykier:
# Do not review yet

I will remove the draft marker once the CI is green. Until then I might rebase. I will remove this section when I am done.

# Adding a Spellchecker API

The proposed PR is a follow up to #3419 that introduces a `Spellchecker` class. This PR deliberate does **not** close #3419, since the public package name should probably be changed first and an `__all__` provided in its `__init__,py`. In its simplest form, a consumer would do something like this:

```
# Per #3419, we may want to shuffle around the top-level name so leaving this one "undefined" at the moment
from ... import Spellchecker

# Loads "clear" + " rare" by default, use `Spellchecker(builtin_dictionaries=[...])` for choose.
s = Spellchecker()
text = """\
some
long text
with
tpyos
"""
# Provide a tokenizer (simple one for the sake of the example)
# The codespell tokenizer is a bit more involved, see `line_tokenizer_factory`
tokenizer = re.compile(r"[^ ]+").finditer
lines = text.splitlines()
for line in lines:
    s.spellcheck_line(line, tokenizer)
```

There is also a `check_lower_cased_word` method for single word checks which as a part of refactoring. I left it in because it has a simpler API if you have words you want to have checked rather than lines.

## Reviewing this PR

I have optimized the review the PR by reviewing each commit individually with the aim of understanding how we got to the end result. It comes with the following commits:

```
$ git log origin/master..HEAD | git shortlog
Niels Thykier (11):
      Refactor: Move some code to new files for reuse
      Replace `data: str` with `candidates: Sequence[str]`
      Refactor dictionary into a new `Spellchecker` class
      Refactor line tokenization to simplify an outer loop
      Rewrite line spellchecking and move most of it into the `Spellchecker`
      De-indent loop body (whitespace-only change)
      Make `Spellchecker()` load builtin dictionaries by default
      Support non-regex based tokens for `spellcheck_line`
      Speed up spellchecking by ignoring whitespace-only lines
      Move `codespell:ignore` check into `Spellchecker`
      Speed up `codespell:ignore` check by skipping the regex in most cases
```

Each commit is written with the intend to be a small self-contained change.  This also means that some commits are just moving code around or even seem a bit sub-optimal with following commits just using them as stepping stone. If you see something in a commit that you want changed, please note that it might be improved in a later commit. I still kept this as one PR since some of the commits ends up with a considerable performance regression in that commit, which is then resolved in later commits. This is all a trade off between keeping commits as standalone and reviewable as possible vs. being able to cut the PR at any given point time. I preferred the former.

Note that GitHub's PR review function (like `Approve` or `Request Changes`) operates on the entire PR even though you review it commit by commit.

### Dictionary loads

The API by default just loads the default builtin dictionaries. The constructor accepts a sequence/list of names such as `clear`, `rare`. After that, the caller can manually load custom dictionaries via `load_dictionary_from_file`. This is the default mode and aimed at the API consumers to quickly get started with the common case.

The `codespell` command is not a common case. Here, dictionaries can be loaded in any other. To facilitate that,  the following flow can be used:

```python
s = Spellchecker(builtin_dictionaries=None)
s.load_dictionary_from_file("my-early-loaded-custom-dictionary.txt")
s.load_builtin_dictionaries(["clear", "rare", "informal"])
s.load_dictionary_from_file("my-late-loaded-custom-dictionary.txt")
```

### Various exclusions or regex based matches/ignores

Exclusions features supported directly in the API:

 * Inline ignores (`codespell:ignore`). Fully handled by default, caller can opt-out.
 * Ignore words, ignore words cased. Partially supported. Caller must parse/provide these to the spellchecker either at initialization or the relevant API call.

Not supported directly by the API in its current form. Caller must facilitate these:

 * Exclude lines (`-X`). Caller should load exclude the line instead.
 * All regex based matches / ignores. The caller should set these up and provide them in the tokenizer.
   (The codespell tokenizer "factory" plus related default regexes could become part of the API if desired)
 * --uri-ignore-words-list: Hidden inside the tokenizer.

### What is to be included in the API

These are the items that I would see go into the `__all__` of the codespell API in a later commit / PR.

 1. `Spellchecker` (introduced in this PR). All its methods would become public API.
 2. `Misspelling` (existing class). All its data fields would become public API (or at least,`candidates` + `fix` would be neede for me).  Also, we should probably rename it to `Correction` or `DictionaryResult`, etc., since it is not a misspelling but the codespell correctional data.
     - I would personally recommend that this class becomes read-only at an API level. However, that would require that `codespell` uses a different way to remember its interactively chosen corrections.
 4. `DetectedMisspelling` (introduced in this PR). Subject to renaming of the class or its properties.
 5. `LineTokenizer` + `Token` + their generic constraint (`T`) for typing purposes.

That is the basic API I envisioned and I would need. This API would still have a considerable amount of "bring your own" code, notably the tokenizer is something people will probably struggle with.  My project has its own with its own rules for how things should be split into tokens (like ignoring quoted words), so this was not a goal for me.  However, it could still be important for you, since it will affect more casual API consumers that did not solve this problem ahead of time. :)

An alternative to more API would to have more examples of how to emulate the codespell command line tool.

## Performance cost of the API

It surprised me a bit that at some point the total regression exceeded 10% on the corpus I used for testing (see #3419 for the details on the performance test setup). It was restored to an ~8% regression before the first `Speed up` commit. 

The first `Speed up` commit reduces the regression to `<= 1.5%` (as in, the `~0.060s` ballpark) The second `Speed up` commit was "unnecessary show off" to get below the baseline. :laughing: 

These commits speed up commits are technically a "compensation" changes. As an example, I believe you would get a similar boost of performance by retrofitting the second speed up commit on the existing code base for a similar win. The first speed up commit is partly a compensation, partly a counter of the regression since the "overhead per line" got measurably higher (accordingly, reducing the number of redundant lines now makes the improvement considerably more worth it).

The end result is `5.6s -> 4.9s` with an API and 1½-2 performance related compensations.


nthykier:
> @nthykier let me know when I should look. I at least just approved CI runs.
> 
> One option to make it so that I don't have to approve them would be to make some other small PR that we merge. For example it could be just your couple of speed improvements. It would make an ugly rebase issue here, though, so understandable if you'd rather make a different simple PR (or just keep waiting fro me to click "approve" on the runs).

Thanks.

This PR should be ready for review now. :)

codespell_lib/_spellchecker.py
```diff
@@ -16,24 +16,267 @@
...
+
+_codespell_ignore_tag = "codespell:ignore"
+_inline_ignore_regex = re.compile(
+    rf"[^\w\s]\s?{_codespell_ignore_tag}\b(\s+(?P<words>[\w,]*))?"
+)
+
+
+class UnknownBuiltinDictionaryError(ValueError):
+    def __init__(self, name: str) -> None:
+        super().__init__(f"Unknown built-in dictionary: {name}")
```
larsoner, line 120: I think codecov is right, these classes aren't used anywhere. Omit for now?

larsoner (commented):
Otherwise looks pretty straightforward !

DimitriPapadopoulos:
After this nice refactoring work, I suspect the maximum of some complexity metrics might be lowered:
https://github.com/codespell-project/codespell/blob/3760e619e2a7b7fa8467debb2889d21a34651853/pyproject.toml#L169-L174

codespell_lib/_spellchecker.py
nthykier: I originally had a `ValueError` with an error message, but that triggers a linter warning about using long error messages with exceptions and that it is better to set the message in the `__init__` (which requires the subclass shown). Therefore, omitting it triggers a red CI and is not an option unless you want me to tweak the linter to skip that check.

Alternative, I can add a test for it, since that would solve the codecov issue while avoiding the linter issue.

nthykier:
> After this nice refactoring work, I suspect the maximum of some complexity metrics might be lowered:
> 
> https://github.com/codespell-project/codespell/blob/3760e619e2a7b7fa8467debb2889d21a34651853/pyproject.toml#L169-L174

Quite possibly. What do you (as a project) expect here? That I lower them to the default values for `pylint`, see if that works and re-add the keys needed for the CI to pass lowered the the exact amount required to have the CI pass?

codespell_lib/_spellchecker.py
DimitriPapadopoulos: To clarify, we (or at least I) do not see `UnknownBuiltinDictionaryError` raised or caught in the current code, and we cannot find `load_builtin_dictionaries` either. Why do you need to define it in this PR?

In any case, the linter rule you refer to has to be [TRY003](https://docs.astral.sh/ruff/rules/raise-vanilla-args/). Without seeing the code that raises or catches the exception, I cannot easily suggest ways to pacify the linter. However, if you only raise the exception once, isn't `ValueError` sufficient?
```python
    message = f"Unknown built-in dictionary: {name}"
    raise ValueError(message)
```

PR #3432, opened 2024-05-24
Exclude bots from generated release notes

hugovk:
Similar to https://github.com/tox-dev/tox-ini-fmt/pull/120.

The generated release notes at https://github.com/codespell-project/codespell/releases/tag/v2.3.0 contains over 30 Dependabot and pre-commit updates.

These are not interesting or very useful for the reader, so let's exclude them from the next one.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,5 @@
+changelog:
+  exclude:
+    authors:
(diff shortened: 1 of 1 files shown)


larsoner:
Thanks @hugovk !

PR #3434, opened 2024-05-25
Refactor: Move some code to new files for reuse

nthykier:
No new code is introduced; only existing code is shuffled around and the functions moved are unchanged as well.

(Subset of #3425)


DimitriPapadopoulos:
Perhaps change `codespell_lib/_text_util.py` → `codespell_lib/_util.py`? Let's keep it short, the whole program is about processing text any way.

codespell_lib/spellchecker.py
```diff
@@ -0,0 +1,75 @@
...
+"""
+
+from typing import (
+    Dict,
+    Set,
+)
+
+# Pass all misspellings through this translation table to generate
+# alternative misspellings and fixes.
+alt_chars = (("'", "’"),)  # noqa: RUF001
```
DimitriPapadopoulos, line 26: Now that the file name is not "private" any more (`spellechecker` instead of `_codespell`) should we make the name of internal variables "private", that is `_alt_chars` instead of `alt_chars`? But perhaps you want to leave that to later PR, to avoid changing the contents in this PR, don't you?

DimitriPapadopoulos: Actually, are any of the (current) contents of `codespell_lib/spellechecker.py` supposed to be part of the API? Perhaps just `build_dict`?

nthykier: None of the current contents are aimed at being the API. Though even at the end, not everything in this file is aimed at being API either. My end assumption was that we would end up with

```
from codespell import Spellchecker

```
That said, I could equally see

```
from codespell.spellchecker import Spellchecker
```

Either of which would require a rename of `codespell_lib` (the former would also require a change of `__all__`). Those changes are not in my current PRs. I intended to do that in a follow up PR when we have established what the API will be and merged the relevant code first. Partly because I am concerned about the `codespell_lib` being a repeated source of complex merge conflicts, so such a change should be short, easy to review and swiftly merged to reduce the pain (merge conflicts) it will inflict.

I am fine with having it be `_spellchecker.py`  until we decide how the API will work. I can also keep the current filename and rename internal variables. The choice of not renaming these variables was about keeping a 1:1 between the code moved (renames would have required a minor change).

DimitriPapadopoulos: With all that in mind, perhaps `_spellchecker.py` is a better name in the short term, until things clarify.

nthykier: Ok, file renamed.

DimitriPapadopoulos: And perhaps `_text_util.py` → `_util.py` as per https://github.com/codespell-project/codespell/pull/3434#issuecomment-2131131938? Unless you disagree of course.

larsoner (approved):
Nice, thanks for making things easier to review @nthykier !

PR #3440, opened 2024-06-02
Consistent error messages

DimitriPapadopoulos:
The argparse error messages look like this:

	usage: codespell [-h] [--version] ...
	                 [files ...]
	codespell: error: argument -i/--interactive: invalid choice: -1 (choose from 0, 1, 2, 3)

For codespell error messages consistent with argparse error messages:
- Use `ArgumentParser.print_usage()` instead of `ArgumentParser.print_help()`.
- Print usage first and error message last.

Noticed these while  finalising #3344.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -1081,29 +1081,29 @@ def main(*args: str) -> int:
             print(f"    {ifile}: {cfg_file}")
 
     if options.regex and options.write_changes:
+        parser.print_usage()
         print(
             "ERROR: --write-changes cannot be used together with --regex",
             file=sys.stderr,
         )
-        parser.print_help()
         return EX_USAGE
     word_regex = options.regex or word_regex_def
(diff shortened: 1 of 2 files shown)


codespell_lib/_codespell.py
```diff
@@ -1081,29 +1081,29 @@ def main(*args: str) -> int:
             print(f"    {ifile}: {cfg_file}")
 
     if options.regex and options.write_changes:
+        parser.print_usage()
```
larsoner, line 1084: This pattern now shows up many times. Probably worth adding a little `_raise_usage_error(parser, msg)` for example to make the code DRY

DimitriPapadopoulos: I have refactored the code as requested.

larsoner (approved):

larsoner:
Thanks @DimitriPapadopoulos

PR #3443, opened 2024-06-04
Add re-using->reusing

korverdev:

PR #3445, opened 2024-06-05
Add reusing misspelling and variants

korverdev:
This contains the same changes as https://github.com/codespell-project/codespell/pull/3443. Creating a second PR to get around the incorrect code coverage check the prior PR was failing.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -45312,6 +45312,9 @@ re-uplods->re-uploads
 re-usable->reusable
 re-use->reuse
 re-useable->reusable
+re-used->reused
+re-uses->reuses
+re-using->reusing
 reaaly->really
 reaarange->rearrange
 reaaranges->rearranges


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
All I can think of is that intermediate coverage reports have somehow been lost.

DimitriPapadopoulos:
@larsoner Based on "[_How to prevent codecov from using a commit older than the base of the PR to do comparison?_](https://community.codecov.com/t/how-to-prevent-codecov-from-using-a-commit-older-than-the-base-of-the-pr-to-do-comparison/3513)" I suspect there is no way to convince Codecov to compare with the last commit instead of 383a32c, probably because intermediate coverage data has been lost. If so, the only solution is to "_change the status checks to be [informational](https://docs.codecov.com/docs/commit-status#informational) until the repository stabilizes_".

larsoner:
Let's see if it stabilizes soon... in theory it should only take one commit, so maybe after a couple Codecov will figure things out. If not then yes we can change to informative

corneliusroemer:
`re-using` isn't necessarily a typo.  Hyphen is used in almost 10% of cases, so this will cause very very many changes.

Is there a separate dictionary for opinionated changes? This shouldn't be in the main dict.

Google Chrome 2024-08-18 18 29 14
https://books.google.com/ngrams/graph?content=reuse%2Cre-use&year_start=1800&year_end=2022&case_insensitive=true&corpus=en&smoothing=3

See #3521

DimitriPapadopoulos:
Curated dictionaries and [SCOWLS (And Friends)](http://wordlist.aspell.net/) only have `reuse`. The grammar rule is _Use the hyphen with the prefix re only when re means again AND omitting the hyphen would cause confusion with another word_.

PR #3466, opened 2024-06-20
Add .venv to .gitignore

skangas:
This is a directory that a lot of developers use for their virtual environment, so it would be useful to have it ignored.


larsoner:
Thanks @skangas

PR #3473, opened 2024-06-26
Add machine-readable output format, and test

ferdnyc:
Passing `--machine-readable` / `-m` to codespell, or setting `machine-readable = true` in the config file, will enable an output format designed to be more easily parsed, and containing the start column of the match (usually not output).

Inspired by the `eslint -f compact` format (which is no longer part of standard eslint), but with an added `@` at the start of each line. (Useful when combined with context or summary options, to parse only the lines containing error data).

Intended primarily for actions-codespell and other wrapper scripts.

A new test, `test_machine_readable`, is added to `test_basic.py` to verify basic functionality of the new option.

The output, specifically, takes the form:
```text
@./path/file.xtn: line 1, col 1, wrongword => rightword (reason if defined)
```

Line and column numbers are 1-based (`i + 1` and `match.start() + 1` in the code, respectively).

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -465,6 +465,13 @@ def parse_options(
         default=False,
         help="print summary of fixes",
     )
+    parser.add_argument(
+        "-m",
+        "--machine-readable",
+        action="store_true",
+        default=False,
+        help="Use alternate output format intended for parsing",
+    )
 
     parser.add_argument(
         "--count",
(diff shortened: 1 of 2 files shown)


ferdnyc:
Hrm. The failure is due to ruff complaining:

> Error: codespell_lib/_codespell.py:843:5: PLR0912 Too many branches (47 > 46)
> Error: codespell_lib/_codespell.py:843:5: PLR0915 Too many statements (123 > 119)

...because I expanded `parse_file()`.

The fact that my changes pushed it _just_ above the limits make me think that every time that function is extended, it exceeds the limits, so they get raised until it's just barely passing again.

That seems like no way to live, but if that's how it's done I guess I can just bump them again...

PR #3475, opened 2024-06-27
Add --ignore-multiline-regex option.

julian-smith-artifex-com:
Allows exclusion of multiline regions. Matching regions are replaced with blank lines.

This would allow codespell to be used with PyMuPDF, where we have many multi-line strings within Python test scripts that need to be excluded.


julian-smith-artifex-com:
Apologies, i wasn't aware of these tests. Will fix in my tree and update this PR.

PR #3476, opened 2024-06-27
Add --ignore-multiline-regex option.

julian-smith-artifex-com:
Allows exclusion of multiline regions. Matching regions are replaced with blank lines.

This would allow codespell to be used with PyMuPDF, where we have many multi-line strings within Python test scripts that need to be excluded.

This is a second PR for these changes; tests now pass.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -36,6 +36,7 @@ from typing import (
     Pattern,
     Sequence,
     Set,
+    TextIO,
     Tuple,
 )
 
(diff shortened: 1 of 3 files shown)


julian-smith-artifex-com:
Could someone approve the above workflows so we can check this PR is passing tests? Thanks.

julian-smith-artifex-com:
> Done

Thanks, looks like everything is passing.

Is there anything i can do to get this PR approved or modified or rejected?

larsoner:
Nice, seems to be well documented, let's give it a shot and fix it if people run into issues. Thanks @julian-smith-artifex-com !

PR #3477, opened 2024-06-28
"Imbed" is a valid alternate form of "embed"

oddhack:
So remove suggested rewrites from the dictionary.

N.b. it is not clear from the README if the philosophy of this project is to drive people to "preferred" spellings or to "correct" spellings. If it is the latter, "imbed" and variants are valid (albeit less common) alternates to "embed" - see for example
https://www.merriam-webster.com/dictionary/imbed

(I had previously created this as #2415 and later deleted my repo fork, without realizing it would also close the PR, so retrying).

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -16825,9 +16825,6 @@ imapcting->impacting
 imapcts->impacts
 imapge->image
 imbaress->embarrass
-imbed->embed
-imbedded->embedded
-imbedding->embedding
 imblance->imbalance
 imbrase->embrace
 imcoming->incoming


DimitriPapadopoulos:
While the OED contains an [`imbed`](https://www.oed.com/search/dictionary/?q=imbed) entry, it is missing from [SCOWL speller dictionaries](http://app.aspell.net/lookup?dict=en_US-large;words=imbed), with this mention:
>  [v] The word is considered a spelling variant. To promote consistent spelling, only one spelling of a word is generally included in a the smaller dictionary. The larger dictionary lets in common variants (level 1).

I tend to see [SCOWL (And Friends)](http://wordlist.aspell.net/) as the reference open source speller dictionary, perhaps an opinionated one. I would therefore suggest you request it is added under https://github.com/en-wl/wordlist/issues — I requested additions [myself](https://github.com/en-wl/wordlist/issues/created_by/DimitriPapadopoulos). However:
* The word list appears to be seldom updated (https://github.com/en-wl/wordlist/issues/394),
* A speller dictionary is not a complete dictionary. It will tend to recommend more common variants. As a non-native speaker, I find standardised spelling helps focus on the message instead of being distracted by rare variants.

In the meantime, I am not sure how to handle this entry:
- either move to the _rare_ dictionary,
- remove it altogether.

The [Google Ngram Viewer](https://books.google.com/ngrams/graph?content=embed%2Cimbed&year_start=1800&year_end=2019&corpus=en-2019) shows that `embed` is 60 times more common than `imbed` in contemporary English. Perhaps not enough to discard `imbed`. I really cannot suggest a proper threshold here.

oddhack:
Fair, and I already have it in an exception list for our docs. I will close this as it was a tentative suggestion to begin with, not something I have a strong opinion about.

oddhack:
N.b. SCOWL appears to be basically dead with no updates in 4 years. Fortunately languages evolve slowly :-)

kevina:
Hi,

I am the maintainer of SCOWL and noticed you mentioned this issue.

SCOWL is not dead.  A major update, with a new format is pending, but I do not have a ETA yet.  Once this is out I will work on going through the backlog of suggestions for new words.

This new format will provide SCOWL is a SQLite3 database format.  The new format will include variant information that might be helpful to you.  You can preview the new format at https://github.com/en-wl/wordlist/tree/v2.

Kevin

PR #3478, opened 2024-06-28
Update "Using a config file" README entry 

### LINKED ISSUE #3474: Still cannot specify custom + default dictionary in codespellrc file
I had thought this was addressed by #2767 / #2727 in codespell 2.3.0, but AFAICT this is still not supported. For example I can successfully do

> codespell --builtin clear,rare -D - -D my-dictionary

with the desired behavior. But trying to replicate this in my codespellrc via

> builtin = clear,rare  
> dictionary = -  
> dictionary = my-dictionary

gives a "ERROR: ill-formed config file: While reading from '/path/to/codespellrc' [line 6] option 'dictionary' in section 'codespell' already exists", while

> builtin = clear,rare  
> dictionary = -,my-dictionary

gives a "error: argument -D/--dictionary: expected one argument" error.

Am I misunderstanding something about how this is supposed to work, such that there is now a way to specify both builtin and custom dictionaries? Or is there still more TBD in this area? @larsoner you apparently closed  #2727 as completed, so if you can provide some insight, I'd appreciate it.

oddhack:
Document behavior with '-' in config file value fields

See https://github.com/codespell-project/codespell/issues/3474#issuecomment-2193878382

Closes #3474

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -178,6 +178,23 @@ which is read using Python's
 `configparser <https://docs.python.org/3/library/configparser.html#supported-ini-file-structure>`_.
 For example, comments are possible using ``;`` or ``#`` as the first character.
 
+Values in an INI file entry cannot start with a ``-`` character, so if you need to do this,
+structure your entries like this:
+
+.. code-block:: ini
+
+    [codespell]
+    dictionary = mydict,-
+    ignore-words = bar,-foo
+
+instead of these invalid entries:
+
+.. code-block:: ini
+
+    [codespell]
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

oddhack:
N.b. I suppose the question might come up of wanting to specify a single value which starts with '-' but that would probably require code changes and is out of scope here, which is just documenting the current behavior.

DimitriPapadopoulos:
Thank you @oddhack!

PR #3482, opened 2024-07-03
Add trusthworth(y|iness)->trustworth(y|iness) correction.

cfi-gb:
See e.g. (Note: grep.app results are low but GitHub code search shows way more):
- https://grep.app/search?q=trusthworthy
- https://grep.app/search?q=trusthworthiness
- https://github.com/search?q=trusthworthy&type=code
- https://github.com/search?q=trusthworthiness&type=code

PR #3483, opened 2024-07-04
Add thrustworth(y|iness)->trustworth(y|iness).

cfi-gb:
Similar to #3482 for another variant, see e.g. (Note: grep.app results are low but GitHub code search shows way more):
- https://grep.app/search?q=thrustworthy
- https://grep.app/search?q=thrustworthiness
- https://github.com/search?q=thrustworthy&type=code
- https://github.com/search?q=thrustworthiness&type=code

Note: Not sure if we should also add `thrust-worthy` and `thrust-worthiness` as correction alternatives?

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -58119,6 +58119,8 @@ thruogh->through
 thruoghout->throughout
 thruoghput->throughput
 thruout->throughout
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3487, opened 2024-07-12
Add spelling correction for separately.

cfi-gb:
See e.g. (For the first there are not that much hits but a few are there):
- https://github.com/search?q=sepparatelly&type=code
- https://grep.app/search?q=sepparatelly
- https://github.com/search?q=sepparately&type=code
- https://grep.app/search?q=sepparately


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3495, opened 2024-07-22
Improve config file documentation in README

### LINKED ISSUE #3493: README: unclarity to do with 'pyproject.toml'
The following is very hard to parse.

>  Codespell will also check in the current directory for a pyproject.toml (or a path can be specified via --toml <filename>) file, and the [tool.codespell] entry will be used, but only if the [tomli](https://pypi.org/project/tomli/) package is installed for versions of Python prior to 3.11.

The main problem is a scope-ambiguity (or at least something like that). For, what does the 'only if' clause qualify? For instance, if I have Python <3.11, can I use a .toml file _at all_?

DimitriPapadopoulos:
- Improve wording for TOML support requirements.
- The issue with config file entries starting with `-` applies to INI and TOML files. Therefore, reject at the end of the section.

Fixes #3493.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -161,7 +161,8 @@ Using a config file
 
 Command line options can also be specified in a config file.
 
-When running ``codespell``, it will check in the current directory for a file
+When running ``codespell``, it will check in the current directory for an
+`INI file <https://en.wikipedia.org/wiki/INI_file>`_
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:

DimitriPapadopoulos:

larsoner (approved):

PR #3510, opened 2024-08-07
Remove "blueish" correction

hadess:
blueish/bluish removal and options for a "sanity check" replacement

### DIFF
--- a/codespell_lib/data/dictionary_usage.txt
+++ b/codespell_lib/data/dictionary_usage.txt
@@ -2,7 +2,6 @@ black-hat->malicious actor, attacker,
 blackhat->malicious actor, attacker,
 blacklist->blocklist
 blacklists->blocklists
-blueish->bluish
 cripples->slows down, hinders, obstructs,
 crippling->attenuating, incapacitating,
 dummy-value->placeholder value


codespell_lib/data/dictionary_usage.txt
```diff
@@ -2,7 +2,6 @@ black-hat->malicious actor, attacker,
 blackhat->malicious actor, attacker,
 blacklist->blocklist
 blacklists->blocklists
-blueish->bluish
```
DimitriPapadopoulos, line 5: [OED](https://www.oed.com/search/dictionary/?q=blueish&tl=true) and other dictionaries show `blueish` as a valid variant of `bluish`.

Additionally, I can find `blueish` in both [en_US](http://app.aspell.net/lookup?dict=en_US;words=blueish) and [en_GB](http://app.aspell.net/lookup?dict=en_GB-ize;words=blueish) SCOWL lists.

codespell_lib/data/dictionary_usage.txt
```diff
@@ -25,6 +24,7 @@ mohammedans->muslims
 mom-test->user test, acceptance test, validation test, ease-of-use test, friendliness test, useability test,
 muhammadan->muslim
 muhammadans->muslims
+sanity->quick, initial, confidence, coherence, soundness, calibration, rationality,
```
DimitriPapadopoulos, line 27: I don't understand the purpose of this entry.

codespell_lib/data/dictionary_usage.txt
hadess: If this is an agreement to removing that incorrect "correction", can you please mark the conversation as resolved? An explanation for the removal is already in the commit message. Happy to add your explanations as well if you feel that's necessary.

codespell_lib/data/dictionary_usage.txt
hadess: "Sanity" isn't usually used in code outside of "sanity check", and while we don't have support for two word sequences (https://github.com/codespell-project/codespell/pull/1657), this is the best we can do.

DimitriPapadopoulos: Usage of codespell is not necessarily limited to source code. In such cases, as with `white list`, we don't check the isolated first word to avoid false positives.

DimitriPapadopoulos (approved):

PR #3218, opened 2023-11-14
don't show stacktrace from KeyboardInterrupt #3217

jcubic:
This fixes #3217 where when pressing CTRL+C codespell shows a useless stack trace.

Issue #3217, opened 2023-11-14
Codespell don't handle KeyboardInterrupt exception


jcubic:
This should be catched and the program should stop gracefully but instead show default stack trace:

```
^CTraceback (most recent call last):
  File "/home/kuba/.local/bin/codespell", line 8, in <module>
    sys.exit(_script_main())
             ^^^^^^^^^^^^^^
  File "/home/kuba/.local/lib/python3.12/site-packages/codespell_lib/_codespell.py", line 1017, in _script_main
    return main(*sys.argv[1:])
           ^^^^^^^^^^^^^^^^^^^
  File "/home/kuba/.local/lib/python3.12/site-packages/codespell_lib/_codespell.py", line 1185, in main
    bad_count += parse_file(
                 ^^^^^^^^^^^
  File "/home/kuba/.local/lib/python3.12/site-packages/codespell_lib/_codespell.py", line 903, in parse_file
    check_matches = extract_words_iter(line, word_regex, ignore_word_regex)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kuba/.local/lib/python3.12/site-packages/codespell_lib/_codespell.py", line 793, in extract_words_iter
    return list(word_regex.finditer(_ignore_word_sub(text, ignore_word_regex)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyboardInterrupt
```

There is no need to show `KeyboardInterrupt` exception stack trace.

PR #3511, opened 2024-08-08
Handle CTRL+C by showing a better message

mwtoews:
This tidies-up messages while using interactive codespell, if the user decides to cancel using CTRL+C.

For example, before this PR:
```console
$ codespell -i1 file.txt
hte
	hte ==> the (Y/n) ^CTraceback (most recent call last):
  File "/tmp/py310/bin/codespell", line 8, in <module>
    sys.exit(_script_main())
  File "/home/mtoews/src/codespell/codespell_lib/_codespell.py", line 1102, in _script_main
    return main(*sys.argv[1:])
  File "/home/mtoews/src/codespell/codespell_lib/_codespell.py", line 1324, in main
    bad_count += parse_file(
  File "/home/mtoews/src/codespell/codespell_lib/_codespell.py", line 1007, in parse_file
    fix, fixword = ask_for_word_fix(
  File "/home/mtoews/src/codespell/codespell_lib/_codespell.py", line 775, in ask_for_word_fix
    r = sys.stdin.readline().strip().upper()
KeyboardInterrupt

$ echo $?
130
```
and after this PR:
```console
$ codespell -i1 file.txt
hte
	hte ==> the (Y/n) ^C
Cancelling '/tmp/py310/bin/codespell'
$ echo $?
130
```

And the same behaviour if a user starts codespell with `python -m codespell_lib ...`

---

Currently there is no test for this. It's probably possible, but not easy for me to construct. If this is important to include, I'd appreciate some tips on how to enable testing [`SIGINT`](https://docs.python.org/3/library/signal.html#signal.SIGINT).

### DIFF
--- a/codespell_lib/__main__.py
+++ b/codespell_lib/__main__.py
@@ -3,7 +3,4 @@ import sys
 from ._codespell import _script_main
 
 if __name__ == "__main__":
-    try:
-        sys.exit(_script_main())
-    except KeyboardInterrupt:
-        sys.exit(f"\ncancelling '{sys.argv[0]}'\n")
+    sys.exit(_script_main())
--- a/codespell_lib/__main__.py
+++ b/codespell_lib/__main__.py
@@ -6,4 +6,4 @@ if __name__ == "__main__":
     try:
         sys.exit(_script_main())
     except KeyboardInterrupt:
-        pass
+        sys.exit(f"\ncancelling '{sys.argv[0]}'\n")
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -1103,7 +1103,8 @@ def _script_main() -> int:
         return main(*sys.argv[1:])
     except KeyboardInterrupt:
         # User has typed CTRL+C
-        sys.stderr.write(f"\nCancelling '{sys.argv[0]}'\n")
+        sys.stdout.write("\n")
+        sys.stderr.write(f"Cancelling '{sys.argv[0]}'\n")
         return 130
 
 
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -1099,7 +1099,11 @@ def flatten_clean_comma_separated_arguments(
 
 def _script_main() -> int:
     """Wrap to main() for setuptools."""
-    return main(*sys.argv[1:])
+    try:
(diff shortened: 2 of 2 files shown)


codespell_lib/__main__.py
```diff
@@ -6,4 +6,4 @@
     try:
         sys.exit(_script_main())
     except KeyboardInterrupt:
-        pass
+        sys.exit(f"\ncancelling '{sys.argv[0]}'\n")
```
luzpaz, line 9: Is there a reason why `cancelling` isn't capitalized?

mwtoews: Sure, that's fine to amend (I'll do this later, in case other changes are requested).

mwtoews:
Another arbitrary decision that can be changed is the return code. I've set this as 1 for now. But it could also be restored to 130 (the default for KeyboardInterruptor) or 0 for "no error". Opinions welcome.

mwtoews:
See related xref #3217 #3218. This PR moves the try/catch to `_script_main` where is is more generically caught.

codespell_lib/_codespell.py
```diff
@@ -1099,7 +1099,13 @@ def flatten_clean_comma_separated_arguments(
 
 def _script_main() -> int:
     """Wrap to main() for setuptools."""
-    return main(*sys.argv[1:])
+    try:
+        return main(*sys.argv[1:])
+    except KeyboardInterrupt:
+        # User has typed CTRL+C
+        sys.stdout.write("\n")
+        sys.stderr.write(f"Cancelling '{sys.argv[0]}'\n")
```
DimitriPapadopoulos, line 1107: Is a message really needed? Most commands I use just exit upon Ctrl+C, without any message.

luzpaz: I'm +1 for the message because it is annoying to see superfluous output from ctrl+c

DimitriPapadopoulos: @mwtoews Could you can remove the messages and rebase? We'll try to merge despite the Codecov warning.

DimitriPapadopoulos (approved):

luzpaz (approved):

DimitriPapadopoulos:
Thank you @mwtoews!

PR #3534, opened 2024-09-07
Add bellow -> below suggestion

algonell:
Inspired by Matthew Wilcox as part of a Kernel newbies task.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -9329,6 +9329,7 @@ belived->believed, beloved,
 belives->believes, beliefs,
 beliving->believing
 belligerant->belligerent
+bellow->below
 bellweather->bellwether
 bellweathers->bellwethers
 belog->belong


luzpaz:
The problem is that it's a valid word and that's why it's already been added to the rare dictionary:
https://github.com/codespell-project/codespell/blob/134c257a42413dcb390bb53e809f724b83a4e922/codespell_lib/data/dictionary_rare.txt#L22

The rare dictionary is used be default in codespell `The default is 'clear,rare'.`

Source:  `codespell --help`

```shell
  --builtin BUILTIN-LIST
                        comma-separated list of builtin dictionaries to include (when "-D -" or no "-D" is passed). Current options are:
                        - 'rare' for rare (but valid) words that are likely to be errors
                        - 'informal' for making informal words more formal
                        - 'usage' for replacing phrasing with recommended terms
                        - 'code' for words from code and/or mathematics that are likely to be typos in other contexts (such as uint)
                        - 'names' for valid proper names that might be typos
                        - 'en-GB_to_en-US' for corrections from en-GB to en-US
                        The default is 'clear,rare'.
```

Closing.

algonell:
My bad, should have `git grep` it before.

Thank you for the detailed explanation!

PR #3535, opened 2024-09-08
More typos

DimitriPapadopoulos:


codespell_lib/data/dictionary.txt
```diff
@@ -46733,6 +46801,8 @@ quantititive->quantitative
 quantitity->quantity
 quantitiy->quantity
 quarantaine->quarantine
+quarantee->guarantee
```
luzpaz, line 46804: very close to 'quarantine', could be a mistake for a non-english speaker perhaps ?

codespell_lib/data/dictionary_rare.txt
```diff
@@ -232,6 +233,10 @@ readding->re-adding, reading,
 readds->re-adds, reads,
 ream->stream
 recuse->recurse
+reduct->deduct, reduce,
```
luzpaz, line 236: also 'redact'

codespell_lib/data/dictionary.txt
DimitriPapadopoulos: I haven't found such a typo out there:
https://grep.app/search?q=quarantee

codespell_lib/data/dictionary_rare.txt
DimitriPapadopoulos: Added.

codespell_lib/data/dictionary.txt
luzpaz: just out of curiosity i through it into google:

Screenshot_20240909_071018

DimitriPapadopoulos: Such dictionary suggestions are based on a distance metric between words, but that distance metric doesn't necessarily reflect the probability of typos in practice. Anyway, I have added `quarantine`.

luzpaz (approved):

PR #3544, opened 2024-09-17
Add spelling correction for appliance and variants.

cfi-gb:
See e.g.
- https://grep.app/search?q=Apliance&words=true
- https://grep.app/search?q=applicance
- https://github.com/search?q=%22apliances%22&type=code (Has many unrelated hits but e.g. `apliances_path` can be seen there)


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3553, opened 2024-09-26
Add spelling correction for revert and variants.

cfi-gb:
See e.g.:
- https://github.com/search?q=referted&type=code
- https://grep.app/search?q=referted
- https://github.com/search?q=referting&type=code
- https://grep.app/search?q=refert&words=true&filter[lang][0]=Python (Only Python included as the other findings seems to include mostly "dummy" text like e.g. `Refert tamen, quo modo`)

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -48456,9 +48456,12 @@ referrrer->referrer
 referrrers->referrers
 referrring->referring
 referrs->refers
+refert->revert, refer, refers,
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Thank you @cfi-gb.

DimitriPapadopoulos:
@cfi-gb Could you please rebase so CI fixes are taken into account?

cfi-gb:
@DimitriPapadopoulos Sure and already done, thanks a lot for the ping :+1:

PR #3558, opened 2024-10-02
Update ruff settings

DimitriPapadopoulos:
@larsoner Could you review this PR?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -137,6 +137,8 @@ select = [
 ignore = [
     "ANN101",
     "B904",
+    "PLR0914",
+    "PLR6201",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -157,6 +134,36 @@ select = [
...
+    "D300",
+    "Q000",
+    "Q001",
+    "Q002",
+    "Q003",
+    "COM812",
+    "COM819",
+    "ISC001",
+    "ISC002",
+]
```
larsoner, line 166: Makes me think we should add `toml-sort` pre-commit at some point

larsoner (approved):

PR #3568, opened 2024-10-15
Add zarr as a fix for zar.

yarikoptic:
Zarr is a new community driven data format, discover more https://en.wikipedia.org/wiki/Zarr_(data_format)
and I ran into this while fixing a Zarr related project

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -63859,7 +63859,7 @@ yuo->you
 yuor->your
 yur->your
 yymmetrical->symmetrical
-zar->czar
+zar->czar, Zarr,
 zars->czars
 zeebra->zebra
 zeebras->zebras


codespell_lib/data/dictionary.txt
```diff
@@ -63596,8 +63596,8 @@ yuo->you
 yuor->your
 yur->your
 yymmetrical->symmetrical
-zar->czar
-zars->czars
+zar->czar, zarr
+zars->czars, zarrs
```
DimitriPapadopoulos, line 63600: `Zarr` is OK, but not sure about the plural `Zarrs`.

Don't forget the trailing comma.

yarikoptic: sorry, what trailing comma?

yarikoptic: Let's remove

```suggestion
```

yarikoptic: I guess not to leave it behind... ok let's do:

```suggestion
zars->czars
```

DimitriPapadopoulos (approved):

PR #3572, opened 2024-10-23
Add correction for seens->seems, seen, scenes,

cfi-gb:
See e.g. https://grep.app/search?q=seens%20to&words=true

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -52328,6 +52328,7 @@ seemd->seemed
 seemes->seems
 seemless->seamless
 seemlessly->seamlessly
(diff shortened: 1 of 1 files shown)


luzpaz:
Could also be `scenes`

cfi-gb:
> Could also be `scenes`

Mhhh, not fully sure because you would need to do two typos here:

s**c**en**e**s -> seens

means you would need to have mixed up `c` with `e` at the beginning and dropped an `e` at the end or do i miss something?

cfi-gb:
Makes sense and thanks a lot for clarification, updated / adjusted the PR accordingly.

luzpaz (approved):

PR #3576, opened 2024-11-07
Add forach->foreach, orach, spelling correction

cfi-gb:
See e.g. https://grep.app/search?q=forach&words=true

~~Note: Wasn't sure about the placement so put it into the dictionary_code.txt first but could be also moved into the more generic one.~~

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -28468,6 +28468,7 @@ footprinst->footprints
 foound->found
 foppy->floppy
 foppys->floppies
+forach->foreach, orach,
 foraign->foreign
 foraigner->foreigner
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Actually, `dictionary_code.txt` is for words that are **not** typos in the context of code. For example, this would be a good candidate:
```
foreach-> for each
```
Therefore I would indeed move to the generic dictionary, perhaps as:
```
forach-> foreach, orach,
```

cfi-gb:
Thanks a lot for the feedback / hint, adjusted the PR accordingly :+1:

DimitriPapadopoulos (approved):

DimitriPapadopoulos (approved):

PR #3578, opened 2024-11-12
Add spelling correction for leadin.

cfi-gb:
- See e.g. https://grep.app/search?q=leadin&words=true (a few are variables like `leadIn` but that IMHO shouldn't matter)
- There is already the `misleadin->misleading, mislead in,` in the dictionary


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3580, opened 2024-11-15
Minor typo fix in README 

DanielYang59:


README.rst
```diff
@@ -67,7 +67,7 @@ The ``-I`` flag can be used for a list of certain words to allow that are in the
 
     codespell -L word1,word2,word3,word4
 
-The ``-L`` flag can be used to allow certain words that are comma-separated placed immediately after it.  See `Ignoring Words`_ for more details.
```
DanielYang59, line 70: `words that are comma-separated placed immediately after it` seems a bit confusing to me, I assume it means "a sequence of words that are comma-separated and don't have spaces around commas"?

larsoner:
Thanks @DanielYang59

peternewman:
Did you test this behaviour @DanielYang59 as I don't believe it's correct. IIRC the readme is effectively trying to say (and codespell's behaviour is):

> When ignoring false positives, note that spelling errors are *case-insensitive* but words to ignore are *case-sensitive* based on how they appear in the dictionary. For example, the dictionary entry ``wrod`` will also match the typo ``Wrod``, but to ignore either version you must pass ``wrod`` (because that is how it is listed in the dictionary).

larsoner:
Oops I did not actually check... PR welcome to fix

DanielYang59:
@peternewman @larsoner Apologize I didn't really check the behavior, I thought it's simply a typo.

I could confirm @peternewman  is correct:
```
wrod = 0  # codespell:ignore wrod
wrod = 0  # codespell:ignore Wrod  # Marked as spell error

Wrod = 1  # codespell:ignore wrod
Wrod = 1  # codespell:ignore Wrod  # Marked as spell error
```

> to ignore either version you must pass wrod (because that is how it is listed in the dictionary).

I now understand what the docs are trying to convey (the ignore pattern from config is case sensitive and should match its corresponding entry from the dictionary).

---

However I still found the documentation confusing to me:

> When ignoring false positives, note that spelling errors are case-insensitive but words to ignore are case-sensitive.

1. What is "spelling error" (the error found from the file to be checked? ) vs "word to ignore" (the ignore pattern given by config, which should match its dictionary entry?) So if I understand correctly, this is basically "when giving a ignore pattern in config, the pattern is case-sensitive and should match its entry from the dictionary", and once the ignore pattern is in-place, the case of the file to be checked doesn't matter.

> because that is how it is listed in the dictionary

2. I believe this should be mentioned in the README, as I'm a new user, and I don't know this behaviour at all after reading the `Ignoring words` section.

3. Also I found it a bit confusing to have different behaviour for different words, however I just had a [quick look at the dictionary](https://github.com/codespell-project/codespell/tree/main/codespell_lib/data), it appears to me most (if not all) entries are lowercase? If that's the case, can we change the README to "the word to ignore should be lowercase" instead of "the word to ignore should be case-sensitive", which might lead to confusion like this PR (I thought to ignore `Wrod` you have to give `Wrod`)

PR #3586, opened 2024-11-22
Add spelling correction for "agos".

cfi-gb:
Doesn't happen that much if looking at https://grep.app/search?q=%20agos%20&words=true but found e.g. this:

> 1 week agos is invalid

and did this typo recently as well.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -2800,6 +2800,7 @@ agonsticism->agnosticism
 agorithm->algorithm
 agorithmic->algorithmic
 agorithms->algorithms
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -2800,6 +2800,7 @@ agonsticism->agnosticism
 agorithm->algorithm
 agorithmic->algorithmic
 agorithms->algorithms
+agos->ago, ages, Lagos,
```
DimitriPapadopoulos, line 2803: Perhaps also `egos`?

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3587, opened 2024-11-22
Do pass posargs to tox invocation

yarikoptic:
So I could do

    tox -- -k test_basic --pdb -s -v

to limit execution and make it more verbose


DimitriPapadopoulos:
I wasn't aware of `{posargs}` which is documented here:
[Using Tox](https://python-refs.readthedocs.io/en/latest/testing_tox.html)

Requires a rebase though.

DimitriPapadopoulos (approved):

PR #3588, opened 2024-11-22
Do not allow for a word to start with punctuation symbols

### LINKED ISSUE #3305: Detection of string delimiters .
Currently, words in double quoted strings are spellchecked, but not strings with simple quotes.

My suggestion is to add a feature that would let codespell determine what a string delimiter is on a line by line basis, possibly limited to certain file patterns (extensions).

The detection would be on a line by line basis.

I understant that this could be somewhat tricky.

One way to implement it would be to have a regex that finds strings and replaces the delimiters of the string with spaces before looking for words.

yarikoptic:
The inspired for me to look use case
- https://github.com/INTERSECT-SDK/python-sdk/pull/19/files/33da9ff31d6162caa0dfc1a1155f321e6d68b1cc#diff-10380fd6e5ecb84c1ae11e135982739946c5aff1a50499378db397cf5034f54e

And then I found the issue this
- Close #3305

Although may be I am missing the use-cases/problems @DimitriPapadopoulos and @mdeweerd discussed back then

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -53,7 +53,7 @@ from ._version import (  # type: ignore[import-not-found]
     __version__ as VERSION,  # noqa: N812
 )
 
-word_regex_def = r"[\w\-'’]+"  # noqa: RUF001
+word_regex_def = r"\w[\w\-'’]*"  # noqa: RUF001
 # While we want to treat characters like ( or " as okay for a starting break,
 # these may occur unescaped in URIs, and so we are more restrictive on the
 # endpoint.  Emails are more restrictive, so the endpoint remains flexible.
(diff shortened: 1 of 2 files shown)


larsoner:
I haven't looked deeply but am I right that this single-line comment:
```
# blah blah 'blah blah' blah
```
would get handled differently from this multi-line one
```
# blah blah 'blah
# blah' blah
```
? To me these are essentially the same code comment so it's weird that codespell would treat them differently.

It's even a bit weird to try to keep track of the quotation level in any way as it seems brittle, especially given people can forget start and end quotes from time to time

yarikoptic:
since no space should be in "word regex", we should be robust to such examples -- those should be separate words (on one line or not)

PR #3590, opened 2024-11-23
Add replacements for "ra"

TheGiraffe3:
But then again, "Ra" is the name of the ancient Egyptian god; should that also be added as a possibility?

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -47170,6 +47170,7 @@ qwesant->croissant
 qwesants->croissants
 qwessant->croissant
 qwessants->croissants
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (commented):
How about moving this entry to [`dictionary_names.txt`](https://github.com/codespell-project/codespell/blob/main/codespell_lib/data/dictionary_names.txt)?

DimitriPapadopoulos:
Also, `Ra` is the symbol for radium and an abbreviation (medical terms, rear admiral, resident assistant, Royal Academy, etc.).

TheGiraffe3:
Hmm, if it's an abbreviation, then it's probably not good to tell codespell it's a typo.

PR #3596, opened 2024-11-28
Handle words surrounded by single quotes

### LINKED ISSUE #3305: Detection of string delimiters .

mwtoews:
This resolves finding words enclosed within single quotes, e.g.:

    This is an 'errror'

Closes #3305

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -759,13 +759,20 @@ def ask_for_word_fix(
     colors: TermColors,
 ) -> Tuple[bool, str]:
     wrongword = match.group()
+    match_start = match.start()
+    match_end = match.end()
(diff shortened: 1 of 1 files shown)


mwtoews:
Currently in draft as I'm not sure if this is the best solution. It seems to work, but open to feedback.

Also, it needs to have unit tests prepared. There are a few examples in the linked issue to work with.

PR #3597, opened 2024-11-29
Add replacements for complasance and complisance

TheGiraffe3:
Haven't been able to find this with grep, but seems like it could be easy to do.


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @TheGiraffe3!

PR #3600, opened 2024-12-02
Add variations for correction: reurn->return

jdufresne:
Found the misspelling "reurns" in a personal project, as in: "It reurns false ..."

The mistake could also be a typo for "rerun", so include that as a potential fix too.


luzpaz:
It could also be `rerun`

jdufresne:
> It could also be `rerun`

Good call. Thanks for the suggestion. I have included "rerun" as an alternate correction.

luzpaz (approved):

TheGiraffe3 (approved):

PR #3601, opened 2024-12-02
Add spelling correction for various variants of everything.

cfi-gb:
See e.g.:
- https://grep.app/search?q=Eerything&words=true
- https://grep.app/search?q=eveerything&words=true

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -23599,6 +23599,7 @@ eeeprom->EEPROM
 eeger->eager
 eegerly->eagerly
 eejus->aegis
+eerything->everything, eerie thing,
 eescription->description
 eevery->every
 eeverything->everything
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -23592,6 +23592,7 @@ eeeprom->EEPROM
 eeger->eager
 eegerly->eagerly
 eejus->aegis
+eerything->everything
```
TheGiraffe3, line 23595: Unlikely, but it's possible that they meant "eerie thing".
```suggestion
eerything->eerie thing, everything,
```

TheGiraffe3 (commented):
At first I thought this would be a huge pull request because it changes everything 😉

codespell_lib/data/dictionary.txt
cfi-gb: Nice, thanks a lot. Wasn't even aware that such a word `eerie` exists :-)

Updated with the last force push but placed `everything` first because this seems to be more likely.

TheGiraffe3 (approved):

DimitriPapadopoulos (approved):

PR #3603, opened 2024-12-04
Add "sems->seems, stems, semis, sens, seams," correction

cfi-gb:
See e.g. https://grep.app/search?q=it%20sems%20&words=true

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -52590,6 +52590,7 @@ sempaphores->semaphores
 semphore->semaphore
 semphores->semaphores
 sempphore->semaphore
+sems->seems, stems, semis, sens, seams,
 senaireo->scenario
 senaireos->scenarios
 senaphore->semaphore


TheGiraffe3 (commented):
"SEMS" is an abbreviation for "scanning electron microscope", so it may not be necessarily wrong. But it will be the majority of the time...

cfi-gb:
Some recent related comments from e.g. #2858 or #2804

> Any (or almost any) typo could be seen as a valid abbreviation.

or

> Don't worry, any typo could be a name (of a person, of a place, of a fictive character or place). We wouldn't have a dictionary of possible typos if we were to take that into account.

I can remember @peternewman writing something similar in the past but can't find that anymore.

codespell_lib/data/dictionary.txt
```diff
@@ -52588,6 +52588,7 @@ sempaphores->semaphores
 semphore->semaphore
 semphores->semaphores
 sempphore->semaphore
+sems->seems, stems, semis, sens,
```
TheGiraffe3, line 52591: Alright, that makes sense. Just one more suggestion.
```suggestion
sems->seems, sees, stems, semis, sens,
```

cfi-gb: Nice, thanks a lot for the additional suggestion :+1:

DimitriPapadopoulos: And perhaps `seams` as well.

TheGiraffe3: Wow, lots of words sems could be a typo of.

cfi-gb: Added as well, thanks a lot again for the suggestions :+1:

TheGiraffe3 (approved):

PR #3604, opened 2024-12-06
Add distinghish->distinguish and variations

fishilico:
It was found for example in https://github.com/v8/v8/blob/d5424dbd69332f7483dde5e900dde3d43f2c35f0/src/compiler/turboshaft/operations.h#L3399 and https://github.com/NatronGitHub/Natron/blob/8326245bf9b646c54ad6171aef5a8ccb292500e7/libs/Eigen3/Eigen/src/Core/GeneralProduct.h#L135 and https://github.com/ligolang/ligo/blob/c61d549ef9ed7dfa427fcc95aa80ab718fb97ab8/gitlab-pages/docs/signatures/extending.md?plain=1#L68

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -22516,6 +22516,11 @@ distince->distance
 distinced->distanced
 distinces->distances
 distincing->distancing
+distinghish->distinguish
+distinghishable->distinguishable
+distinghished->distinguished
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @fishilico.

PR #1330, opened 2019-11-09
Import of spell fixer from Linux Kernel

Gelma:
From commit:
00aff6836241ae5654895dcea10e6d4fc5878ca6

import of
scripts/spelling.txt

PR #3606, opened 2024-12-09
remove misformed->malformed

klensy:
This was removed previously in https://github.com/codespell-project/codespell/pull/555, but later added back in https://github.com/codespell-project/codespell/pull/1330.

Naive(probably often) suggestion, but there will be nice to have list of corrections that was removed from dict, to prevent oscillation for adding\removing words.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -38192,7 +38192,6 @@ miselaneous->miscellaneous
 miselaneously->miscellaneously
 misellaneous->miscellaneous
 misellaneously->miscellaneously
-misformed->malformed
 misfourtunes->misfortunes
 misile->missile
 mising->missing


TheGiraffe3 (approved):
Naive as it may be (I don't know) I believe that suggestion is a good one.

DimitriPapadopoulos:
https://www.oed.com/search/dictionary/?q=misformed

DimitriPapadopoulos (dismissed):

DimitriPapadopoulos (commented):
Indeed, `misformed` is a valid word according to the [OED](https://www.oed.com/search/dictionary/?q=misformed). However, it is seldom used nowadays according to the [Google Books Ngram Viewer](https://books.google.com/ngrams/graph?content=misformed%2Cmalformed&year_start=1800&year_end=2022&corpus=en), which is probably why it is in the list of typos of the Linux kernel and missing from [SCOWL (And Friends)](http://app.aspell.net/lookup?dict=en_US-large;words=misformed).

Perhaps it should be moved to the _rare_ dictionary.

PR #3619, opened 2025-01-18
Run pytest GitHub Action on an ARM processor

cclauss:
Speed up a long running GitHub Action by running it on an ARM processor.
[Standard GitHub-hosted runners for public repositories](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) --> `os: ubuntu-24.04-arm`

Compare with other runtimes…
https://github.com/codespell-project/codespell/actions/workflows/codespell-private.yml


larsoner (approved):

larsoner:
Thanks @cclauss

PR #3631, opened 2025-02-03
Revert "Add rare typo `lien->line`"

### LINKED ISSUE #3630: Consider reverting `lien->line`
https://github.com/codespell-project/codespell/pull/3460 added a `lien->line` typo, but lien is a valid word - [definition](https://www.google.com/search?q=define+lien).

In our case it is used fairly commonly within our terraform code in relation to the [google_resource_manager_lien](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/resource_manager_lien) resource.  This resource configured a lien within Google Cloud Platform which prevents the deletion of a project or resource.

It's used within multiple repos and whilst I could add ignored words everywhere, `lien` is a dictionary word and we won't be the only people using the google_resource_manager_lien resource in terraform.

nikolaik:
Since `lien` is a word frequently used when working with the Google
Terraform Provider, it feels wrong to treat it as a typo.

Humbly suggesting to revert that change

This reverts commit 536ccb530ef9f955d17932e201a0d0f4e27bf2f9.

Fixes #3630

### DIFF
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -154,8 +154,6 @@ lamdas->lambdas
 leaded->led, lead,
 leas->least, lease,
 lief->leaf, life,
-lien->line
-liens->lines
 lightening->lightning, lighting,
 loafing->loading
 lod->load, loud, lode,


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Note that this typo is in the _rare_ dictionary. Perhaps the _rare_ dictionary should not be selected by default.

DimitriPapadopoulos:
@nikolaik It seems to me that `lien` is **not** related to code, but a specific domain.

nikolaik:
> @peternewman @nikolaik It seems to me that `lien` is **not** related to programming, but to a specific domain.

I agree with you @DimitriPapadopoulos but no strong opinion, so open to go either way

peternewman:
> @peternewman @nikolaik It seems to me that `lien` is **not** related to programming, but to a specific domain.

When you say specific domain do you mean Google Terraform? Isn't that the same as say:
https://github.com/codespell-project/codespell/blob/4644f584fe6d9b98847064889a95c2bd03e7917c/codespell_lib/data/dictionary_code.txt#L52

Which is only relevant to Assembler (AFAIK), to pick probably one of the more obscure ones?

rjmoseley:
From my perspective it's not particularly related to programming, it's a legal term that just so happens to be used by Google Cloud Platform. Whilst it's that usage that has resulted in this debate, fundamentally it's just a word, and one I (native British English speaker) was familiar with long before I came across it's usage in GCP.

nikolaik:
From what I'm hearing in this discussion and #3630 is that adding well defined word with current use as a misspelling would not make sense and do more harm than good. Originally added in #3460 as a misspelling with the justification "Found in Emacs."

I changed this PR back back to a plain revert, since that now seems like the best outcome to me.

DimitriPapadopoulos (approved):

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
In the long term, I would rather keep rare words in the _rare_ dictionary, that's what it's for. Perhaps the _rare_ dictionary shouldn't be selected by default.

I'll settle for removing this entry for now. Thank you @nikolaik.

Just noticed another possible alternative. From the README:

> 3. Entry with one word, but with automatic fix disabled:
>    
>    ```
>    clas->class, disabled because of name clash in c++
>    ```
> Note that there isn't a comma at the end of the line. The last argument is treated as the reason why a suggestion cannot be automatically applied.

Perhaps we should disable automatic fixes by adding a comment to all or part of the words in the _rare_ dictionary. Example:
```console
$ cat DICTIONARY.txt 
lien->line, actually a word in the OED and SCOWL (And Friends)
$ 
$ cat test.txt 
lien
$ 
$ codespell -w -D DICTIONARY.txt test.txt 
test.txt:1: lien ==> line  | actually a word in the oed and scowl (and friends)
$ 
$ codespell -D DICTIONARY.txt test.txt 
test.txt:1: lien ==> line  | actually a word in the oed and scowl (and friends)
$ 
```

nikolaik:
Thanks Dimitri!

nikolaik:
Autofixing cases which rely on a person using their discretion, where there is ambiguity (dictionary definition exists) sounds wrong to me.

DimitriPapadopoulos:
> Autofixing cases which rely on a person using their discretion, where there is ambiguity (dictionary definition exists) sounds wrong to me.

So would you agree to reintroducing `lien` in the _rare_ dictionary, but with autofixing disabled?

DimitriPapadopoulos:
So removing the _rare_ dictionary from the default dictionaries would be the best option, wouldn't it?

nikolaik:
I can't say, it depends a lot on the contents of the _rare_ dictionary, but it might

DimitriPapadopoulos:
> 'rare' for rare (but valid) words that are likely to be errors

PR #3645, opened 2025-02-20
"ane" could have been "one"

yarikoptic:
Ran into within https://github.com/napari/docs/pull/587

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -4426,7 +4426,7 @@ androis->androids
 andthe->and the
 andtoid->android
 andtoids->androids
-ane->and
+ane->and, one,
 anecdatal->anecdotal
 anecdatally->anecdotally
 anecdate->anecdote


yarikoptic:
> `error ane: multiple corrections must end with trailing ","`

I wonder if there is some rationale for such formatting requirement?

DimitriPapadopoulos:
> Note the last comma! You need to use it, otherwise the last suggestion will be discarded (see below for why).

> Note that there isn't a comma at the end of the line. The last argument is treated as the reason why a suggestion cannot be automatically applied.

yarikoptic:
sorry @DimitriPapadopoulos , must be a Friday but I am still missing while trailing `,` is needed.

DimitriPapadopoulos (approved):

yarikoptic:
I can see that as a "feature" for a single choice, but with multiple choices - none of them can  be automatically applied, so semantic of the trailing `,` is redundant/not needed, right?

DimitriPapadopoulos:
Even with multiple choices, whatever's after the last coma is a comment. It's probably too late to change this.

PR #3646, opened 2025-02-20
Fix  "ti" as "it"

yarikoptic:
Not sure about this one though -- since so short, might raise alarm too often in false positive cases etc.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -58939,6 +58939,7 @@ thursdey->Thursday
 thurver->further
 thw->the, thaw,
 thyat->that
+ti->it
 tich->thick, tick, titch, stitch,
 tichened->thickened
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Indeed, `Ti` is the symbol for titanium, `TI` is the inversion time in MRI, `ti` is an alternative spelling of the note `si`, it's even a tree:
* [OED](https://www.oed.com/search/dictionary/?q=ti)
* [SCOWL (And Friends)](http://app.aspell.net/lookup?dict=en_GB-large;words=ti)

At the very least, it should be moved to the _rare_ dictionary.

DimitriPapadopoulos (changes requested):
Move to _rare_ dictionary, and add comment such as "_actually a word in OED and SCOWL (And Friends)_".

PR #3655, opened 2025-02-24
Add casue->case and a variant

TheGiraffe3:


codespell_lib/data/dictionary.txt
```diff
@@ -11445,9 +11445,9 @@ cassims->chasms
 cassm->chasm
 cassms->chasms
 cassowarry->cassowary
-casue->cause
+casue->case, cause,
```
DimitriPapadopoulos, line 11448: The probability of `casue->case` is rather low on common keyboards. Out of curiosity, have you actually seen this typo out there?

TheGiraffe3: I don't believe I have.

luzpaz: yea, i haven't encountered that, AFAICT

PR #3665, opened 2025-03-10
agreegate, lesda, realod, colouer

mdeweerd:
A few suggestions.

FYI: I considerd that a person mistyping "colouer" meant to write "colour" and not "color".  And also in this case, it was a british person that made the typo.


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @mdeweerd.

PR #3666, opened 2025-03-12
Update pre-commit version in documentation

prchoward:
Closes: https://github.com/codespell-project/codespell/issues/3662- Update user documentation 

Update the version of pre-commit in the documentation from v2.2.4 to v2.4.1 to align with current standards. The previous change was 2 years ago.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -237,7 +237,7 @@ codespell also works with `pre-commit <https://pre-commit.com/>`_, using
 .. code-block:: yaml
 
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.4
+    rev: v2.4.1
     hooks:
     - id: codespell
 
@@ -246,7 +246,7 @@ If one configures codespell using the `pyproject.toml` file instead use:
 .. code-block:: yaml
 
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.4
+    rev: v2.4.1
     hooks:
     - id: codespell
       additional_dependencies:


prchoward:
This is my first opensource contribution so feedback on contribution etiquette or standards is greatly appreciated.

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
While it doesn't sound reasonable to automate such updates, we could add a comment in the pre-commit config file. It would remind us to update the README together with the pre-commit file:
https://github.com/codespell-project/codespell/blob/c9cea7c08e7c2443c1f016df45f59b5bfc4b6470/.pre-commit-config.yaml#L70-L71

```diff
-   rev: v2.4.1 
+   rev: v2.4.1  # update README.rst accordingly
```

DimitriPapadopoulos:
@larsoner Same problem again with stalled checks displaying "_Expected — Waiting for status to be reported_". The strange thing is that these stalled checks are sort of duplicates with different full names: the stalled check displays only the the job name `3.8`, while the successful check displays the full name combining the action name `Pytest` and the job name `3.8`.

According to [_Github actions: Stuck in "Expected — Waiting for status to be reported"_](https://github.com/orgs/community/discussions?discussions_q=Expected+%E2%80%94+Waiting+for+status+to+be+reported), these failures could be caused by `cancel-in-progress`:
https://github.com/codespell-project/codespell/blob/c9cea7c08e7c2443c1f016df45f59b5bfc4b6470/.github/workflows/codespell-private.yml#L8
There doesn't seem to be a solution other than removing `cancel-in-progress`.

Unless of course there is a discrepancy between branch protection rules and actual job names, but that would result in consistent errors for each PR, wouldn't it?

larsoner:
The cancel-in-progress would be weird... the jobs actually ran and had the correct names. I have a different more hacky solution that I think should work, I'll open a PR shortly

PR #3671, opened 2025-03-17
Add spelling corrections for vulnerabilities and variants.

cfi-gb:
This can happen if some one e.g. first wants to write `vulnerability` but change it to `vulnerabilities` then wrongly.

See e.g.:
- https://github.com/search?q=vulnerabilityies&type=code
- https://github.com/search?q=%22vulnerabilitiesy%22&type=code (the "other way around" to the previous but none found so far, still adding it)


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3675, opened 2025-03-24
New option: --per-file-ignores

### LINKED ISSUE #3673: New setting: `per-file-ignores` to ignores specific words into specific files
I recently discorve codespell and it was a great one, particularly in so far as I am not fluent in English.

However, after experimenting on various projects, I strongly suggest adding a feature `per-file-ignores` to ignores specific words into specific files. This would give finer control over the errors returned by codespell, as we sometimes use some variable names which are flag by codespell, but we don't want to exclude them globally they could really be a legitimate typo error in others context.

The toml configuration would look like this:
```toml
[codespell]
[codespell.per-file-ignores]
"file1" = [
  "word1",
  "word1",
]
"file2" = [
  "word3",
  "word4",
]
```

The inline/command format will be a bit more like that:
```
codespell  --per-file-ignores "file1:[word1,word2],file2:[word3,word4]"
```

un-pogaz:
Close #3673

This implementation use `nargs=2` to pass the file glob and the words list with a pair of arguments, rather than try to extract the data from a single complexe string. Also use `action="append"` so that several `--per-file-ignores` can be provided in one comand. This need to be a bit tricky to rebuild the resulting dict throught the config file.
Done a larger edit than targeted because of various Ruff warning about too many statements and complexity.


un-pogaz:
@Kristinita I suggest you to take a look at the already existing feature `--exclude-file`: each line into the given file will be ignore by codespell.
Since codespell ignore only the *exact* lines provided into the exclude file, and since the probability that your code will contain *exactly* the same line several times in different files, you can exclude some error with even greater precision than `--per-file-ignore`.

`--per-file-ignore` is an intermediate solution that target when `--exclude-file` will be too tedious to use (to many lines) and too risky to completely ignore the word globally with `--ignore-words-list`.

README.rst
```diff
@@ -156,6 +156,26 @@ Words should be separated by a comma.
        def wrod(wrods) # codespell:ignore
            pass
 
+Per-file ignores
+----------------
+
+To give a finer control, is possible to specified a additional set of words to ignore into a specific file only.
```
TheGiraffe3:
```suggestion
To give a finer control, it is possible to specify an additional set of words to ignore in a specific file.
```

README.rst
```diff
@@ -156,6 +156,26 @@ Words should be separated by a comma.
...
+
+1. ``--per-file-ignores``: A pair of arguments into the command line. The first provide a file, or a glob, and the second a comma-separated list of word to ignore for the given file:
```
TheGiraffe3:
```suggestion
1. ``--per-file-ignores``: A pair of arguments in the command line. The first provides a file, or a glob, and the second a comma-separated list of words to ignore for the given file:
```

Issue #3300, opened 2024-01-21
Feature: Ignore files not under version control


mdeweerd:
Context:
I created a [script to add lines to the file ignore list](https://gist.github.com/mdeweerd/edecd82d542b150859f65e6b73bdef79) that I can run if there are only false codespell positives.

I want to exclude files that are not under git version control as these do not need to be added as exceptions.

It seems appropriate and useful that 'codespell' has an option to do so.

For example, the following shell command filters to only files that are under version control:
`git ls-files $(ls -p | grep -v -E '/$')`.

~~I could apply something like that to the script by using the codespell output two times or running it a second time on the filtered files.~~  Done.

However, excluding files that are not under version controls seems like a good filter for codespell.

EDIT: updating the ignored lines file could even be a feature in codespell.

DimitriPapadopoulos:
Currently _codespell_ is a simple piece of software focused on processing the arguments it is given, without filters — except for hidden files. I am not certain adding filters is worth the extra complexity. You could instead modify the arguments passed to codespell. Any way, I'll let maintainers decide.

mdeweerd:
The use case is that when running it from the CLI with local files used for development, these files are codespell'ed as well and when running the script to ignore line that are misspelled, lines from files not under version control also got added.
I updated my script to avoid that, but a CLI codespell run still shows these "false positives".

yarikoptic:
this might also be a very nice alternative to needing to explicit list to ignore all of those other folders (`.git/` , `.mypy_cache/`) etc. And it would also help to avoid side-effects of fixing some files not under git, while committing only the ones under git.  So overall I would have liked to have such mode and making it configurable via config.

DimitriPapadopoulos:
OK, you convinced me. Which version control though? Not everyone uses git.

mdeweerd:
I vote for git - I moved almost all my subversion projects to git, rcs/cvs has been long gone, and it's what I encounter the most.

yarikoptic:
well, could be a setting `ignore_vcs = git` and then puke `NotImplementedError` for anything but `git` ATM so keeping it open for others to add andling for what they care. Minor note: for the test, make sure you have some files with obscure filenames [*] and use `git ls-files -z` and separate based on `\000` not newline , and also I think that would avoid you need to deal with quotes:

use without -z
```shell
❯ mkdir /tmp/test; cd /tmp/test; git init; echo 1 > "new
[...]

PR #3676, opened 2025-03-25
Fix #1897 #3300 Add git_filter

### LINKED ISSUE #1897: Inspect only git tracked files when run under git repos
Instead of linting all the folders under current directory or asking the user to maintain a long list of excludes, it would be far better to detect when the tool is run inside a git repository and perform linting only on tracked files: `git ls-files`

This would almost remove the need to maintain exclude lists and provide a very good user experience when running the tool for the first time on any repository.

I already used this approach and testing if inside a git repo is also another very simple command.

mdeweerd:


mdeweerd:
Open for review of the functionnality, feedback on how you want to handle the security notices (ignore, do it differently).

Regardubg tests; I accept contributions to my fork to update this PR.

codespell_lib/_codespell.py
```diff
@@ -1096,6 +1103,88 @@ def flatten_clean_comma_separated_arguments(
     ]
 
 
+def get_git_executable_path() -> str:
+    # Find the full path to the git executable
```
DimitriPapadopoulos, line 1107: Why find the full patch to the git executable instead of just running `git`?

codespell_lib/_codespell.py
```diff
@@ -1096,6 +1103,88 @@ def flatten_clean_comma_separated_arguments(
...
+
+    # Add pattern to exclude hidden files if check_hidden is False
+    if not check_hidden:
+        exclude_patterns.append(":(exclude)**/.*")
+
+    git_executable = get_git_executable_path()
+
+    try:
+        result = subprocess.run(
+            [git_executable, "ls-files", *file_args, *exclude_patterns],
```
DimitriPapadopoulos, line 1138: For consistency, I wouldn't let `git` handle exclude patterns. I think the exclude patterns should be handled by the same codespell code, whatever the option. Said otherwise, I would create a list of tracked files, and let codespell do all the filtering stuff, based on that list and other exclude patterns.

I may be missing an obvious obstacle though.

mdeweerd: Letting git do the work here:
- Avoids having to create a separate git list, filter the already existing lisdt against that list.
- Letting git do it is faster IMHO as it uses its internal database, not the file system.  Walking throught the file system is slower.  Avoiding the extra filtering step is also a performance gain.

mdeweerd:
@DimitriPapadopoulos - Thank you for the feedback.

codespell_lib/_codespell.py
DimitriPapadopoulos: The security notice can be disabled. It doesn't make sense.

PR #3688, opened 2025-04-11
Exit properly on exclude-file error

### LINKED ISSUE #3687: BUG: pre-commit hook gets invoked after file deletion and crashes if excluded file does not exist
After having deleted a file, the `pre-commit` hook fails with
```
codespell................................................................Failed
- hook id: codespell
- exit code: 1

  File "/home/user/.cache/pre-commit/repoga7ec3e6/py_env-python3.11/bin/codespell", line 8, in <module>
    sys.exit(_script_main())
             ^^^^^^^^^^^^^^
  File "/home/user/.cache/pre-commit/repoga7ec3e6/py_env-python3.11/lib/python3.11/site-packages/codespell_lib/_codespell.py", line 1102, in _script_main
    return main(*sys.argv[1:])
           ^^^^^^^^^^^^^^^^^^^
  File "/home/user/.cache/pre-commit/repoga7ec3e6/py_env-python3.11/lib/python3.11/site-packages/codespell_lib/_codespell.py", line 1261, in main
    build_exclude_hashes(exclude_file, exclude_lines)
  File "/home/user/.cache/pre-commit/repoga7ec3e6/py_env-python3.11/lib/python3.11/site-packages/codespell_lib/_codespell.py", line 726, in build_exclude_hashes
    with open(filename, encoding="utf-8") as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'block_reason.txt'
```
where
```
$ git status 
On branch dev
Your branch is up to date with 'origin/dev'.

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	deleted:    block_reason.txt
```
and
```
$ cat .pre-commit-config.yaml
...
- repo: https://github.com/codespell-project/codespell
  rev: v2.4.1
  hooks:
  - id: codespell
    args: ['--exclude-file', 'block_reason.txt']
...
```
when removing the `args` part it works again but I think it might be worth wile to either ignore the argument if the file does not exist and emit a warning and (or) not at all invoke `pre-commit` on deleted files.

DimitriPapadopoulos:
If the `exclude-file` argument cannot be read, catch the exception and exit properly with a meaningful error message.

Fixes #3687.


larsoner (commented):
Worth adding some test that these can actually be hit? Makes sure that things like `e.strerror` actually work in the strings

DimitriPapadopoulos:
@larsoner Do you think we should exit when a "secondary" file argument is missing, or just ignore the missing file as suggested in https://github.com/codespell-project/codespell/issues/3687#issuecomment-2796877747?

larsoner:
Ahh for an `--exclude` I think it's okay if the file is missing, it shouldn't lead to an error

PR #3704, opened 2025-05-08
Add scriptiong->scripting spelling correction.

cfi-gb:
See e.g.: https://grep.app/search?words=true&q=scriptiong


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3707, opened 2025-05-15
dict: multiple suggestions

mdeweerd:
Multiple sugestions.


codespell_lib/data/dictionary.txt
```diff
@@ -32517,6 +32537,8 @@ incremet->increment
 incremetal->incremental
 incremeted->incremented
 incremeting->incrementing
+incremint->increment
+incremment->increment
```
DimitriPapadopoulos, line 32541: I would recommend adding all verbal forms, if you have the time and patience.

mdeweerd: Always short on time, but it's not too long.
Added a bonus for another similar case.

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @mdeweerd.

PR #3722, opened 2025-06-19
Speed up spellchecking by ignoring whitespace-only lines

nthykier:
Whitespace only lines cannot have typos in them at all on account of not containing any words. Nevertheless, codespell did not exit early on them causing quite some overhead.

By special-casing empty and whitespace only, it is possible to reduce the runtime from ~5.4s to ~5.0s on the corpus mentioned in #3419.


larsoner:
Seems reasonable to me, thanks @nthykier !

PR #3723, opened 2025-06-23
Avoid deprecated license specification format

DimitriPapadopoulos:
Fix this pip warning:

	********************************************************************************
	Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

	By 2026-Feb-18, you need to update your project and remove deprecated calls
	or your builds will no longer be supported.

	See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
	********************************************************************************

[PEP 639](https://peps.python.org/pep-0639/) support requires setuptools [77](https://setuptools.pypa.io/en/stable/history.html#v77-0-0) which requires Python 3.9.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-          - "3.8"
           - "3.9"
           - "3.10"
           - "3.11"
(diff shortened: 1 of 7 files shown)


larsoner:
Thanks @DimitriPapadopoulos !

PR #3724, opened 2025-06-26
Add godd(s)->good(s), god(s)

luzpaz:


DimitriPapadopoulos (approved):
Could also be `goad` but that typo is probably too rare.

PR #3725, opened 2025-07-02
Add spelling correction for calid->valid and variants.

cfi-gb:
See e.g.:
- https://grep.app/search?words=true&q=calidate
- https://grep.app/search?words=true&q=calidation
- https://grep.app/search?words=true&q=calidated

Notes:
- Sees this happen because the `v` is usually next to the `c`
- I have left out https://www.collinsdictionary.com/dictionary/english/calidity for now as this seems to be some kind of valid (but obsolete) word

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -10817,6 +10817,15 @@ calibraitons->calibrations
 calibratin->calibrating, calibration,
 calibrte->calibrate
 calibrtion->calibration
+calid->valid, canid,
+calidate->validate
+calidated->validated
+calidates->validates
+calidating->validating
+calidation->validation
+calidations->validations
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Indeed, `calidity` seems far-fetched, but what about `canid`?

cfi-gb:
> but what about `canid`?

So you mean `calid->valid, canid,`?

If yes i can absolutely add this :+1:

DimitriPapadopoulos:
Yes, that's what I had in mind.

cfi-gb:
Wasn't fully sure, applied now with the latest version of this PR. Thanks a lot for the suggestion :+1:

DimitriPapadopoulos:
Thank you @cfi-gb.

PR #3743, opened 2025-07-21
Fix device typo

AJIOB:


AJIOB:
@DimitriPapadopoulos , I found the `dervice` word [here](https://en.wiktionary.org/wiki/dervice), so, I moved it to the rare dictionary

AJIOB:
Hi @DimitriPapadopoulos,

Is it OK to be merged?

PR #3744, opened 2025-07-23
dict: prefacted and related

mdeweerd:
Add prefacted->perfacted and variants.  Also added missing variants in similar corrections.
Prefect was already present as an alternative so added in more suggestions now


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @mdeweerd.

PR #3751, opened 2025-08-03
add faulyt->faulty and aulty->faulty

TheGiraffe3:


codespell_lib/data/dictionary.txt
```diff
@@ -8197,6 +8197,7 @@ auhtorized->authorized
 auhtorizes->authorizes
 auhtorizing->authorizing
 auhtors->authors
+aulty->faulty
```
DimitriPapadopoulos, line 8200: It could perhaps be a typo of `aunty`, but then I don't know if there is actually such a typo out there.

TheGiraffe3: I just realized I never responded to this...
```suggestion
aulty->aunty, faulty,
```
Even if that typo doesn't exist yet, it's worth having just in case.

DimitriPapadopoulos:
Thank you @TheGiraffe3.

PR #3759, opened 2025-08-13
dictionary: add regoin->region

thiagowfx:
Observed in our codebase.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -49440,7 +49440,7 @@ regnerated->regenerated
 regnerates->regenerates
 regnerating->regenerating
 regneration->regeneration
-regoin->region
+regoin->regain, region,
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -49440,6 +49440,7 @@ regnerated->regenerated
 regnerates->regenerates
 regnerating->regenerating
 regneration->regeneration
+regoin->region
```
TheGiraffe3, line 49443:
```suggestion
regoin->regain, region,
```

thiagowfx: Thanks. For some reason, I can't apply these changes directly via github:

Image

I'll make a manual change

DimitriPapadopoulos:
Thank you @thiagowfx and @TheGiraffe3.

PR #3762, opened 2025-08-19
Remove ACI from dictionary

Zer0x00:
ACI is a product of [Cisco](https://www.cisco.com/site/de/de/products/networking/cloud-networking/application-centric-infrastructure/index.html) and stands for Application Centric Infrastructure


DimitriPapadopoulos:
We can't remove all possible acronyms. There would be no typos left in the dictionary! Add it to your skip list.

PR #3765, opened 2025-08-25
Use a maintained mirrors-prettier for pre-commit

hugovk:
https://github.com/pre-commit/mirrors-prettier was archived in April 2024 so let's use a maintained mirror such as https://github.com/rbubley/mirrors-prettier.

(This also helps https://github.com/j178/prek/issues/545 which is caused by https://github.com/pre-commit/mirrors-prettier not setting a minimum Node version for Prettier)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,8 +62,8 @@ repos:
     hooks:
       - id: ruff-check
       - id: ruff-format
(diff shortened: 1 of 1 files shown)


DimitriPapadopoulos:
Let's apply this for now. However the Prettier documentation recommends using a different [Pre-commit Hook](https://prettier.io/docs/precommit), typically [`pretty-quick`](https://github.com/prettier/pretty-quick).

Issue #3770, opened 2025-09-16
Support for Oxford English


johann1764:
Oxford English is considered by some by the most "neutral" English.
It is basically British English **except** that words endings in `-ize/izes/ized/ization` are like US English.
Special case: `colourized` (GB: `colourised` US: `colorized`)

It is therefore natural to want to use this convention.

Codespell does not support it currently.
But the file codespell_lib/data/dictionary_en-GB_to_en-US.txt can be used to make two new ones:
 - `dictionary_en-GB_to_en-OX.txt` by keeping only -iz{e,es,ed,zing} words.
 - `dictionary_en-US_to_en-OX.txt` by skipping -iz{e,es,ed,zing} words and reversing the others from `a->b` to `b->a`.
 - `dictionary_en_to_en-OX.txt` can be done by merging the two previous ones.

(This requires a bit of special case management)

My question is twofold:
 1. Would you be interested if I PR this `dictionary_en_to_en-OX.txt` file ?
 2. If not, how do we do with the licensing ? I intend to include it in Naev https://codeberg.org/naev/naev which is also under GPL.

DimitriPapadopoulos:
Sounds like a good idea. It would be great if dictionary generation could be automated using a script.

johann1764:
Ok, I do that.

PR #3771, opened 2025-09-16
Support for Oxford English

johann1764:
**This PR has been replaced by #3772.**
This PR aims at addressing issue #3770.

 - generate `dictionary_en_to_en-OX.txt` from `dictionary_en-GB_to_en-US.txt` with new script `gen_OX.sh` (I let you give it a better name, if you have ideas).
 - while at it, added one line in `dictionary_en-GB_to_en-US.txt`: `saleable->salable`
 - Manage this new file in `_codespell.py`
 - added a bit of vocabulary (the actual examples found in naev doc)
 - added the 'defense' exception that is spelled in Ox English the same as US, not GB.
 - managed the storey[s] case that should not be backward-translated into story (because GB story also exists, with a different meaning)
 - fixed a mistake in `dictionary_en-GB_to_en-US.txt`
 - more helpful error message when `pytest-cov` not installed


larsoner:
`make check-dictionaries` is failing, so is a test... I think maybe it needs to be added to the help docs or something?

johann1764:
I finally found a way to install all I need to `make check`.
However, even with the **main branch**:
```
===================================== test session starts =====================================
platform linux -- Python 3.13.3, pytest-8.4.2, pluggy-1.6.0
rootdir: /home/nicolas/codespell
configfile: pyproject.toml
plugins: cov-7.0.0
collected 0 items / 1 error                                                                   

=========================================== ERRORS ============================================
___________________ ERROR collecting codespell_lib/tests/test_dictionary.py ___________________
'dependency' not found in `markers` configuration option
---------------- generated xml file: /home/nicolas/codespell/junit-results.xml ----------------
!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!
====================================== 1 error in 2.12s =======================================
make: *** [Makefile:15: check-dictionaries] Error 2
```

Without an error message related to my branch, there is not much I can do.

Could you try running the workflows, so that I can get error output ?

larsoner:
Running now!

johann1764:
Thanks !

This allowed to point out that the tests were run before the new dictionary was generated. I fixed the Makefile (stupid mistake of mine), that's one less error.

Now, let's try again !

larsoner:
Locally you can run `pytest codespell_lib` at least to make sure it passes. Should reduce the number of commits needed to get things working

johann1764:
> Locally you can run `pytest codespell_lib` at least to make sure it passes. Should reduce the number of commits needed to get things working

It always results in:
```
===================================== test session starts =====================================
platform linux -- Python 3.13.3, pytest-8.4.2, pluggy-1.6.0
rootdir: /home/nicolas/codespell
configfile: pyproject.toml
plugins: cov-7.0.0
collected 60 items / 1 error                                                                  

=========================================== ERRORS ============================================
___________________ ERROR collecting codespell_lib/tests/test_dictionary.py ___________________
'dependency' not found in `markers` configuration option
---------------- generated xml file: /home/nicolas/codespell/junit-results.xml ----------------
!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!
====================================== 1 error in 9.12s =======================================
```

which is not helpful.

It might not be obvious, but we're actually making progress :-)

EDIT: If you want, I can rebase all this when it works, and submit the rebased version as another PR.

larsoner:
You need pytest-dependency

https://github.com/codespell-project/codespell/blob/87a280039336c73d45faba7829aea860e68de4a3/pyproject.toml#L39

johann1764:
Ok, now the remaining error.. is not an error.

The dictionary loop test realized that `dictionary_en-GB_to_en-US.txt` often offers opposite substitutions with `dictionary_en_to_en-OX.txt`... which was kind of the point.

en_to_en-OX contains acknowledgment->acknowledgement, which is correct.
en-GB_to_en-US contains acknowledgement->acknowledgment, which is also correct.
This makes the tests fail.

The fundamental problem here is that `dictionary_en-GB_to_en-US.txt` and `dictionary_en_to_en-OX.txt` are incompatible, and it appears that the tests aim at avoiding that.

I don't know what to do.

larsoner:
You can adjust the test to accommodate the exception if needed

larsoner:
... i.e., make it smarter about which dictionaries it checks, add a `pytest.param(..., marks=pytest.xfail(...))` or whatever is needed

johann1764:
I found a clean way to manage this !

In `test_dictionaries.py`, there is a list `allowed_dups` that seems to serve the same purpose as what I was looking for.

Let us see if it works.

EDIT: stupid me, forgot one line.

johann1764:
I will definitely rebase all that when it works.

larsoner:
We can squash+merge, no need for rebase or messing with commits

johann1764:
According to the workflows on my fork, it works on all but windows, I don't know why (yet).

For the squashing scheme, I suggest squashing ".", "oops" and "(cont)" commits.

larsoner:
Okay I think things are a bit more complicated than needed because we don't actually *ship* the new dictionary in this PR. You generate it on the fly, right? And on Windows it fails for some reason. And when we build and ship our wheels, we'll need to make sure to do it then, too. This seems a little too magical and will increase maintenance burden a bit.

How about we "just" ship the dictionary instead? And if it needs to be updated based on some future PR, we can have an [autofix.ci GitHub action](https://github.com/mne-tools/mne-python/blob/main/.github/workflows/autofix.yml) run the `make <whatever>` that might cause it to be updated, which will then push a commit to the given PR. I've used it in other repos to keep stuff up to date and it works really well and with very limited additional maintenance burden. Seems easier than the approach here...

johann1764:
> Okay I think things are a bit more complicated than needed because we don't actually _ship_ the new dictionary in this PR. You generate it on the fly, right? And on Windows it fails for some reason. And when we build and ship our wheels, we'll need to make sure to do it then, too. This seems a little too magical and will increase maintenance burden a bit.

I am having a look at the Makefile. It is possible that the order of subgoals is not respected in some cases, maybe a more robust Makefile will solve it.

> 
> How about we "just" ship the dictionary instead? And if it needs to be updated based on some future PR, we can have an [autofix.ci GitHub action](https://github.com/mne-tools/mne-python/blob/main/.github/workflows/autofix.yml) run the `make <whatever>` that might cause it to be updated, which will then push a commit to the given PR. I've used it in other repos to keep stuff up to date and it works really well and with very limited additional maintenance burden. Seems easier than the approach here...

I am not strongly against that, but I agreed to @DimitriPapadopoulos 's suggestion to do it that way. Furthermore, having generated files shipped with the project has the inconvenient that someone not aware this file is generated could work on it and lose all his/her work just by calling make, which is a serious caveat.

Let me try a few things first. When it works, we'll see more clearly the impact on the project (meaning we will see how the Makefile looks like)

larsoner:
> has the inconvenient that someone not aware this file is generated could work on it and lose all his/her work just by calling make, which is a serious caveat.

Yeah, this is why such files I typically put an all-caps "# THIS FILE IS AUTOGENERATED BY ..." and so far have had zero cases of people reporting trying to modify them. So this is mitigated easily enough I think. In the meantime I'm still worried about things like needing to add infra to make sure it *is* generated when we make a wheel.

Just for another example, when people `git clone` then `pip install -e .` (standard nowadays), the necessary file won't be there and things won't work correctly when they try to run `codespell`. I think there are more drawbacks here than it seems at first glance...

johann1764:
> Yeah, this is why such files I typically put an all-caps "# THIS FILE IS AUTOGENERATED BY ..." and so far have had zero cases of people reporting trying to modify them. So this is mitigated easily enough I think. In the meantime I'm still worried about things like needing to add infra to make sure it _is_ generated when we make a wheel.

I didn't know you could add comments in this dictionary files. But if it is the case, then yes, this mitigates the problem.

> 
> Just for another example, when people `git clone` then `pip install -e .` (standard nowadays), the necessary file won't be there and things won't work correctly when they try to run `codespell`. I think there are more drawbacks here than it seems at first glance...

Are you saying make isn't even called in such a situation ??

Makefiles are made to manage these kind of cases, but if you don't use it, then it is logical to want the work already be done.

Anyway, the only difference here is whether or not we add the file.
This can be done anytime with 3 lines and a single commit:
```shell
make
git add codespell_lib/data/dictionary_en_to_en-OX.txt
git commit -m 'added `dictionary_en_to_en-OX.txt`'
```

Even if we do that, `Makefile` dependencies must be correct in order to work with files that are up-to-date, so that does not change what I need to do.

larsoner:
> I didn't know you could add comments in this dictionary files. But if it is the case, then yes, this mitigates the problem.

I'm actually not sure if we can... but we can name the file `dictionary_en_to_en-OX_AUTOGENERATED.txt` or something if we want

> This can be done anytime with 3 lines and a single commit:

... that and plus an `autofix.ci` that runs it should keep everything up to date. I can push a commit for that if it would help

johann1764:
Ok, I really don't understand your build system.

So I included the dictionary. That way, it will work in any situation.
Now it works.

> > I didn't know you could add comments in this dictionary files. But if it is the case, then yes, this mitigates the problem.
> 
> I'm actually not sure if we can... but we can name the file `dictionary_en_to_en-OX_AUTOGENERATED.txt` or something if we want

From what I saw in the parser, I guess we can't. But I may have missed it.

For the file name, can you do it, or do you want me to do it ?
It is just about substituting the old name by the new one in Makefile, _codespell.py and test_dictionary.py. And renaming the file, of course.

EDIT: I finally did it.

> ... that and plus an `autofix.ci` that runs it should keep everything up to date. I can push a commit for that if it would help

I am not sure what it means, but please do it if you think it is a good idea.

johann1764:
That should be good.

johann1764:
**This PR has been replaced by #3772**

PR #3772, opened 2025-09-18
Support for Oxford English (rebased)

johann1764:
**This is a rebased version of #3771.**

This PR aims at addressing issue #3770.

 - generate `dictionary_en_to_en-OX.txt` from `dictionary_en-GB_to_en-US.txt` with new script `gen_OX.sh` (I let you give it a better name, if you have ideas).
 - while at it, added one line in `dictionary_en-GB_to_en-US.txt`: `saleable->salable`
 - Manage this new file in `_codespell.py`
 - added a bit of vocabulary (the actual examples found in naev doc)
 - added the 'defense' exception that is spelled in Ox English the same as US, not GB.
 - managed the storey[s] case that should not be backward-translated into story (because GB story also exists, with a different meaning)
 - fixed a mistake in `dictionary_en-GB_to_en-US.txt`
 - more helpful error message when `pytest-cov` not installed

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1728,7 +1728,7 @@ activsted->activated
 activstes->activates
 activte->activate
 activted->activated
-activtes->activates
+activtes->activates, activities,
 activties->activities
 activting->activating
 activtion->activation
@@ -7059,6 +7059,7 @@ askign->asking
 askin->asking, ask in, akin, skin, a skin,
 askreddt->AskReddit
 aslo->also
+asncronous->asynchronous
 asnd->and
 asnwer->answer
 asnwered->answered
@@ -13314,6 +13315,7 @@ cloess->closes
 cloesst->closest
 cloisonay->cloisonné
 cloisonays->cloisonnés
+clokwise->clockwise
 clonez->clones, cloner,
 clonin->cloning
 clonne->clone
@@ -13775,6 +13777,7 @@ coloered->colored, coloured,
 coloering->coloring, colouring,
 coloers->colors, colours,
 coloful->colorful, colourful,
+colokwise->clockwise
 colombus->Columbus
 colomn->column
 colomns->columns
@@ -13787,6 +13790,7 @@ colorizoer->colorizer
 colorpsace->colorspace
 colorpsaces->colorspaces
 colose->close
+coloskwise->clockwise
 colouer->colour
 coloum->column
 coloumn->column
@@ -13804,6 +13808,8 @@ colsest->closest, coldest,
 colsing->closing
 colud->could, cloud,
 colum->column
+colume->column, volume,
+columes->columns, volumes,
 columm->column
 colummn->column
 colummns->columns
@@ -23395,7 +23401,7 @@ downsteramer->downstreamer
 downsteramers->downstreamers
 downsteraming->downstreaming
 downsterams->downstreams
-dows->does
+dows->does, down,
 dowt->doubt
 doxgen->doxygen
 doygen->doxygen
@@ -23459,7 +23465,7 @@ drammatic->dramatic
 dramtic->dramatic
 dran->drawn
 drapin->draping
-drastical->drastically
+drastical->drastically, drastic,
 drasticaly->drastically
 drats->drafts
 draughtman->draughtsman
@@ -28206,6 +28212,7 @@ fiields->fields
 filal->final
 filaname->filename
 filanames->filenames
+filaneme->filename
 filder->folder, filter, finder, fielder, filer,
 filders->folders, filters, finders, fielders, filers,
 fileand->file and
@@ -32680,6 +32687,10 @@ incresed->increased
 increses->increases
 incresing->increasing
 incrfemental->incremental
+incriease->increase
+incrieased->increased
+incrieases->increases
+incrieasing->increasing
 incrmenet->increment
 incrmenetd->incremented
 incrmeneted->incremented
@@ -40969,6 +40980,8 @@ notifictions->notifications
 notififation->notification
 notififations->notifications
 notifing->notifying
+notifition->notification
+notifitions->notifications
 notifiy->notify
 notifiying->notifying
 notifyable->notifiable
@@ -41575,8 +41588,8 @@ offred->offered
 offsence->offence
 offsense->offense
 offsenses->offenses
-offser->offset
-offsers->offers
+offser->offer, offset,
+offsers->offers, offsets,
 offsest->offsets, offset,
 offseted->offsetted
 offsetes->offsets
@@ -42561,6 +42574,7 @@ otherwhile->otherwise
 otherwhise->otherwise
 otherwice->otherwise
 otherwide->otherwise
+otherwiese->otherwise
 otherwis->otherwise
 otherwize->otherwise
 otherwordly->otherworldly
@@ -44773,6 +44787,7 @@ playgrouns->playgrounds, playground,
 playin->playing, play in,
 playist->playlist
 playists->playlists
+playlise->playlist
 playright->playwright
 playwrite->playwright
 playwrites->playwrights
@@ -46903,6 +46918,10 @@ propagatin->propagating, propagation,
 propage->propagate, propane,
 propaged->propagated
 propages->propagates
+propagete->propagate
+propageted->propagated
+propagetes->propagates
+propageting->propagating
 propaging->propagating
 propatagion->propagation
 propator->propagator
@@ -47398,6 +47417,8 @@ puplishes->publishes
 puplishing->publishing
 pupose->purpose
 puposes->purposes
+puprose->purpose
+puproses->purposes
 pupular->popular
 pupularity->popularity
 pupulate->populate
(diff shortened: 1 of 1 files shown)


bobbens:
I've been testing it a bit and some comments:
1. I don't believe the documentation is updated (the text on builtin dictionaries in `_codespell.py`)
2. `focused` should be correct in Oxford spelling and I believe is included in the generated dictionary.
3. `defence` is correct in Oxford English and I believe the rule `defence->defense` is being generated
4. `civilisation->civilization` is missing

johann1764:
> I've been testing it a bit and some comments:
> 
> 1. I don't believe the documentation is updated (the text on builtin dictionaries in `_codespell.py`)

I'll have a look after fixing next points.

> 2. `focused` should be correct in Oxford spelling and I believe is included in the generated dictionary.

That appears to be true. I'll fix this.

> 3. `defence` is correct in Oxford English and I believe the rule `defence->defense` is being generated

It is included in the 'US spelling list', but could be rather be included in the 'ignored list'.

Both spellings are correct in OX. But do we want both to co-exist in the same corpus ?
IMO one should be picked, and on the corpus I checked, defense were more frequent.

What do you think ?

> 4. `civilisation->civilization` is missing

I will add it to en-GB_to_en-US and update the generated dict accordingly.

bobbens:
> Both spellings are correct in OX. But do we want both to co-exist in the same corpus ? IMO one should be picked, and on the corpus I checked, defense were more frequent.

OK, I was looking at https://en.wikipedia.org/wiki/Oxford_spelling and it seems to indicate "defence" is correct. However, it is true that the oxendict dictionary allows both, so I'm not sure.

johann1764:
> > Both spellings are correct in OX. But do we want both to co-exist in the same corpus ? IMO one should be picked, and on the corpus I checked, defense were more frequent.
> 
> OK, I was looking at https://en.wikipedia.org/wiki/Oxford_spelling and it seems to indicate "defence" is correct. However, it is true that the oxendict dictionary allows both, so I'm not sure.

I leave it as is for now. This can easily be changed in the future by moving `'defenc'` from the `EXCEPTIONS` array to the `IGNORE` array below:
```shell
# choose US for these ones
EXCEPTIONS=(
   'defenc'
   'focussed'
)
PAT2="^$(one_of "${EXCEPTIONS[@]}")"

# these one should be left out
IGNORE=(
   'storey'
   'practise'
)
PAT3="^$(one_of "${IGNORE[@]}")"
```

Now only point 1. remains to be dealt with.

johann1764:
> 1. I don't believe the documentation is updated (the text on builtin dictionaries in `_codespell.py`)

`_builtin_dictionaries` is up to date (lines 126-134), therefore `builtin_opts` should also be, therefore the help on `--builtin` should also be up to date.

Am I missing something ?

bobbens:
> > 1. I don't believe the documentation is updated (the text on builtin dictionaries in `_codespell.py`)
> 
> `_builtin_dictionaries` is up to date (lines 126-134), therefore `builtin_opts` should also be, therefore the help on `--builtin` should also be up to date.
> 
> Am I missing something ?

I meant the descriptions of the dictionaries that appears with `codespell --help` at https://github.com/codespell-project/codespell/blob/87a280039336c73d45faba7829aea860e68de4a3/codespell_lib/_codespell.py#L118 .

johann1764:
These are the lines you mention:
```python
    (
        "en-GB_to_en-US",
        "for corrections from en-GB to en-US",
        "_en-GB_to_en-US",
        True,
        True,
        ("en_GB",),
        ("en_US",),
    ),
    (
        "en_to_en-OX",
        "for corrections from en to en-OX",
        "_en_to_en-OX_AUTOGENERATED",
        None,
        None,
        None,
        None,
    ),
)
_builtin_default = "clear,rare"
```

Which looks up-to-date.

Maybe the codespell you run is not this up-to-date version ?

bobbens:
> These are the lines you mention:
> 
> ```python
>     (
>         "en-GB_to_en-US",
>         "for corrections from en-GB to en-US",
>         "_en-GB_to_en-US",
>         True,
>         True,
>         ("en_GB",),
>         ("en_US",),
>     ),
>     (
>         "en_to_en-OX",
>         "for corrections from en to en-OX",
>         "_en_to_en-OX_AUTOGENERATED",
>         None,
>         None,
>         None,
>         None,
>     ),
> )
> _builtin_default = "clear,rare"
> ```
> 
> Which looks up-to-date.
> 
> Maybe the codespell you run is not this up-to-date version ?

Ah, I see. I must have missed it. Sorry for the noise.

larsoner:
Okay for me to push the `autofix.ci` GHA run I was thinking of?

johann1764:
Ok!

Makefile
```diff
@@ -20,14 +25,14 @@ check-dictionaries:
 	@if command -v pytest > /dev/null; then \
 		pytest codespell_lib/tests/test_dictionary.py; \
 	else \
-		echo "Test dependencies not present, install using 'pip install -e \".[dev]\"'"; \
+		echo "Test dependencies not present, install using 'pip install -e \".[dev]\" pytest-cov --include-deps --force'"; \
```
larsoner: Not sure why you need `pytest-cov` here, it's in `dev`. I'm not sure why you need `--include-deps` or `--force` here either? The `--include-deps` should be the default, and `--force` seems unsafe and non-standard. I think we should revert this...

Makefile
```diff
@@ -1,16 +1,21 @@
-DICTIONARIES := codespell_lib/data/dictionary*.txt codespell_lib/tests/data/*.wordlist
+DICTIONARIES = codespell_lib/data/dictionary*.txt codespell_lib/tests/data/*.wordlist
 
 PHONY := all check check-dictionaries sort-dictionaries trim-dictionaries check-dist pytest pypi ruff clean
 
-all: check-dictionaries codespell.1
+GENERATED := codespell_lib/data/dictionary_en_to_en-OX_AUTOGENERATED.txt
 
-check: check-dictionaries check-dist pytest ruff
+all: $(GENERATED) check-dictionaries codespell.1
```
larsoner, line 7: This generates a `codespell.1` that doesn't seem to be used in packaging. Do we need it?

If we want to keep it, it should at least live in `.gitignore`

Makefile
```diff
@@ -1,16 +1,21 @@
...
+
+check: $(GENERATED) check-dictionaries check-dist pytest ruff
+
+$(GENERATED): codespell_lib/data/dictionary_en-GB_to_en-US.txt ./codespell_lib/gen_OX.sh
```
larsoner: I am not good at makefiles... is there a `make <something>` I could use to run this? If so please fix the `autofix.ci` which will just duplicate the command below...

tools/gen_OX.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
larsoner, line 1: One other idea... thinking this shouldn't live in `codespell_lib` because it's not meant for end users. Maybe a new `tools/gen_OX.sh` makes more sense

Makefile
johann1764: ok.
I had to do this to make it "work", but maybe this is not supposed to be the case.
Let's revert it.

Makefile
johann1764: I completed the `Makefile` line:
```make
all: check-dictionaries codespell.1
```
with:
```make
all: $(GENERATED) check-dictionaries codespell.1
```

I did not change anything about `codespell.1`.

Your observation may be relevant (I can't possibly know that),
but this PR is not where it should be fixed if need be, as it is unrelated.

Makefile
johann1764:
```shell
$ make codespell_lib/data/dictionary_en_to_en-OX_AUTOGENERATED.txt
```
updates the file if needed. If you call it now, you should currently get:
```shell
make: « codespell_lib/data/dictionary_en_to_en-OX_AUTOGENERATED.txt » is up to date.
```

Regarding the autofix, I guess you want me to update the `.github/workflows/autofix.yml`
by replacing:
```yaml
      - run: ./codespell_lib/gen_OX.sh codespell_lib/data/dictionary_en-GB_to_en-US.txt > codespell_lib/data/dictionary_en_to_en-OX_AUTOGENERATED.txt
```
with
```yaml
      - run: make codespell_lib/data/dictionary_en_to_en-OX_AUTOGENERATED.txt
```

Is that correct ?

larsoner: Yeah that should do it!

If you want to test that it works, delete some line from the `AUTOGENERATED.txt` and `autofix.ci` should put it back... it's what I did earlier above but you can try again if you want

tools/gen_OX.sh
johann1764: > this shouldn't live in codespell_lib because it's not meant for end users

True, but neither does `tests/`.
Maybe creating a `tools/` dir for a single script is overkill.

Here I see different options:
 1. /
 2. /tools/
 3. /codespell_lib/
 4. /codespell_lib/tools/

Let me know your favourite.

larsoner:
> True, but neither does tests/

FWIW, as an end user of libraries I have done the equivalent of `pytest -m codespell_lib` for an installed module, so there are some use cases at least! I think it's something we can support users doing and running as it's pretty standard in Python libraries... the script stuff added here, not so much. Let's go with `/tools/`

johann1764:
That should be it.

johann1764:
You're welcome !
Using the new feature might lead me to propose new entries in the dic in the future.

PR #3789, opened 2025-10-21
Add spelling correction for inherent and variant.

cfi-gb:
Not much hits but still happening from time to time:
- https://github.com/search?q=inheent&type=code
- https://github.com/search?q=inheently&type=code


DimitriPapadopoulos:
Thank you @cfi-gb.

Issue #3696, opened 2025-04-27
aspell-python-py3 → pyspelling / textutils3


DimitriPapadopoulos:
Because of pending PRs and lack of release since 2016, should we switch from [aspell-python-py3](https://github.com/WojciechMula/aspell-python/) to [PySpelling](https://github.com/facelessuser/pyspelling) or [textutils3](https://pypi.org/project/textutils3/)?

PR #3791, opened 2025-10-26
DRAFT: Add free threaded Python 3.14t to test matrix

cclauss:
* #3778
```
___________ ERROR collecting codespell_lib/tests/test_dictionary.py ____________
codespell_lib/tests/test_dictionary.py:22: in <module>
    import aspell  # type: ignore[import]
    ^^^^^^^^^^^^^
E   RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'aspell', which
    has not declared that it can run safely without the GIL. To override this behavior and keep the
    GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```
https://pypi.org/project/aspell-python-py3

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,6 +62,8 @@ jobs:
       - run: codespell --help
       - run: codespell --version
       - run: make check
+        env:
+          PYTHON_GIL: 0
         if: startsWith(matrix.os, 'ubuntu')
       - run: pytest codespell_lib
         if: startsWith(matrix.os, 'windows')


DimitriPapadopoulos:
Perhaps switching to [PySpelling](https://github.com/facelessuser/pyspelling) or [textutils3](https://pypi.org/project/textutils3/) could help, see #3696.

DimitriPapadopoulos:
I don't expect codespell to have compatibility issues.

I haven't found yet how to easily define `PYTHON_GIL=0` for free threaded Python version 3.13t and 3.14t only. It might be easier to conditionally pass `-Xgil=0` to Python.

cclauss:
I think we are OK if we use 3.14 and not 3.14t.

PR #3805, opened 2025-11-04
Add several corrections for mouting in rare dictionary

FloEdelmann:


codespell_lib/data/dictionary.txt
```diff
@@ -39115,6 +39115,7 @@ mountin->mounting, mountain, mountie, mount in,
 mountpiont->mountpoint
 mountpionts->mountpoints
 mouspointer->mousepointer
+mouting->mounting
```
DimitriPapadopoulos: It looks like [mouting](https://www.oed.com/search/dictionary/?q=mouting) is a variant of moulting | molting. Should we move it to the _rare_ dictionary?

DimitriPapadopoulos: in any case:
```suggestion
mouting->mounting, moulting, mouthing, pouting,muting,
```

FloEdelmann: I moved it to the _rare_ dictionary now and added the other corrections. Thanks for the suggestions!

DimitriPapadopoulos (approved):

Issue #3651, opened 2025-02-24
Disable automatic fixes for words in the ***rare*** dictionary


DimitriPapadopoulos:
See https://github.com/codespell-project/codespell/pull/3631#issuecomment-2677798780.

From the README:

> 3. Entry with one word, but with automatic fix disabled:
>    
>    ```
>    clas->class, disabled because of name clash in c++
>    ```
> Note that there isn't a comma at the end of the line. The last argument is treated as the reason why a suggestion cannot be automatically applied.

Perhaps we should disable automatic fixes by adding a comment to all or part of the words in the _rare_ dictionary. Example:
```console
$ cat DICTIONARY.txt 
lien->line, actually a word in the OED and SCOWL (And Friends)
$ 
$ cat test.txt 
lien
$ 
$ codespell -w -D DICTIONARY.txt test.txt 
test.txt:1: lien ==> line  | actually a word in the oed and scowl (and friends)
$ 
$ codespell -D DICTIONARY.txt test.txt 
test.txt:1: lien ==> line  | actually a word in the oed and scowl (and friends)
$ 
```

johann1764:
> Perhaps we should disable automatic fixes by adding a comment to all or part of the words in the rare dictionary

Considering that, by definition, rare means "correct forms", it seems logical to me to comment them all. The good news is it is quickly done:
```shell
$ sed -i 's/^\([^,]*\)$/\1,/' dictionary_rare.txt
```

that results in:
<details><summary>unfold</summary>

```diff
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -1,315 +1,315 @@
 ablet->able, tablet,
-acapella->a cappella
-accreting->accrediting
-acter->actor
-acters->actors
-aer->are
-afterwords->afterwards
+acapella->a cappella,
+accreting->accrediting,
+acter->actor,
+acters->actors,
+aer->are,
+afterwords->afterwards,
 amination->animation, lamination,
 aminations->animations, laminations,
 aminator->animator, laminator,
 aminators->animators, laminators,
-anc->and
-ans->and
-arithmetics->arithmetic
+anc->and,
+ans->and,
+arithmetics->arithmetic,
 attache->attaché, attached, attach,
-automatize->automate
-automatized->automated
-automatizes->automates
-backword->backward
-backwords->backwards
+automatize->automate,
+automatized->automated,
+automatizes->automates,
+backword->backward,
[...]

PR #3826, opened 2025-11-18
Remove "sting" correction from dictionary_rare.txt

TheGiraffe3:
"sting" is a word.

### DIFF
--- a/codespell_lib/data/dictionary_rare.txt
+++ b/codespell_lib/data/dictionary_rare.txt
@@ -269,8 +269,6 @@ spacial->special, spatial,
 specif->specify, specific,
 statics->statistics
 steams->streams
-sting->string
-stings->strings
 straightaway->straight away
 straighted->straightened, straighten,
 stye->style


DimitriPapadopoulos:
Indeed, `sting` is not even a rare word:
https://books.google.com/ngrams/graph?content=sting%2Cstring&year_start=1800&year_end=2022&corpus=en

I guess that this is a frequent typo in the context of source code, which was the initial target of this tool.

TheGiraffe3:
Should I move it to `dictionary_code`?

peternewman:
> "sting" is a word.

Everything in the rare dictionary is a word, that's its literal definition!
https://github.com/codespell-project/codespell/blob/e61ac16e1f01a5478067e400080282fadc457889/codespell_lib/_codespell.py#L74

You can either not use the rare dictionary, or exclude this particular word depending on your preference.

> Indeed, `sting` is not even a rare word: https://books.google.com/ngrams/graph?content=sting%2Cstring&year_start=1800&year_end=2022&corpus=en
> 
> I guess that this is a frequent typo in the context of source code, which was the initial target of this tool.

I think the ngrams would look very different against a corpus of source code (if anyone has data for that).

As I raised ages ago ( #1469 ), should we split the rare dictionary. Tangentially also as @DimitriPapadopoulos raised, should it not auto-correct #3651 . I assume that happened and caused this PR @TheGiraffe3 ?

> Should I move it to `dictionary_code`?

No, that's specifically for words that are valid in code, but not elsewhere (e.g. thead is a table header in HTML, but would be a typo elsewhere).

TheGiraffe3:
No, I opened this PR because a GitHub Action is failing on one of my repositories every time I change a specific line (and forget to readd it to .codespell.exclude), and "sting" seemed to be a perfectly valid word to me.

I can close the PR (as with #1902) if you want, and I'll figure out how to disable the rare dictionary.

peternewman:
> No, I opened this PR because a GitHub Action is failing on one of my repositories every time I change a specific line (and forget to readd it to .codespell.exclude), and "sting" seemed to be a perfectly valid word to me.
> I'll figure out how to disable the rare dictionary.

You can also ignore just the specific word `sting` if you'd prefer, which means you'll still get notified if you type `chancel` when you meant `cancel` for example, it depends on your appetite for potential false positives around the rare dictionary. Whereas (in theory) no valid words, but potentially some less frequently used domains-specific acronyms or abbreviations, should be in the main dictionary.

PR #3828, opened 2025-11-19
Add woork->work and formace->format and friends

peternewman:


codespell_lib/data/dictionary.txt
```diff
@@ -28947,6 +28947,8 @@ foriners->foreigners
 forld->fold
 forlder->folder
 forlders->folders
+formace->format
+formaces->formats
```
DimitriPapadopoulos: Perhaps:
```suggestion
formace->format, formate,
formaces->formats, formates,
```

peternewman: For some reason it wasn't possible to apply the suggestion in any of the three ways, so I've manually added it in @DimitriPapadopoulos .

peternewman: Seemed kind of niche, but probably better than a false positive of auto-correcting to the wrong thing.

peternewman: It turns out formate and formates were incorrectly in the dictionary as typos, so I've now moved them to rare instead!

DimitriPapadopoulos (approved):

PR #3832, opened 2025-11-20
Fix three problems with --ignore-multiline-regex

### LINKED ISSUE #3642: With ignore-multiline-regex set, `--write` doesn't work properly
When `ignore-multiline-regex` is set in config, and I run `codespell -w`, it removes newlines in edited files, which results in files being reformatted to a single line.

```
$ echo -e "Thsi line contains a typo\nWhile this line is correct" > test.txt
$ cat test.txt 
Thsi line contains a typo
While this line is correct
$ codespell test.txt 
> Thsi line contains a typo
test.txt:1: Thsi ==> This
$ codespell test.txt -w --ignore-multiline-regex="codespell:ignore-begin.*codespell:ignore-end"
FIXED: test.txt
$ cat test.txt 
$ cat test.txt 
This line contains a typoWhile this line is correct
```

vries:
This series consists of four patches.

Patches 1, 2 and 4 each fix a problem with `--ignore-multiline-regex` (described in the commit messages), and add a corresponding unit test.

Patch 3 is a refactoring patch, making patch 4 smaller.

Patch 2 fixes [this issue](https://github.com/codespell-project/codespell/issues/3642) .

Closes #3642

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -227,12 +227,14 @@ class FileOpener:
 
         self.encdetector = UniversalDetector()
 
-    def open(self, filename: str) -> tuple[list[str], str]:
+    def open(self, filename: str) -> tuple[list[tuple[bool, int, list[str]]], str]:
         if self.use_chardet:
             return self.open_with_chardet(filename)
         return self.open_with_internal(filename)
 
-    def open_with_chardet(self, filename: str) -> tuple[list[str], str]:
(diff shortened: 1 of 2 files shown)


DimitriPapadopoulos:
Looks good to me, I think the coverage issue predates these patches. @larsoner Could you have a look too?

larsoner:
Thanks @vries !

PR #3838, opened 2025-11-25
Add two tests in test_stdin to increase coverage

vries:
Coverage increases:
- parse_lines: 92% -> 94%
- parse_file : 79% -> 87%

### DIFF
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -1429,7 +1429,7 @@ def run_codespell_stdin(
     return output.count("\n")
 
 
-def test_stdin(tmp_path: Path) -> None:
+def test_stdin(tmp_path: Path, capsys: pytest.CaptureFixture[str]) -> None:
     """Test running the codespell executable."""
     input_file_lines = 4
     text = ""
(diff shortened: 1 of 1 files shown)


codespell_lib/tests/test_basic.py
```diff
@@ -1429,7 +1429,7 @@ def run_codespell_stdin(
     return output.count("\n")
 
 
-def test_stdin(tmp_path: Path) -> None:
+def test_stdin(tmp_path: Path, capsys: pytest.CaptureFixture[str]) -> None:
```
DimitriPapadopoulos, line 1432: The `capsys` argument does not seem to be used. Why did you add it?

vries: There's an implicit use in MainWrapper.main:
```
        capsys = frame.f_locals["capsys"]
```
so if I drop the parameter from test_stdin, I get:
```
______________________________________ test_stdin _______________________________________
codespell_lib/tests/test_basic.py:1449: in test_stdin
    result = cs.main("-", "-w", std=True)
codespell_lib/tests/test_basic.py:52: in main
    capsys = frame.f_locals["capsys"]
E   KeyError: 'capsys'
--------------------------------- Captured stdout call ----------------------------------
```

PR #3841, opened 2025-11-28
codespell_lib/_codespell.py: read additional args from file with with…

julian-smith-artifex-com:
This allows arguments to be read from a file with `@PATH`.

It uses https://docs.python.org/3/library/argparse.html#fromfile-prefix-chars.

I'd like this change specifically to avoid having to specify a large number of paths on the command line - when codespell reports errors, ptest dumps the entire command line which makes it difficult to find the actual codespell output.

There is currently one test failure, of `codecov/patch`. I don't understand what's going on here -  the new function `convert_arg_line_to_args()` is definitely called by new test `test_args_from_file()`.

### DIFF
--- a/codespell_lib/tests/test_basic.py
+++ b/codespell_lib/tests/test_basic.py
@@ -1519,7 +1519,7 @@ def test_args_from_file(
     assert "tmp2:1: exmaple ==> example\n" in stdout, f"{stdout=}"
     assert "tmp3:1: abilty ==> ability\n" in stdout, f"{stdout=}"
     assert code, f"{code=}"
-    
+
     # Run same test via cs_.main() so code coverage checks work.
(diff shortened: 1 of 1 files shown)


julian-smith-artifex-com:
I'm not sure why the codecov/patch test is failing - the new test definitely ends up using the new `convert_arg_line_to_args()` function.

julian-smith-artifex-com:
I've added an explicit call to `cs_.main()` to persuade code coverage that we are indeed testing the new code, so all checks now pass.

[This call duplicates what we are already doing by running a child process, which presumably is not being seen by code coverage?]

Please review.

larsoner:
Looks reasonable, thanks @julian-smith-artifex-com !

PR #3843, opened 2025-12-01
Add `appedix->appendix` and variants to dictionary

nathanjmcdougall:


codespell_lib/data/dictionary.txt
```diff
@@ -5467,7 +5467,10 @@ appeciating->appreciating
 appeciation->appreciation
 appeciative->appreciative
 appeded->appended
+appedices->appendices
```
DimitriPapadopoulos, line 5471: Could you add singular?
```suggestion
appedice->appendice
appedices->appendices
```

nathanjmcdougall: I don't think `appendice` is a real word. The singular form is `appendix`.

DimitriPapadopoulos: It's a word in biology: [appendice](https://www.oed.com/search/dictionary/?q=appendice). Probably from French instead of directly from Latin.

But I guess it's rare.

nathanjmcdougall: I'm happy to put it in the rare dictionary, but I would propose adding it as `appedice->appendices, appendice, appendix`

mhrtalmshar419-sys:
Hype me too mint 🥰

codespell_lib/data/dictionary.txt
DimitriPapadopoulos: You can put it in the regular dictionary. The rare dictionary is for typos that are actual (but rare) words.

DimitriPapadopoulos:
Thank you @nathanjmcdougall.

PR #3851, opened 2025-12-18
Add spelling correction for Mitigate and variants.

cfi-gb:
See e.g. https://grep.app/search?q=mitgation

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -38781,6 +38781,12 @@ misusin->misusing
 mitake->mistake
 mitaken->mistaken
 mitakes->mistakes
+mitgate->mitigate
+mitgated->mitigated
+mitgates->mitigates
+mitgating->mitigating
+mitgation->mitigation
+mitgations->mitigations
 miticate->mitigate
 miticated->mitigated
 miticateing->mitigating


codespell_lib/data/dictionary.txt
```diff
@@ -38781,6 +38781,10 @@ misusin->misusing
 mitake->mistake
 mitaken->mistaken
 mitakes->mistakes
+mitgate->mitigate
+mitgated->mitigated
```
DimitriPapadopoulos, line 38785: Perhaps add all verbal forms?
```suggestion
mitgate->mitigate
mitgated->mitigated
mitgates->mitigates
mitgating->mitigating
```

cfi-gb: Ah, yes. Absolutely. Missed these completely, thanks a lot! :+1:

cfi-gb: For some reason i wasn't able to apply the suggestion due to a recurring `This diff has recently been updated. Please and try again.` while nothing had been updated. Maybe a bug in GitHub itself so have force-pushed a new commit including these :+1:

DimitriPapadopoulos: This is a GitHub bug. I had reported it 1 or 2 years ago, but patching big files is a bit of a niche case, not a priority for Microsoft.

cfi-gb: Good to know for the future (as i have seen this only here so far), thanks a lot for the additional background info :+1:

PR #3877, opened 2026-02-22
PEP 735 compliance: dependency groups

DimitriPapadopoulos:
Only for development and build time dependencies, not run time depdendencies that are meant to be published with the package.


larsoner:
Thanks @DimitriPapadopoulos

PR #3884, opened 2026-03-04
Improve output in interactive mode

### LINKED ISSUE #1925: Show filename in interactive mode
When running in interactive mode, only the line to modify and the suggestion are displayed. Would it be possible to also display the filename and the line number, so that we can have some context if necessary to check whether the suggestion is pertinent? Right now, they are shown *after* we have answered, which is a odd... Thanks!

darkmattercoder:
When in interactive mode, the file name and
line number is now printed before the User
is asked for confirmation.

This adds context to the suggestion. Before, the file was only displayed after the change had been applied.

Also the filename color is applied in the
`FIXED: ` printout.

Fix codespell-project/codespell#1925

See: 

grafik


darkmattercoder:
Am I supposed to to something about the failing checks?

Especially the decrease in coverage, which imho mainly originates from added line breaks.

I still do not know why some of the pytests fail. It looks as if that were due to some dependency issue. They ran successfully on main a few days ago, and I did not do anything regarding dependencies in my commits, so I am a bit clueless here atm.

larsoner:
Should be fixed by #3886 which I just merged into your branch

darkmattercoder:
What is the review policy for the codespell project? Am I supposed to do something else / provide more information? I was happy to do so. 

What do you need from my side to get this integrated?

larsoner:
@darkmattercoder maintainers aren't pinged on commits, so best to add a comment once everything is green and ready for review!

LGTM so in it goes, thanks @darkmattercoder !

PR #3916, opened 2026-04-15
Add spelling correction for accorss/accors.

cfi-gb:
See e.g.:
- https://grep.app/search?q=accorss
- https://grep.app/search?words=true&q=accors

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -1011,6 +1011,8 @@ accoridngly->accordingly
 accoring->according, occurring,
 accoringly->accordingly
 accorndingly->accordingly
+accors->accord, across,
+accorss->accords, across,
 accort->accord
 accortance->accordance
 accorted->accorded


codespell_lib/data/dictionary.txt
```diff
@@ -1011,6 +1011,8 @@ accoridngly->accordingly
 accoring->according, occurring,
 accoringly->accordingly
 accorndingly->accordingly
+accors->accord, across,
+accorss->accords, across,
```
DimitriPapadopoulos, line 1015: Perhaps `acosts` too?

cfi-gb: Mhhh, you mean `accost` / `accosts` as an additional correction option for both or something else?

DimitriPapadopoulos: I was indeed thinking of `accosts` as an additional correction option:
```
accorss->accords, across, accosts,
```
But then it's perhaps far-stretched. [Occurrences of `accorss`](https://github.com/search?q=accorss&type=code&p=2) I have found are actually a misspelled `across`.

Let's leave it as is.

DimitriPapadopoulos (approved):

codespell_lib/data/dictionary.txt
cfi-gb: Ah, i see. I'm fine with leaving it as is, thanks a lot for the review anyway :+1:

PR #3917, opened 2026-04-16
Allow use --builtin=all to use every builtin dictionary

### LINKED ISSUE #3221: Feature Request: "--builtin=all"
It would be nice to have a surrogate value for the `builin` option to select all available dictionaries.

    codespell --builtin=all

AlightSoulmate:
Close #3221

Although using all built-in dictionaries is not a common operation, as different dictionaries correspond to different usage scenarios or languages, I still recommend maintainers introduce the feature.


larsoner:
Seems reasonable to me, marking for merge-when-green and merged `main` into the branch. Thanks in advance @AlightSoulmate !

PR #3943, opened 2026-05-27
Add rourter->router and friends

peternewman:


codespell_lib/data/dictionary.txt
```diff
@@ -52280,6 +52280,10 @@ rountriping->roundtripping, round-tripping, round tripping,
 rountripped->roundtripped, round-tripped, round tripped,
 rountripping->roundtripping, round-tripping, round tripping,
 rountrips->roundtrips, round-trips, round trips,
+rourted->routed
+rourter->router
+rourters->routers
+rourting->routing
```
DimitriPapadopoulos: Perhaps these could also be typos of::
```suggestion
rourted->routed, rousted, courted,
rourter->router, rouster, courter,
rourters->routers, rousters, courters,
rourting->routing, rousting, courting,
```
Probably rare, so I'm not certain it's worth adding them. What do you think?

peternewman: > Probably rare, so I'm not certain it's worth adding them. What do you think?

I'm always keen to avoid auto-correcting to an incorrect fix, so these seem plausible.

Do you want to try merging your suggestion. GitHub isn't letting me in any of three ways (an issue I had the other day too)...

DimitriPapadopoulos: GitHub cannot change files larger than ~ 50 000 lines; That has been a known issue for more than a year, it probably won't be fixed.

peternewman: Ah perhaps its just here I've hit it before then!

DimitriPapadopoulos (approved):

PR #3946, opened 2026-05-28
Add correction for 'egineering' to 'engineering'

Flo3561:


codespell_lib/data/dictionary.txt
```diff
@@ -24090,6 +24090,7 @@ egenralize->generalize
 egenralized->generalized
 egenralizes->generalizes
 egenrally->generally
+egineering->engineering
```
DimitriPapadopoulos, line 24095: How about adding other words with the same root?
```suggestion
egineer->engineer
egineered->engineered
egineering->engineering
egineers->engineers
```

Flo3561: good point. Added.

DimitriPapadopoulos (approved):

PR #3950, opened 2026-06-03
feat: add --ignore-sic to skip misspellings marked with [sic]

### LINKED ISSUE #3764: Feature Request: [sic]
How about ignoring mistakes in front of "[sic]"?

kojiromike:
Closes #3764.

## Summary

Adds an opt-in `--ignore-sic` flag that skips a misspelling immediately followed by the editorial `[sic]` marker.

- Only the **single occurrence** preceding the marker is ignored; other misspellings on the same line are still reported.
- A closing quote may sit between the word and the marker, which is the common case when documenting a corrected typo in a changelog, e.g. `correct the "wrod" [sic] typo`.
- Case-insensitive (`[sic]`/`[SIC]`), with optional whitespace before the marker.

## Motivation

The existing inline mechanisms are all annotations aimed at the tool: `codespell:ignore` (whole line), `codespell:ignore <word>` (by spelling, not position — it can't target a single occurrence and requires naming the typo in a comment), and `codespell:ignore-next-line`. None of these fit prose where an intentional/quoted misspelling appears, such as a changelog entry announcing a typo fix.

`[sic]` is a long-standing editorial convention that already belongs in the text on its own merits, so reusing it costs the document nothing and needs no tooling comment. Because it changes default behavior only when explicitly requested, it is gated behind a flag rather than folded into `codespell:ignore`.

## Implementation

- New `sic_regex` and an `--ignore-sic` argument (default off).
- The check sits beside the existing backslash-escape false-alarm lookbehind in the match loop, using a positional `re.match(line, match.end())` anchor.

## Test plan

- [x] New parametrized `test_ignore_sic`: bare/quoted/whitespace/case variants, the changelog quoted form, single-occurrence scoping, and negative cases (`(sic)`, `[sick]`, intervening word). Asserts off-by-default behavior too.
- [x] `ruff check`, `ruff format --check`, `mypy --strict` clean.
- [x] codespell self-check passes (examples reuse the already-allowlisted `wrod`, so no new ignore-words entry).
- [x] Full `test_basic.py` suite: 82 passed, 1 skipped.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -172,6 +172,23 @@ Words should be separated by a comma.
        def wrod(wrods):
            pass
 
+Ignoring misspellings marked with "[sic]"
+-----------------------------------------
+
+The ``--ignore-sic`` option tells codespell to skip a misspelling that is
+followed by the editorial ``[sic]`` marker (case-insensitive). Only the single
+occurrence preceding the marker is ignored, so other misspellings on the same
+line are still reported. A closing quote may sit between the word and the
+marker, which is the common case when documenting a corrected typo (for example
+in a changelog):
+
+.. code-block:: text
+
+    correct the "wrod" [sic] typo in a changelog entry
+
+Unlike ``codespell:ignore``, the marker is part of the prose itself and does not
+require naming the word in a tooling comment.
+
 Using a config file
 -------------------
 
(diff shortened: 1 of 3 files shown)


larsoner:
Seems reasonable to me but tests are failing

kojiromike:
Sorry, I thought `wrod` was already baked in as an ignore in the CI self-checks. It wasn't.

larsoner:
Looks very simple in the end, marking for merge when green. Thanks in advance @kojiromike !

PR #3960, opened 2026-06-17
Add spelling corrections for simpe and variants.

cfi-gb:
See e.g.:
- https://grep.app/search?words=true&q=simpest
- https://grep.app/search?words=true&q=simpe

Note: `simper` might be a typo of `simpler` but is also a valid word so i have left that out.

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -54560,6 +54560,8 @@ simmetry->symmetry
 simmilar->similar
 simmilarity->similarity
 simmilarly->similarly
+simpe->simple, simper,
+simpest->simplest
 simpicity->simplicity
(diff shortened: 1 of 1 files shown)


codespell_lib/data/dictionary.txt
```diff
@@ -54560,6 +54560,8 @@ simmetry->symmetry
 simmilar->similar
 simmilarity->similarity
 simmilarly->similarly
+simpe->simple
```
DimitriPapadopoulos: Could it also be `simper` with a missing `r`?

cfi-gb: Yes, indeed / absolutely. Added :+1:

DimitriPapadopoulos (approved):

PR #3971, opened 2026-07-13
Add --min-word-length option to ignore short words

### LINKED ISSUE #2211: Whishlist: Optionally ignore short words
I often find the need to add to `--ignore-words-list` the names of variables made of two and three letters, but I would find it useful to ignore them on the base of their lengths with a new command line option.

ekanshul:
Closes #2211

Adds a `--min-word-length LENGTH` option so that words shorter than LENGTH are not checked. This helps with codebases full of 2-3 letter variable names that would otherwise each need an entry in `--ignore-words-list`.

Implementation: entries shorter than LENGTH are dropped from the misspelling dictionary right after it is built, so short words are ignored consistently everywhere (file contents, stdin and `--check-filenames`) with no per-word cost during scanning. The default (`0`) keeps current behavior.

Example:

```
$ codespell file.txt
file.txt:1: nwe ==> new
file.txt:1: abandonned ==> abandoned
$ codespell --min-word-length 4 file.txt
file.txt:1: abandonned ==> abandoned
```

Tests included (`test_min_word_length`) covering the length boundaries, case-insensitivity and `--check-filenames`.

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -514,6 +514,17 @@ def parse_options(
         "how they are written in the dictionary file.",
     )
     parser.add_argument(
+        "--min-word-length",
+        action="store",
+        type=int,
+        default=0,
+        metavar="LENGTH",
+        help="only check words at least LENGTH characters long, "
+        "e.g., use 4 to not check words of 3 or fewer characters "
+        "(such as short variable names). "
+        "The default is %(default)s (check all words).",
+    )
+    parser.add_argument(
         "--uri-ignore-words-list",
         action="append",
         metavar="WORDS",
@@ -1433,6 +1444,15 @@ def main(*args: str) -> int:
     misspellings: dict[str, Misspelling] = {}
     for dictionary in use_dictionaries:
         build_dict(dictionary, misspellings, ignore_words)
+    if options.min_word_length > 0:
+        # Words shorter than the requested length can never match a
+        # misspelling, so dropping them from the dictionary up front ignores
+        # them everywhere (file contents and --check-filenames alike).
+        misspellings = {
+            key: value
+            for key, value in misspellings.items()
+            if len(key) >= options.min_word_length
+        }
     colors = TermColors()
     if not options.colors:
         colors.disable()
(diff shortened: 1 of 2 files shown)


chuenchen309:
Applied this branch and exercised it rather than just reading it — it works, and filtering at the dictionary level is the right layer. Sharing the checks in case they're useful to a maintainer:

```
$ cat short.txt
The aas value and the abandonned config.

$ codespell short.txt
short.txt:1: aas ==> ass, as
short.txt:1: abandonned ==> abandoned

$ codespell --min-word-length 4 short.txt
short.txt:1: abandonned ==> abandoned          # short key dropped, long one still caught
```

The one thing I went looking for was whether `len(key)` could mean something other than "word length" — a multi-word dictionary key would make `--min-word-length 4` compare against the whole phrase rather than a word. It can't: **no builtin dictionary has a key containing a space** (I checked all of `codespell_lib/data/dictionary*.txt` — zero). So `len(key)` is unambiguously the word length here. Worth knowing it holds by data rather than by construction, in case custom `-D` dictionaries are ever expected to carry phrases.

Filtering the dictionary once after `build_dict` rather than at each match site also means `--check-filenames` and the `-w` path get the same treatment for free, which matches what the PR description claims.

For context on the size of what it buys: 162 of the default dictionary's keys are ≤3 characters, so `--min-word-length 4` is a meaningful reduction for a codebase full of short identifiers — the use case in #2211.

(Not a maintainer, just a contributor who was recently in this file.)

ekanshul:
Gentle ping on this one — it's been about four weeks, so I wanted to check it hadn't just slipped off the radar.

One thing that may be worth a maintainer's attention: the GitHub Actions workflows on this PR have never actually run. All five runs on the head commit (`Pytest`, `Build`, `codespell`, `autofix.ci`) sit at `completed/action_required`, which is the first-time-contributor approval gate. `pre-commit.ci` is green, but since the required workflows never started, the PR reports as blocked on missing checks rather than on anything in the diff — so there's currently no way for me to show it passing from my side. Approving the workflow run should clear that.

It isn't specific to this PR: there are currently 42 runs in `action_required` across the repo, so a few other contributors are likely stuck the same way.

For context, this closes #2211, which is labelled `good first issue`, and @peternewman's last comment there invited anyone to pick it up. Thanks also to @chuenchen309, who applied the branch and verified the behaviour independently above.

The same gate applies to #3972 and #3973 if they're being triaged together.

Happy to rebase, split this up, or take a different approach if you'd prefer it done another way — just let me know.

PR #3976, opened 2026-07-16
Switch tox.ini to dependency groups

mtelka:
This got forgotten during the initial dependency group work in #3877.


DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Good catch, I wasn't aware that `tox.ini` itself might require changes.

PR #3986, opened 2026-07-22
Add `evernt`->`event` and variants

nathanjmcdougall:


codespell_lib/data/dictionary.txt
```diff
@@ -25799,6 +25799,13 @@ everage->average
 everaged->averaged
 everbody->everybody
 everithing->everything
+evernt->event
+everntful->eventful
+evernts->events
+everntual->eventual
+everntuality->eventuality
+everntually->eventually
+everntuate->eventuate
```
DimitriPapadopoulos: I believe `eventuate` is a rare verb. Anyway, if you do want to add it, I suggest adding all verbal forms:
```suggestion
everntuate->eventuate
everntuated->eventuated
everntuates->eventuates
everntuating->eventuating
```

nathanjmcdougall: I'm not particularly inclined to add it. I'll remove it.

For what it's worth, I don't think it's very rare as much as quite formal.

DimitriPapadopoulos (approved):

PR #3994, opened 2026-08-14
Do not report a fix that --ignore-regex made impossible

### LINKED ISSUE #2056: --ignore-regex does not work with --write-changes
Reproducible in `python3.9-slim` Docker image:
```
$ pip install codespell==2.1.0 &> /dev/null
$ echo 1nd >> /tmp/test
$ echo 1nd_2nd >> /tmp/test
$ cat /tmp/test 
1nd
1nd_2nd
$ codespell /tmp/test  # correctly finds error on first line but not second
/tmp/test:1: 1nd ==> 1st
$ echo $?  # correct non-zero exit code
65
$ codespell /tmp/test --ignore-regex _  # correctly finds error on first and second lines
/tmp/test:1: 1nd ==> 1st
/tmp/test:2: 1nd ==> 1st
$ echo $?  # correct non-zero exit code
65
$ codespell /tmp/test --ignore-regex _ --write-changes  # incorrectly only fixes first error
FIXED: /tmp/test
$ echo $?  # confusing exit code as errors were found
0
$ cat /tmp/test  # only first error is corrected
1st
1nd_2nd
$ codespell /tmp/test --ignore-regex _ --write-changes  # repeated invocations say "fixed" but make no changes
FIXED: /tmp/test
$ echo $?
0
$ cat /tmp/test
1st
1nd_2nd
```

I would expect either:
* The file to be correctly fixed
* An error message like `ERROR: --write-changes cannot be used together with --ignore-regex`

Eljees:
Fixes #2056.

### What happens

```console
$ printf '1nd\n1nd_2nd\n' > test.txt
$ codespell test.txt --ignore-regex _ --write-changes
FIXED: test.txt
  test.txt:1: 1nd ==> 1st
  test.txt:2: 1nd ==> 1st
$ echo $?
0
$ cat test.txt
1st
1nd_2nd
```

Line 2 is reported as fixed, is not fixed, and does not affect the exit code — so a CI run goes green with the misspelling still in the file, and running codespell again prints the same false `FIXED:` forever.

### Why

Detection and writing look at two different texts. Detection walks the line with `--ignore-regex` substituted out (`_ignore_word_sub` replaces the match with a space), so it sees `1nd 2nd` and finds `1nd`. `--write-changes` then substitutes on the **original** line:

```python
if options.write_changes and fix:
    changed = True
    lines[i] = re.sub(rf"\b{word}\b", fixword, lines[i])
    fixed_words.add(word)
    changes_made.append((line_number + 1, word, fixword))
    continue
```

In `1nd_2nd` there is no word boundary between `1nd` and `_`, so `re.sub` matches nothing. But `changed` is set before the result is inspected, `changes_made` records a change that never happened, and `continue` skips the warning branch that would have set the exit code.

### The change

Keep the result of the substitution, and only claim the fix when the line really changed; otherwise fall through to the normal warning. `--ignore-regex` semantics are untouched — the word is still detected, it is just reported instead of silently dropped. This is the second of the two outcomes the reporter said would be acceptable ("either fix it, or tell me it can't").

After:

```console
FIXED: test.txt
  test.txt:1: 1nd ==> 1st
test.txt:2: 1nd ==> 1st
$ echo $?
65
```

### Tests

`test_ignore_regex_with_write_changes` in `codespell_lib/tests/test_basic.py`, next to the existing `--ignore-regex` tests. It asserts all three symptoms: line 1 is rewritten, line 2 is **not** listed as fixed, and the exit code is non-zero. On `main` it fails on the middle assertion.

`pytest codespell_lib/tests/test_basic.py`: **85 passed** on `main`, **86 passed** with this change, no failures either way. `ruff check`, `ruff format --check` and `mypy` are clean (mypy reports the same two pre-existing `chardet` stub errors on an untouched tree).

### DIFF
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -1100,10 +1100,8 @@ def parse_lines(
                         fixed_words.add(word)
                         changes_made.append((line_number + 1, word, fixword))
                         continue
-                    # The word was found in the --ignore-regex substituted text
-                    # but cannot be located in the original line, so no fix can
-                    # be written.  Fall through and report it as a warning
-                    # instead of claiming a fix that did not happen.
+                    # Not found in the original line (e.g. --ignore-regex split
+                    # it out of a larger word), so report it instead (GH-2056).
 
                 # otherwise warning was explicitly set by interactive mode
                 if (
--- a/codespell_lib/_codespell.py
+++ b/codespell_lib/_codespell.py
@@ -1093,11 +1093,17 @@ def parse_lines(
                     continue
 
                 if options.write_changes and fix:
-                    changed = True
-                    lines[i] = re.sub(rf"\b{word}\b", fixword, lines[i])
-                    fixed_words.add(word)
-                    changes_made.append((line_number + 1, word, fixword))
-                    continue
+                    new_line = re.sub(rf"\b{word}\b", fixword, lines[i])
+                    if new_line != lines[i]:
+                        changed = True
+                        lines[i] = new_line
+                        fixed_words.add(word)
+                        changes_made.append((line_number + 1, word, fixword))
+                        continue
+                    # The word was found in the --ignore-regex substituted text
+                    # but cannot be located in the original line, so no fix can
+                    # be written.  Fall through and report it as a warning
+                    # instead of claiming a fix that did not happen.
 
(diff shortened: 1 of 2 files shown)


larsoner:
Thanks @Eljees !

PR #4008, opened 2026-09-03
Add correction for 'dcitionary' to 'dictionary'

Flo3561:


codespell_lib/data/dictionary.txt
```diff
@@ -19019,6 +19019,7 @@ daugthers->daughters
 daybue->debut
 dbeian->Debian
 DCHP->DHCP
+dcitionary->dictionary
```
DimitriPapadopoulos, line 19023: How about adding plural?
```suggestion
dcitionaries>dictionaries
dcitionary->dictionary
```

Flo3561: added

PR #4013, opened 2026-09-08
Add new correction intop->into

Flo3561:


codespell_lib/data/dictionary.txt
```diff
@@ -35343,6 +35343,7 @@ intitiators->initiators
 intity->entity
 intoduction->introduction
 intoloerance->intolerance
+intop->into
```
DimitriPapadopoulos: Although `in top` is rare:
```suggestion
intop->into,in top,
```

Flo3561: added

PR #4015, opened 2026-09-08
Add corrections for 'tsk' and 'tsks' in dictionary

Flo3561:


codespell_lib/data/dictionary.txt
```diff
@@ -60997,6 +60997,8 @@ tryin->trying, try in,
 tryinng->trying
 trys->tries
 tryying->trying
+tsk->task
```
DimitriPapadopoulos: I suspect [`tsk tsk`](https://www.oed.com/search/dictionary/?q=tsk) might appear frequently...

Flo3561: good point, removed it

PR #4017, opened 2026-09-12
relus->rules

George-Ogden:
~relu -> rule is invalid because [ReLU](https://en.wikipedia.org/wiki/Rectified_linear_unit) is a thing~

### DIFF
--- a/codespell_lib/data/dictionary.txt
+++ b/codespell_lib/data/dictionary.txt
@@ -50081,6 +50081,7 @@ relpaces->replaces
 relpacing->replacing
 relpy->reply
 reltive->relative
+relus->rules, relu,
 relvance->relevance
 relvancy->relevancy
 relvant->relevant


DimitriPapadopoulos:
`ReLU` is not in any dictionary. You may ignore this fix if needed.

George-Ogden:
Should I change it to `relus->rules` then?

DimitriPapadopoulos:
Yes, I would recommend that.

codespell_lib/data/dictionary.txt
```diff
@@ -50081,6 +50081,7 @@ relpaces->replaces
 relpacing->replacing
 relpy->reply
 reltive->relative
+relus->rules
```
DimitriPapadopoulos: THis would at least trigger a manual review in case `relus` stands for `ReLU`:
```suggestion
relus->rules,resus,rebus,
```

DimitriPapadopoulos (approved):

DimitriPapadopoulos:
Thank you @George-Ogden.

PR #4020, opened 2026-09-15
Update dictionary with correction for 'fielsd'

Flo3561:
Added additional correction for 'fielsd' to the dictionary.


codespell_lib/data/dictionary.txt
```diff
@@ -28296,6 +28296,7 @@ fielesystems->filesystems
 fielname->filename
 fielneame->filename
 fiels->feels, fields, files, phials,
+fielsd->feels, fields, files,
```
DimitriPapadopoulos:
```suggestion
fielsd->fields,field,feels,files,
```

Flo3561: added

PR #4032, opened 2026-09-21
Warn when both pyproject.toml and .codespellrc configure codespell

### LINKED ISSUE #3999: Feature: Warn if pyproject.toml and .codespellrc found
If `pyproject.toml` and `.codespellrc` exist (by accident!) both, only the later is used.

Would be nice if `codespell` would warn or error about the situation having two config files.

ekanshul:
Fixes #3999

If both `pyproject.toml` (`[tool.codespell]`) and `.codespellrc` (`[codespell]`) are present, only the later file is applied. The TOML file was not listed in "Used config files:" either, so the override was easy to miss.

This change:

- includes `pyproject.toml` in the used-config list when it has a codespell table
- prints a warning that `.codespellrc` takes precedence
- keeps the warning behind `--quiet-level=32` (same bit as other config-file messages)

A test covers the warning and the quiet-level suppression.


larsoner:
Nice, thanks @ekanshul

